PR #21, opened 2015-06-16
updater: bail out if we're not on a booted OSTree deployment

cosimoc:
Or we'll hit a critical assertion when calling into
ostree_sysroot_get_merge_deployment() when we're not on an OSTree
system.

endlessm/eos-shell#5087


sgnn7:
@cosimoc CR pass.

PR #22, opened 2015-06-16
Handle systemd units better

dbnicholson:
[endlessm/eos-shell#5087]

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -17,5 +17,9 @@ override_dh_auto_configure:
 override_dh_strip:
 	dh_strip --dbg-package=eos-updater-dbg
 
+# Don't start the services on install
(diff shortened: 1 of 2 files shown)


debian/rules
```diff
@@ -17,5 +17,9 @@ override_dh_auto_configure:
 override_dh_strip:
 	dh_strip --dbg-package=eos-updater-dbg
 
+# Don't start the services on install
+override_dh_systemd_start:
+	dh_systemd_start --no-start
```
sgnn7: @dbnicholson Why wouldn't we want to start the service on install even on converted systems?

cosimoc:
@sgnn7 Because on converted systems the service literally does nothing :)

PR #23, opened 2015-07-22
Makefile: Move to AM_DISTCHECK_CONFIGURE_FLAGS

ptomato:
Since automake 1.11.2 it is recommended that packages
use AM_DISTCHECK_CONFIGURE_FLAGS instead of
DISTCHECK_CONFIGURE_FLAGS as the latter is intended
to be a user variable.

[endlessm/eos-sdk#3303]


ptomato:
@smspillaz Still needs an Automake version bump to 1.11.2.

ptomato:
**CR pass**

PR #24, opened 2015-09-24
Disallow downgrades

jhiesey:
Report no updates are available if the "new" version doesn't
have a newer timestamp than the currently booted version.
This matches the behavior of "ostree admin upgrade", except
having an older version available isn't reported as an error.

If a downgrade is needed for testing,
"ostree admin upgrade --allow-downgrade" can be used.

[endlessm/eos-shell#2586]


jhiesey:
The associated issue is on the eos 2.5 milestone, so this doesn't need to be merged right away.

PR #26, opened 2016-02-05
Select OSTree branch at run-time based on product model

jprvita:
[endlessm/eos-shell#6079]

### DIFF
--- a/src/eos-updater-util.c
+++ b/src/eos-updater-util.c
@@ -170,7 +170,6 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
   gs_free gchar *baseurl = NULL;
   gs_free gchar *uri = NULL;
   gs_free gchar *query = NULL;
-  gs_unref_ptrarray GPtrArray *cur_deployments = NULL;
   gs_unref_object OstreeDeployment *merge_deployment = NULL;
   const gchar *osname;
   const gchar *booted;
--- a/src/eos-updater-util.c
+++ b/src/eos-updater-util.c
@@ -22,6 +22,7 @@
 
 #include "eos-updater-util.h"
 #include <ostree.h>
+#include <libsoup/soup.h>
 
 static const GDBusErrorEntry eos_updater_error_entries[] = {
   { EOS_UPDATER_ERROR_WRONG_STATE, "com.endlessm.Updater.Error.WrongState" }
@@ -106,6 +107,48 @@ eos_updater_local_repo (void)
   return ret;
 }
 
+static gchar *
+cleanstr (gchar *s)
+{
+  gchar *i, *j, *k;
+
+  if (!s)
+    return s;
+
+  for (i = s; *i; i++)
+    /* only allow printable */
+    if (*i < 32 || *i > 126)
+      for (j = i; *j; j++)
+        {
+          k = j + 1;
+          *j = *k;
+        }
+
+  return s;
+}
+
+static const gchar *BRANCHES_CONFIG_PATH = "eos-branch";
+static const gchar *DEFAULT_GROUP = "Default";
+static const gchar *OSTREE_REF_KEY = "OstreeRef";
+static const gchar *ON_HOLD_KEY = "OnHold";
+static const gchar *DT_COMPATIBLE = "/proc/device-tree/compatible";
+static const gchar *DMI_PATH = "/sys/class/dmi/id/";
+static const gchar *dmi_attributes[] =
+  {
+    "bios_date",
+    "bios_vendor",
+    "bios_version",
+    "board_name",
+    "board_vendor",
+    "board_version",
+    "chassis_vendor",
+    "chassis_version",
+    "product_name",
+    "product_version",
+    "sys_vendor",
+    NULL,
+  };
+
 gboolean
 eos_updater_resolve_upgrade (EosUpdater  *updater,
                              OstreeRepo *repo,
@@ -115,14 +158,28 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
                              GError    **error)
 {
   gboolean ret = FALSE;
+  guint status = 0;
   gs_free gchar *o_refspec = NULL;
   gs_free gchar *o_remote = NULL;
   gs_free gchar *o_ref = NULL;
+  gs_free gchar *vendor = NULL;
+  gs_free gchar *product = NULL;
+  gs_free gchar *p_id = NULL;
+  gs_free gchar *p_ref = NULL;
+  gs_free gchar *osgroup = NULL;
+  gs_free gchar *baseurl = NULL;
+  gs_free gchar *uri = NULL;
+  gs_free gchar *query = NULL;
   gs_unref_ptrarray GPtrArray *cur_deployments = NULL;
   gs_unref_object OstreeDeployment *merge_deployment = NULL;
   const gchar *osname;
   const gchar *booted;
   GKeyFile *origin;
+  GKeyFile *repo_config;
+  GHashTable *hw_descriptors = NULL;
+  gs_unref_object SoupSession *soup = NULL;
+  gs_unref_object SoupMessage *msg = NULL;
+  gs_unref_keyfile GKeyFile *bkf = NULL;
 
   OstreeSysroot *sysroot = ostree_sysroot_new_default ();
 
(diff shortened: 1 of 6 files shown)


dbnicholson:
To @ramcq's point about garbage in the DMI table, you can have a look at `g_utf8_validate ()` and possibly `g_utf8_normalize ()`. See https://developer.gnome.org/glib/stable/glib-Unicode-Manipulation.html.

dbnicholson:
So, I have a couple minor comments, but in general this looks good. However, we can't actually apply it as is since none of the ostree branches being specified actually exist. To get the code in, you may want to change the eos-updater-branches.conf if this PR to only have a commented out example.

src/eos-updater-util.c
```diff
@@ -154,14 +211,147 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
...
   if (booted_checksum)
     *booted_checksum = g_strdup (booted);
 
+  /* Get branch configuration data baseurl */
+  repo_config = ostree_repo_get_config (repo);
+  osgroup = g_strdup_printf ("remote \"%s\"", osname);
+  baseurl = g_key_file_get_string (repo_config, osgroup, "url", NULL);
+  if (!baseurl)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND,
```
cosimoc: Can't you just let `g_key_file_get_string()` set the error?

src/eos-updater-util.c
```diff
@@ -154,14 +211,147 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
...
+                   "No url found for %s, cannot upgrade", osname);
+      goto out;
+    }
+
+  /* Get product identifier */
+  hw_descriptors = g_hash_table_new_full (g_str_hash, g_str_equal,
+                                          g_free, g_free);
+  if (g_file_test (DT_COMPATIBLE, G_FILE_TEST_EXISTS))
```
cosimoc: Instead of testing this beforehand, you could use `g_file_get_contents()` or `g_file_load_contents()` and fall back to x86 detection if that fails.

src/eos-updater-util.c
```diff
@@ -154,14 +211,147 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
...
+        {
+          gs_free gchar *path = NULL;
+          gs_unref_object GFile *fp = NULL;
+          gs_free gchar *fc = NULL;
+          gsize len;
+
+          path = g_strconcat (DMI_PATH, dmi_attributes[i], NULL);
+          fp = g_file_new_for_path (path);
+          if (g_file_load_contents (fp, NULL, &fc, &len, NULL, NULL))
+          {
```
cosimoc: Inconsistent indentation

data/eos-branch
```diff
@@ -0,0 +1,37 @@
+# Endless product-specific branches configuration
+#
+# Copyright (C) 2016 Endless Mobile Inc.
```
cosimoc: I don't think this necessarily needs the copyright - it's just a keyfile. But if it does, it should probably also mention the license.

src/eos-updater-util.c
```diff
@@ -154,14 +211,147 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
...
+  vendor = g_strdup (g_hash_table_lookup (hw_descriptors, "sys_vendor"));
+  product = g_strdup (g_hash_table_lookup (hw_descriptors, "product_name"));
+  p_id = g_strconcat (cleanstr (vendor), " ", cleanstr (product), NULL);
+  message ("Product group: %s", p_id);
+
+  /* Build the HTTP URI */
+  g_hash_table_insert (hw_descriptors, g_strdup ("ref"), g_strdup (o_ref));
+  g_hash_table_insert (hw_descriptors, g_strdup ("commit"), g_strdup (booted));
+  query = soup_form_encode_hash (hw_descriptors);
+  uri = g_strconcat (baseurl, "/", BRANCHES_CONFIG_PATH, "?", query, NULL);
```
cosimoc: I wonder if you couldn't use SoupURI here instead

src/eos-updater-util.c
```diff
@@ -154,14 +211,147 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
...
+  status = soup_session_send_message (soup, msg);
+  if (!SOUP_STATUS_IS_SUCCESSFUL (status))
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_HOST_NOT_FOUND,
+                   "Failed to download branch config data (HTTP %d),"
+                   " cannot upgrade", status);
+      goto out;
+    }
+
+  g_object_get (G_OBJECT (msg), "response-body", &msg_body, NULL);
```
cosimoc: There's no need to cast to `G_OBJECT()` because `g_object_get()` takes a `gpointer`. Also, pretty sure you need to free `msg_body` here.

src/eos-updater-util.c
```diff
@@ -154,14 +211,147 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
...
+  if (g_key_file_has_group (bkf, p_id))
+    {
+      if (g_key_file_get_boolean (bkf, p_id, ON_HOLD_KEY, NULL))
+        {
+          message ("Product is on hold, nothing to upgrade here");
+          ret = TRUE;
+          goto out;
+        }
+      p_ref = g_key_file_get_string (bkf, p_id, OSTREE_REF_KEY, NULL);
+      if (!p_ref)
```
cosimoc: Same as above - could probably just use the error that would be returned. You can use `g_prefix_error()` if really needed

src/eos-updater-util.c
```diff
@@ -154,14 +211,147 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
...
+    {
+      if (g_key_file_get_boolean (bkf, DEFAULT_GROUP, ON_HOLD_KEY, NULL))
+        {
+          message ("No product-specific configuration and %s is on hold, "
+                   "nothing to upgrade here", DEFAULT_GROUP);
+          ret = TRUE;
+          goto out;
+        }
+      p_ref = g_key_file_get_string (bkf, DEFAULT_GROUP, OSTREE_REF_KEY, NULL);
+      if (!p_ref)
```
cosimoc: Ditto

dbnicholson:
You should add `libsoup-2.4` to the `PKG_CHECK_MODULES([EOS_UPDATER]` setting in configure.ac. It might be pulled in implicitly, but it's best to be explicit here. Not sure if you need a specific version here or not.

src/eos-updater-util.c
```diff
@@ -106,6 +107,48 @@ eos_updater_local_repo (void)
...
+cleanstr (gchar *s)
+{
+  gchar *i, *j, *k;
+
+  if (!s)
+    return s;
+
+  for (i = s; *i; i++)
+    /* only allow printable */
+    if (*i < 32 || *i > 126)
```
cosimoc: Use `g_ascii_isprint()` or something like that?

src/eos-updater-util.c
dbnicholson: I looked at the `msg_body` case in the documentation. It should be freed automatically when `msg` goes away.

https://developer.gnome.org/libsoup/stable/SoupMessageBody.html#soup-message-body-free

cosimoc: True, but `g_object_get()` will copy it for you. If it's accessed directly like you suggest in another comment, it should not be necessary to free it.

dbnicholson: Ah, that makes sense. Thanks for the explanation.

data/eos-branch
jprvita: Actually, we shouldn't merge this file here, as it is going to live in eos-administration and be deployed by some jenkins+ansible magic (see endlessm/eos-shell#6417). But I'll take your point and remove the copyright notice there.

src/eos-updater-util.c
jprvita: I didn't look at what range does `g_ascii_isprint()` considers as printable chars, but many definitions of it I found in the web says DEL (127) is also part of the printable range, and we do not want that as part of our query/key. I like having it explicit in this case.

src/eos-updater-util.c
jprvita: Probably yes, I have no reason to set a custom message here.

src/eos-updater-util.c
jprvita: Ack.

src/eos-updater-util.c
jprvita: Ack

src/eos-updater-util.c
jprvita: Thanks for the catch, my editor is not properly configured for GNU coding style.

src/eos-updater-util.c
jprvita: I could, but I didn't see any clear advantage of one over the other when checking libsoup's API. Plus, I was put off of diving more into the differences by this comment in SoupURI's documentation:

> Many applications will not need to use SoupURI directly at all; on the client side, soup_message_new() takes a stringified URI (...)

Please alert me if we would be gaining something relevant by switching to SoupURI, as I'm not an experienced libsoup user, but if it is just a matter of preference I would rather keep it this way.

src/eos-updater-util.c
jprvita: In my initial attempt at this I was freeing `msg_body` and getting a double free crash, so I'm pretty sure it is not being copied.

src/eos-updater-util.c
jprvita: I personally think the current way is a bit easier to read, as I would have to move `fp` declaration out of the `if` branches and use separate variables for each branch, adding more lines to the already incredibly long list of local vars this function has. Do you think your suggestion has a performance or memory usage advantage comparing with the current proposal?

dbnicholson:
From my side, the only thing I'm still unsure of is the handling of the `upgrade_ref` out parameter and what happens in the `OnHold` case. I think the intention is that for `OnHold`, you return `TRUE`, but leave `upgrade_ref` set to `NULL`? Is that right? What is the caller supposed to do in this case? Because it appears that whenever `eos_updater_resolve_upgrade` returns `TRUE`, the callers continue on assuming everything's working.

src/eos-updater-util.c
cosimoc: No, just that I got into the habit of never doing test+read as a pattern, because it's racy. It doesn't matter for this file though.

src/eos-updater-util.c
cosimoc: That's OK if you don't want to change it. I was just trying to avoid the bare strconcat.

src/eos-updater-util.c
cosimoc: Looks to me like it should be copied: https://git.gnome.org/browse/libsoup/tree/libsoup/soup-message.c#n302
But again, should not matter if you follow @dbnicholson's suggestion.

jprvita:
@dbnicholson You're right about the `OnHold` case, thanks for catching that. I think I'm starting to get a bit myopic when looking at this code.

src/eos-updater-util.c
```diff
@@ -154,14 +210,141 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
...
+  g_hash_table_destroy (hw_descriptors);
+  message ("Branches configuration URI: %s", uri);
+
+  /* Download branch configuration data */
+  soup = soup_session_new ();
+  msg = soup_message_new ("GET", uri);
+  status = soup_session_send_message (soup, msg);
+  if (!SOUP_STATUS_IS_SUCCESSFUL (status))
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_HOST_NOT_FOUND,
```
cosimoc: Nitpick: `HOST_NOT_FOUND` does not always seem appropriate here. Consider using the generic `G_IO_ERROR_FAILED`

jprvita: Right, thanks for the suggestion.

dbnicholson:
@jprvita Just one last question about the default group name. Otherwise, I think this is OK.

PR #27, opened 2016-02-10
Shell/6079 debian

jprvita:
[endlessm/eos-shell#6079]


debian/control
```diff
@@ -17,7 +18,8 @@ Package: eos-updater
 Section: misc
 Architecture: any
 Depends: ${shlibs:Depends}, ${misc:Depends},
-	systemd (>= 200)
+	systemd (>= 200),
+	libsoup2.4-1
```
dbnicholson: You don't need this as `${shlibs:Depends}` will add this for you.

jprvita: Ok, tks!

PR #29, opened 2016-02-25
Shell/6407

jprvita:
[endlessm/eos-shell#6407]

### DIFF
--- a/src/eos-updater-util.c
+++ b/src/eos-updater-util.c
@@ -23,6 +23,17 @@
 #include "eos-updater-util.h"
 #include <ostree.h>
 #include <libsoup/soup.h>
+#include <eosmetrics/eosmetrics.h>
+
+
+/*
+ * Records which branch will be used by the updater. The payload is a 4-tuple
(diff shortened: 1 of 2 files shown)


src/eos-updater-util.c
```diff
@@ -352,6 +366,16 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
...
+  if ((p_ref || on_hold) && !metric_sent)
+    {
+      emtr_event_recorder_record_event (emtr_event_recorder_get_default (),
+                                        EOS_UPDATER_BRANCH_SELECTED,
+                                        g_variant_new ("(sssb)", vendor,
+                                                       product,
+                                                       on_hold ? o_ref : p_ref,
+                                                       on_hold));
+      metric_sent = TRUE;
+    }
```
dbnicholson: I kinda feel like this is the wrong place to be shoving this, but that would require providing the vendor and product back out to the caller. That would require more refactoring. For another day...

PR #30, opened 2016-03-29
T6407

jprvita:


src/eos-updater-util.c
```diff
@@ -367,14 +367,18 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
...
-  if ((p_ref || on_hold) && !metric_sent)
+  if (( (upgrade_ref && *upgrade_ref) || on_hold) && !metric_sent)
     {
-      emtr_event_recorder_record_event (emtr_event_recorder_get_default (),
-                                        EOS_UPDATER_BRANCH_SELECTED,
-                                        g_variant_new ("(sssb)", vendor,
-                                                       product,
-                                                       on_hold ? o_ref : p_ref,
-                                                       on_hold));
+      message ("Recording metric event %s: (%s, %s, %s, %d)",
```
Kurt-von-Laven: This seems duplicitous with https://github.com/endlessm/eos-metrics/blob/master/eosmetrics/emtr-event-recorder.c#L724-L736, but please feel free to keep it if there is some rational for it I am unaware of.

src/eos-updater-util.c
```diff
@@ -367,14 +367,18 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
   shuffle_out_values (upgrade_ref, p_ref, NULL);
 
 out:
-  if ((p_ref || on_hold) && !metric_sent)
+  if (( (upgrade_ref && *upgrade_ref) || on_hold) && !metric_sent)
```
Kurt-von-Laven: nit: rm the space between the second and third open parentheses per https://github.com/endlessm/eos-documentation/wiki/C-Coding-Style#whitespace.

src/eos-updater-util.c
```diff
@@ -367,14 +367,18 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
   shuffle_out_values (upgrade_ref, p_ref, NULL);
```
Kurt-von-Laven: nit: "decided" --> "decide" in second commit message.

Kurt-von-Laven: nit: Consider explaining in the third commit message why the synchronous API should be used here rather than the asynchronous API.

src/eos-updater-util.c
jprvita: I would rather keep this one around, together with the other debug messages `eos-updater` already always emits by default.

Kurt-von-Laven: Works for me.

Kurt-von-Laven:
**CR Pass**

PR #33, opened 2016-03-30
build: Require libsystemd instead of libsystemd-journal

dbnicholson:
After the systemd upgrade to 229, libsystemd-journal is gone.

https://phabricator.endlessm.com/T11107


dbnicholson:
Same merge in https://github.com/endlessm/eos-updater/pull/31. Closing this.

PR #34, opened 2016-03-30
Depend on libsystemd-dev instead of libsystemd-journal-dev

dbnicholson:
After the systemd upgrade to 229, libsystemd-journal-dev is gone.

https://phabricator.endlessm.com/T11107


dbnicholson:
Same merge in https://github.com/endlessm/eos-updater/pull/32. Closing this.

PR #35, opened 2016-05-09
Cleanup old remote ref

dbnicholson:
Keep track of the original update refspec so it can be deleted and not keep a reference to a no longer wanted tree.

https://phabricator.endlessm.com/T11010

### DIFF
--- a/src/eos-updater-apply.c
+++ b/src/eos-updater-apply.c
@@ -73,10 +73,12 @@ apply (GTask *task,
        GCancellable *cancel)
 {
   EosUpdater *updater = EOS_UPDATER (object);
+  OstreeRepo *repo = OSTREE_REPO (task_data);
   GError *error = NULL;
   GMainContext *task_context = g_main_context_new ();
   const gchar *update_id = eos_updater_get_update_id (updater);
   const gchar *update_refspec = eos_updater_get_update_refspec (updater);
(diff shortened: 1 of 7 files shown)


src/eos-updater-apply.c
```diff
@@ -111,6 +113,30 @@ apply (GTask *task,
                                    &error))
     goto error;
 
+  /* If the original refspec is not the update refspec, then we may have
+   * a ref to a no longer needed tree. Delete that remote ref so the the
```
cosimoc: `the the`

src/eos-updater.xml
```diff
@@ -7,6 +7,7 @@
     <property name="State"            type="u" access="read"/>
     <property name="UpdateID"         type="s" access="read"/>
     <property name="UpdateRefspec"    type="s" access="read"/>
+    <property name="OriginalRefspec"  type="s" access="read"/>
```
cosimoc: Any reason this gets exposed on the bus?

dbnicholson: @cosimoc Probably just my inexperience with gdbus-codegen. All the other EosUpdater state corresponds to these bus properties. I didn't know how to add a non-exported property to the object.

On the other hand, I don't know any reason this should be hidden from the client. It might be good fit them to know when the upgrade path changes from the on disk configuration.

dbnicholson:
I fixed the comment and left a reply about the bus property.

PR #36, opened 2016-05-12
Cleanups

krnowak:
These are some cleanups and refactorings I did when working on avahi stuff. Probably best reviewed commit-by-commit.

@iaguis: You wanted to see this.

### DIFF
--- a/src/eos-updater-apply.c
+++ b/src/eos-updater-apply.c
@@ -81,10 +81,10 @@ apply (GTask *task,
   const gchar *orig_refspec = eos_updater_get_original_refspec (updater);
   gint bootversion = 0;
   gint newbootver = 0;
-  gs_unref_object OstreeDeployment *merge_deployment = NULL;
-  gs_unref_object OstreeDeployment *new_deployment = NULL;
+  g_autoptr(OstreeDeployment) merge_deployment = NULL;
+  g_autoptr(OstreeDeployment) new_deployment = NULL;
   GKeyFile *origin = NULL;
-  gs_unref_object OstreeSysroot *sysroot = NULL;
+  g_autoptr(OstreeSysroot) sysroot = NULL;
 
   g_main_context_push_thread_default (task_context);
 
@@ -120,7 +120,7 @@ apply (GTask *task,
    */
   if (g_strcmp0 (update_refspec, orig_refspec) != 0)
     {
-      gs_free gchar *rev = NULL;
+      g_autofree gchar *rev = NULL;
 
       if (!ostree_repo_resolve_rev (repo, orig_refspec, TRUE, &rev, &error))
         goto error;
@@ -165,7 +165,7 @@ handle_apply (EosUpdater            *updater,
               GDBusMethodInvocation *call,
               gpointer               user_data)
 {
-  gs_unref_object GTask *task = NULL;
+  g_autoptr(GTask) task = NULL;
   EosUpdaterState state = eos_updater_get_state (updater);
 
   switch (state)
--- a/src/eos-updater-apply.c
+++ b/src/eos-updater-apply.c
@@ -21,6 +21,7 @@
  */
 
 #include "eos-updater-apply.h"
+#include "eos-updater-data.h"
 #include <ostree.h>
 
 static void
@@ -71,7 +72,8 @@ apply (GTask *task,
        GCancellable *cancel)
 {
   EosUpdater *updater = EOS_UPDATER (object);
-  OstreeRepo *repo = OSTREE_REPO (task_data);
+  EosUpdaterData *data = task_data;
+  OstreeRepo *repo = data->repo;
   GError *error = NULL;
   GMainContext *task_context = g_main_context_new ();
   const gchar *update_id = eos_updater_get_update_id (updater);
@@ -163,7 +165,6 @@ handle_apply (EosUpdater            *updater,
               GDBusMethodInvocation *call,
               gpointer               user_data)
 {
-  OstreeRepo *repo = OSTREE_REPO (user_data);
   gs_unref_object GTask *task = NULL;
   EosUpdaterState state = eos_updater_get_state (updater);
 
@@ -180,8 +181,8 @@ handle_apply (EosUpdater            *updater,
     }
 
   eos_updater_set_state_changed (updater, EOS_UPDATER_STATE_APPLYING_UPDATE);
-  task = g_task_new (updater, NULL, apply_finished, repo);
-  g_task_set_task_data (task, g_object_ref (repo), g_object_unref);
+  task = g_task_new (updater, NULL, apply_finished, user_data);
+  g_task_set_task_data (task, user_data, NULL);
   g_task_run_in_thread (task, apply);
 
   eos_updater_complete_apply (updater, call);
--- a/src/eos-updater-apply.c
+++ b/src/eos-updater-apply.c
@@ -31,8 +31,6 @@ apply_finished (GObject *object,
   EosUpdater *updater = EOS_UPDATER (object);
   GTask *task;
   GError *error = NULL;
-
-  OstreeRepo *repo = OSTREE_REPO (user_data);
   gboolean bootver_changed = FALSE;
 
   if (!g_task_is_valid (res, object))
--- a/src/eos-updater-data.h
+++ b/src/eos-updater-data.h
@@ -1,0 +1,41 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*-
+ *
+ * Copyright 2016 Kinvolk GmbH
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ *
+ * Author: Krzesimir Nowak <krzesimir@kinvolk.io>
+ */
+
+#pragma once
+
+#include <ostree.h>
+
+G_BEGIN_DECLS
+
+typedef struct EosUpdaterData EosUpdaterData;
+
+struct EosUpdaterData
+{
+  OstreeRepo *repo;
+};
+
+void eos_updater_data_init (EosUpdaterData *data,
+                            OstreeRepo *repo);
+
+void eos_updater_data_clear (EosUpdaterData *data);
+
+G_END_DECLS
--- a/src/eos-updater-data.h
+++ b/src/eos-updater-data.h
@@ -33,9 +33,13 @@ struct EosUpdaterData
   OstreeRepo *repo;
 };
 
+#define EOS_UPDATER_DATA_INIT { NULL }
+
 void eos_updater_data_init (EosUpdaterData *data,
                             OstreeRepo *repo);
 
 void eos_updater_data_clear (EosUpdaterData *data);
+
+G_DEFINE_AUTO_CLEANUP_CLEAR_FUNC (EosUpdaterData, eos_updater_data_clear)
 
 G_END_DECLS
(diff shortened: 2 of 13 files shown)


src/eos-updater-util.c
```diff
@@ -160,127 +153,136 @@ static const gchar *dmi_attributes[] =
...
     }
 
-  merge_deployment = ostree_sysroot_get_merge_deployment (sysroot, NULL);
-  osname = ostree_deployment_get_osname (merge_deployment);
-  origin = ostree_deployment_get_origin (merge_deployment);
-  booted = ostree_deployment_get_csum (merge_deployment);
+  return g_object_ref (booted_deployment);
+}
+
+static gchar *
```
cosimoc: Does this really need to duplicate the string? It looks to me that the only client that needs this duplicated (the hash table code below) could do it there instead.

src/eos-updater-util.c
```diff
@@ -160,127 +153,136 @@ static const gchar *dmi_attributes[] =
...
                            "Not an ostree system");
-      goto out;
+      return NULL;
     }
 
-  merge_deployment = ostree_sysroot_get_merge_deployment (sysroot, NULL);
-  osname = ostree_deployment_get_osname (merge_deployment);
-  origin = ostree_deployment_get_origin (merge_deployment);
-  booted = ostree_deployment_get_csum (merge_deployment);
+  return g_object_ref (booted_deployment);
```
cosimoc: Does this really need to add a ref? It does not look like the clients really have that requirement

src/eos-updater-util.c
```diff
@@ -291,17 +293,25 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
     g_hash_table_insert (hw_descriptors, g_strdup ("product_name"),
                          g_strdup ("EOSUNKNOWN"));
 
-  vendor = g_strdup (g_hash_table_lookup (hw_descriptors, "sys_vendor"));
-  product = g_strdup (g_hash_table_lookup (hw_descriptors, "product_name"));
-  p_id = g_strconcat (cleanstr (vendor), " ", cleanstr (product), NULL);
-  message ("Product group: %s", p_id);
+  *sys_vendor = g_strdup (g_hash_table_lookup (hw_descriptors, "sys_vendor"));
+  *product_name = g_strdup (g_hash_table_lookup (hw_descriptors, "product_name"));
+  return hw_descriptors;
```
cosimoc: I wonder if you really need to also return these values here separately, instead of looking them up from the table where you need them

src/eos-updater-util.c
```diff
@@ -313,73 +323,180 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
...
+}
+
+static gboolean
+process_single_group (GKeyFile *bkf,
+                      const gchar *group_name,
+                      gboolean *on_hold,
+                      gchar **p_ref,
+                      GError **error)
+{
+  if (g_key_file_get_boolean (bkf, group_name, ON_HOLD_KEY, NULL))
```
cosimoc: Pass the error here

src/eos-updater-util.c
```diff
@@ -313,73 +323,180 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
...
-          message ("No product-specific configuration and %s is on hold, "
-                   "nothing to upgrade here", DEFAULT_GROUP);
-          on_hold = TRUE;
-          ret = TRUE;
-          goto out;
-        }
-      p_ref = g_key_file_get_string (bkf, DEFAULT_GROUP, OSTREE_REF_KEY, error);
-      if (!p_ref)
-        goto out;
+      if (!process_single_group (bkf, group_name, on_hold, p_ref, error))
```
cosimoc: Should be `DEFAULT_GROUP` instead of `group_name`

src/eos-updater-util.c
```diff
@@ -313,73 +323,180 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
...
+
+static void
+maybe_send_metric (const gchar *vendor,
+                   const gchar *product,
+                   const gchar *ref,
+                   gboolean on_hold)
+{
+  static gboolean metric_sent = FALSE;
+
+  if (G_LIKELY (metric_sent))
```
cosimoc: We usually don't use `G_LIKELY` and `G_UNLIKELY` because they make the code more complicated and their usefulness is questionable... Please remove

src/eos-updater-util.c
```diff
@@ -313,73 +323,180 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
...
+  message ("Product group: %s", product_group);
+
+  g_hash_table_insert (hw_descriptors, g_strdup ("ref"), g_strdup (booted_ref));
+  g_hash_table_insert (hw_descriptors, g_strdup ("commit"), get_booted_checksum (booted_deployment));
+  bkf = download_branch_file (baseurl, hw_descriptors, error);
+  if (bkf == NULL)
+    return FALSE;
+
+  if (!process_branch_file (bkf, product_group, &on_hold, &ref, error))
+    return FALSE;
```
cosimoc: Add newline

src/eos-updater-util.c
```diff
@@ -313,73 +323,180 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
...
+        {
+          message ("No product-specific branch configuration or %s found, "
+                   "following the origin file", DEFAULT_GROUP);
+          ref = g_strdup (booted_ref);
+        }
+
+      message ("Using product branch %s", ref);
+      *upgrade_refspec = g_strdup_printf ("%s:%s", booted_remote, ref);
+      *original_refspec = g_strdup_printf ("%s:%s", booted_remote, booted_ref);
+    }
```
cosimoc: Add newline

src/eos-updater-util.c
```diff
@@ -313,73 +323,180 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
...
+  gs_unref_object OstreeDeployment *booted_deployment = NULL;
+
+  g_return_val_if_fail (OSTREE_IS_REPO (repo), FALSE);
+  g_return_val_if_fail (upgrade_refspec != NULL, FALSE);
+  g_return_val_if_fail (original_refspec != NULL, FALSE);
+
+  booted_deployment = get_booted_deployment (error);
+  if (booted_deployment == NULL)
+    return FALSE;
+  return get_upgrade_info (repo,
```
cosimoc: Add newline

src/eos-updater-util.h
```diff
@@ -37,21 +37,29 @@
 G_BEGIN_DECLS
 
 #define EOS_UPDATER_ERROR (eos_updater_error_quark())
-GQuark eos_updater_error_quark (void);
+GQuark
```
cosimoc: It's normal for headers to have the return type on the same line in GObject style code.

src/eos-updater.c
```diff
@@ -85,6 +85,7 @@ on_bus_acquired (GDBusConnection *connection,
 
   /* We are deliberately not emitting a signal here. This
    * isn't a state change, it's our initial state.
+   * krnowak: ORLY? It calls eos_updater_emit_state_changed…
```
cosimoc: lol

src/eos-updater-apply.c
```diff
@@ -78,7 +78,6 @@ apply (GTask *task,
   GMainContext *task_context = g_main_context_new ();
   const gchar *update_id = eos_updater_get_update_id (updater);
   const gchar *update_refspec = eos_updater_get_update_refspec (updater);
-  const gchar *orig_refspec = eos_updater_get_original_refspec (updater);
```
cosimoc: I don't really understand this commit - the code is not exporting the property just to propagate it, but it "stores" it locally on the `GDBusProxy` object not differently from how you store it in the updater data. The property is separately exported on the bus for convenience of other possible clients (but I don't believe it is in use now).

src/eos-updater-util.c
krnowak: `ostree_sysroot_get_booted_deployment` does not transfer the ownership, so if we don't ref it, the booted deployment might be deleted when we unref the sysroot. It used to work without adding a ref because sysroot was actually leaking its deployments earlier (https://github.com/ostreedev/ostree/pull/287).

src/eos-updater-util.c
krnowak: The `eos_updater_get_booted_checksum` also uses this function.

src/eos-updater-util.c
krnowak: I wanted to contain the use of the magic strings like "sys_vendor" and "product_name" just in one place, but I guess I can hide those names behind some `#define`.

src/eos-updater-util.c
krnowak: Oops, thanks for spotting.

src/eos-updater-util.c
krnowak: Oops. Will fix.

src/eos-updater-util.c
krnowak: Sure.

src/eos-updater-util.h
krnowak: I had an impression that the types were always in a separate line in gobject, but I see I was wrong. Will revert it.

src/eos-updater-apply.c
krnowak: Hm, I remember seeing in a review that introduced those changes that exporting this property was just a side-effect of adding a property which was meant to be used only internally (read: there was no `EosUpdaterData` struct then, which has this very purpose).

But sure, if you think that exporting this property might be useful for client then I'll just remove this commit.

src/eos-updater-util.c
krnowak: Hm, looks like we have to handle it a bit more specially - the OnHold key might be missing and that's fine, we don't want to propagate the error in this case, right?

krnowak:
Updated, addressed some of the comments.

src/eos-updater-apply.c
cosimoc: Yeah that was my review I think :-)
@dbnicholson convinced me that this property could be useful for other clients, so I'd rather leave it in.

src/eos-updater-util.c
cosimoc: Oh, good point.

src/eos-updater-util.c
```diff
@@ -313,73 +323,193 @@ eos_updater_resolve_upgrade (EosUpdater  *updater,
...
+                        G_KEY_FILE_ERROR_KEY_NOT_FOUND))
+    {
+      g_propagate_error (error, local_error);
+      return FALSE;
+    }
+  g_clear_error (&local_error);
+  *p_ref = g_key_file_get_string (bkf, group_name, OSTREE_REF_KEY, error);
+  if (*p_ref == NULL)
+    return FALSE;
+  return TRUE;
```
cosimoc: I think you should set `*on_hold = FALSE` before returning `TRUE` here. When the error is not set, all the out parameters should always be filled.

src/eos-updater-apply.c
dbnicholson: I'll be honest that I originally did it because I wasn't sure the best way to store private data and that was the fastest way to make the property available. However, I do think it's useful for the client to see that this has changed, and it's also useful for debugging.

src/eos-updater-util.c
```diff
@@ -160,148 +153,165 @@ static const gchar *dmi_attributes[] =
...
+  if (!ostree_repo_remote_get_url (repo, osname, &url, error))
+    return NULL;
+
+  return g_steal_pointer (&url);
+}
 
-  if (!ostree_parse_refspec (o_refspec, &o_remote, &o_ref, error))
-    goto out;
+#define EOS_VENDOR_KEY "sys_vendor"
+#define EOS_PRODUCT_KEY "product_name"
```
cosimoc: I would define this without and `EOS` prefix... maybe `DMI_VENDOR_KEY` etc?

src/eos-updater-data.h
```diff
@@ -0,0 +1,43 @@
...
+G_BEGIN_DECLS
+
+typedef struct EosUpdaterData EosUpdaterData;
+
+struct EosUpdaterData
+{
+  OstreeRepo *repo;
+};
+
+void
```
cosimoc: These should also be in the same line as the declaration

cosimoc:
@krnowak Done another round of review, thanks.

krnowak:
Looks like github is acting up. Whenever I update my avahi branch which is based on this cleanups branch, it marks most of the comments as obsolete. Meh.

src/eos-updater-util.c
krnowak: I'll rename it to `VENDOR_KEY`. The value also originates from device-tree stuff, so it is not really DMI-specific.

src/eos-updater-util.c
krnowak: Sure.

krnowak:
Updated. I removed the commit that moved the `original_refspec` variable to `EosUpdaterData` and added a small commit quieting down some make rules (so they print `GEN foo` when running `make V=0`).

PR #43, opened 2016-10-28
Disable all update steps on live sessions

wjt:
eos-autoupdater.timer is already stopped and masked from the initramfs,
but this prevents any updates being fetched or installed.

The rationale here is that endless.live_boot sessions have a read-only
filesystem, with overlay+tmpfs mounted over directories which need to be
read-write. Running an ostree update on such a system is at best
wasteful and frustrating (it will be discarded as soon as you reboot, so
can never be used), and at worst disastrous (it fills /run, because
hardlinking files on overlay requires copying the file from the lower to
the upper filesystem, and your live system breaks until you reboot).

https://phabricator.endlessm.com/T13951

### DIFF
--- a/src/eos-updater-live-boot.c
+++ b/src/eos-updater-live-boot.c
@@ -1,0 +1,56 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*-
+ *
+ * Copyright © 2016 Endless Mobile
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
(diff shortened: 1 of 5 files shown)


src/eos-updater-live-boot.c
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Author: Will Thompson <wjt@endlessm.com>
+ */
+
+#include "eos-updater-live-boot.h"
+
+gboolean
+is_live_boot (void)
+{
+  const gchar *force = g_getenv ("EU_FORCE_LIVE_BOOT");
```
wjt: (I'm open to replacing this with a command-line argument, or removing it entirely.)

PR #46, opened 2017-01-24
T14894 merge debian

pwithnall:
Merge the `debian-master` branch into `master` so that the `debian` directory can be maintained in sync with everything else in the project. This is justified because we are the upstream for eos-updater, and we know it’s only going to be used on EOS.

---

When merging this, I suspect some changes will be needed in Jenkins; and the `debian-master` branch will need to be deleted.


dbnicholson:
Looks good. I'll disable the master jenkins job before merging so we can get it in cleanly.

dbnicholson:
Going now in http://ci.endlessm-sf.com:8080/job/obs-eos-updater-master/60/console. I deleted the debian-master branch from the repo and changed the jenkins job not to run builds on the origin/debian-master branch.

If this needs to go back to eos3.1, we'll also need to do the dance there.

pwithnall:
> If this needs to go back to eos3.1, we'll also need to do the dance there.

I don’t think it does.

Thanks for the fast merge.

PR #47, opened 2017-01-25
T14894 kinvolk

pwithnall:
PR to merge the work by Kinvolk on adding LAN and USB support for `eos-updater`. The commits by Krzesimir have all been reviewed by Kinvolk and myself, so only need a cursory review — although their use of `/ostree/repo/extensions` for storing new metadata could do with some investigation. It’s probably best if I write this up in the task before anyone looks at it though. The commits by me all need review. They are either fixes for problems I’ve found through code review of the branch; or fixes for problems I found while reading the `eos-updater` code to get more familiar with it. I have deliberately not squashed my fixes into the commits from Kinvolk so they can be reviewed separately. However, I have rebased them to be in the right place in the commit history as much as I could.

I am still in the process of testing this manually, and I am aware that one of the recent commits (e5cbfacb) broke `make check`. I will fix that tomorrow morning (did some debugging tonight and am not really sure what’s going on yet). I have not yet got LAN updates working with my local testing, although they may be closer to working with the latest patches on the branch — I haven’t yet had a chance to re-test due to `make check` failing.

Accordingly, this cannot be merged yet; but will hopefully be mergeable by the end of the week. So if you could time your reviews accordingly, that would be great. Thanks!

---

https://phabricator.endlessm.com/T14894


pwithnall:
Pushed some more changes to [drop `eos-branch` file support](https://phabricator.endlessm.com/T14894#329859). There’s a bit of follow-up cleanup to do, but no more functional changes to make in this area as far as I can tell. I will try and do the cleanup tomorrow and push additional commits, but they shouldn’t affect pending reviews much.

pwithnall:
Finished pushing commits to drop `eos-branch` file support.

src/eos-updater-apply.c
```diff
@@ -120,42 +161,74 @@ apply (GTask *task,
...
+                            error))
+    return FALSE;
+  g_clear_object (&data->extensions);
+
+  head_commit_timestamp = g_date_time_new_from_unix_utc (ostree_commit_get_timestamp (update_commit));
+
+  if (!eos_avahi_generate_service_file (repo,
+                                        head_commit_timestamp,
+                                        error))
+    return FALSE;
```
dbnicholson: I don't think that failures in `eos_extensions_save` or `eos_avahi_generate_service_file` should propagate here. This code is about whether the update was applied, not whether it could be served to someone else. A failure from this function returns to the user saying that their system could not be upgraded, which is not the case here. I'd turn these into warnings or find some other way to hook into the upgrade path.

src/eos-updater-fetch.c
```diff
@@ -72,6 +74,31 @@ update_progress (OstreeAsyncProgress *progress,
...
+  g_variant_builder_init (&builder, G_VARIANT_TYPE ("a{sv}"));
+  g_variant_builder_add (&builder, "{s@v}", "refs",
+                         g_variant_new_variant (g_variant_new_strv (&ref, 1)));
+  if (url_override != NULL)
+    g_variant_builder_add (&builder, "{s@v}", "override-url",
+                           g_variant_new_variant (g_variant_new_string (url_override)));
+
+  options = g_variant_ref_sink (g_variant_builder_end (&builder));
+  return ostree_repo_pull_with_options (self, remote_name, options,
+                                        progress, cancellable, error);
```
dbnicholson: When we pull from the Endless server, we rely on HTTPS to guarantee we're we can trust the content. For that reason, we don't bother to setup `gpg-verify-summary=true` in the remote configuration. Maybe we should do that anyway (flatpak does), but in the case of LAN or USB updates, I think it's very important to do that since ostree won't otherwise validate the summary file and we'll have a broken chain of trust.

You can actually pass that in as a boolean variant option to `ostree_repo_pull_with_options`. The option is called `gpg-verify-summary`. I think the thing to do is always pass that in as `TRUE` unless the config explicitly disables it. Something like:
```
gboolean gpg_verify_summary;
/* Get gpg-verify-summary setting, defaulting to TRUE */
if (!ostree_repo_get_remote_boolean_option (repo, remote_name, "gpg-verify-summary", TRUE, &gpg_verify_summary, error)
    /* Handle error reading configuration */
g_variant_builder_add (&builder, "{s@v}", "gpg-verify-summary", g_variant_new_variant (g_variant_new_boolean (gpg_verify_summary)));
```

pwithnall:
Those review comments all make sense, thanks. Let me know when you’re done reviewing and I’ll push some fixup commits for the comments. Then those can be reviewed separately, squashed in, and the branch merged. That’s probably the clearest way of doing it.

src/eos-prepare-volume.c
```diff
@@ -0,0 +1,162 @@
...
+
+  raw_usb_path = remaining[0];
+  usb_path = g_file_new_for_commandline_arg (raw_usb_path);
+  if (!g_file_query_exists (usb_path, NULL))
+    {
+      return fail (quiet, "Path ‘%s’ does not exist", raw_usb_path);
+    }
+
+  sysroot = ostree_sysroot_new_default ();
+  if (!ostree_sysroot_load (sysroot, NULL, &error))
```
dbnicholson: I think the sysroot should be locked before continuing so that the source repo isn't updated while the objects and refs are pulled into the dest repo.

src/eos-prepare-usb-update.c
```diff
@@ -0,0 +1,665 @@
...
+                         g_variant_new_variant (g_variant_new_strv (&refspec->ref, 1)));
+  options = g_variant_ref_sink (g_variant_builder_end (&builder));
+  // ostree remote add --repo=dir/eos-update --no-gpg-verify eos https://endless:upgtnFSxLoDfUJnt@ostree.endlessm.com/ostree/eosfree-i386 eos2/i386
+  if (!ostree_repo_remote_add (usb_repo,
+                               refspec->remote,
+                               url,
+                               options,
+                               cancellable,
+                               error))
+    return FALSE;
```
dbnicholson: Adding the remote to the USB repo seems a little superfluous, but I guess this gives you the opportunity to later update the USB repository independently.

src/eos-prepare-usb-update.c
```diff
@@ -0,0 +1,665 @@
...
+  g_autoptr(GError) local_error = NULL;
+  g_autoptr(EosPullData) pull_data = NULL;
+
+  if (!get_server_and_uri (source_repo,
+                           refspec,
+                           cancellable,
+                           &server,
+                           &server_uri,
+                           error))
+    return FALSE;
```
dbnicholson: What's the purpose of spinning up a server to pull from? We can just pass in the path to the source repo and do the equivalent of `pull-local`, which will avoid all the HTTP overhead. Oh, I guess this is because the server will be a fake archive-z2 repo. But if the USB preparation is running at root, the pull can be done directly. OSTree handles this fine as long as the source repo uri scheme is `file`.

src/eos-prepare-usb-update.c
```diff
@@ -0,0 +1,665 @@
...
+                         "refs",
+                         g_variant_new_variant (g_variant_new_strv (refs, 1)));
+  g_variant_builder_add (&builder,
+                         "{s@v}",
+                         "override-commit-ids",
+                         g_variant_new_variant (g_variant_new_strv (override_commit_ids, 1)));
+  g_variant_builder_add (&builder,
+                         "{s@v}",
+                         "depth",
+                         g_variant_new_variant (g_variant_new_int32 (0)));
```
dbnicholson: I think that this pull should set the mirroring option in the pull flags. Then the refs will be local in the USB repo and the client can pull directly from it without putting some magic refs/remotes -> refs/heads translator in front of it. Like this:
```
OstreeRepoPullFlags pullflags = OSTREE_REPO_PULL_FLAGS_MIRROR;
g_variant_builder_add (&builder, "{s@v}", "flags", g_variant_new_variant (g_variant_new_int32 (pullflags)));
```
But maybe I'm really missing how the USB repo is handled as a source.

src/eos-prepare-usb-update.c
dbnicholson: The server also does the refs/remotes -> refs/heads translation so you can pull from the system repo as if it's a remote. So, maybe this needs to stay, unfortunately.

src/eos-updater-apply.c
pwithnall: I need to think about this some more, but I am wondering if using an Avahi `.service` file is still actually the best option. There are no parts of the updater which run continuously apart from Avahi — `eos-autoupdater` runs on a timer, `eos-updater` only runs when the state machine changes state, and `eos-update-server` only runs when socket-activated on an incoming HTTP connection.

Unless we could use path-based activation (`man systemd.path`) to run a little service whenever `/ostree/repo/refs/remotes` changes, which would update the Avahi `.service` file then exit?

src/eos-repo-server.c
```diff
@@ -0,0 +1,910 @@
...
+  return g_steal_pointer (&dup);
+}
+
+static gboolean
+generate_faked_config (OstreeRepo *repo,
+                       GBytes **out_faked_config_contents,
+                       GError **error)
+{
+  g_autoptr(GKeyFile) config = key_file_copy (ostree_repo_get_config (repo),
+                                              error);
```
pwithnall: Fixed in 36cba02.

src/eos-updater-fetch.c
pwithnall: Fixed in e8fd158 and 64f9633.

data/Makefile.am
```diff
@@ -1,21 +1,28 @@
...
 
 systemdsystemunit_DATA = $(eos_autoupdater_in:%.in=%)
 
-sysconf_DATA = eos-updater.conf
-
-eos-autoupdater.service: eos-autoupdater.service.in
-	$(AM_V_GEN)$(edit) $< >$@
+sysconf_DATA = \
+	eos-updater.conf \
+	eos-updater-daemon.conf \
```
pwithnall: Fixed in 5e6df13, c29741b, 570156f, 15306aa, cb773e1.

src/eos-updater-poll.c
```diff
@@ -17,199 +18,279 @@
...
 
- out:
-  if (error)
+  download_order_strv = g_key_file_get_string_list (config,
+                                                    DOWNLOAD_GROUP,
+                                                    ORDER_KEY,
+                                                    NULL,
+                                                    error);
+  if (download_order_strv == NULL)
+    return FALSE;
```
pwithnall: I think it should error out here, since the user has provided an invalid configuration. With c29741b, there is now an example configuration file installed in `/usr/share/eos-updater/etc/eos-updater.conf` which the user can copy, and which is loaded if no configuration file exists in `/etc`, but not if loading an existing file fails.

src/eos-updater-avahi.c
```diff
@@ -0,0 +1,659 @@
...
+      "<service-group>\n"
+      "  <name replace-wildcards=\"yes\">EOS update service on %%h</name>\n"
+      "  <service>\n"
+      "    <type>%s</type>\n"
+      "    <port>%" G_GUINT16_FORMAT "</port>\n"
+      "    <txt-record>eos_txt_version=%s</txt-record>\n"
+      "%s"
+      "  </service>\n"
+      "</service-group>\n",
+      type_escaped, port, txt_version_escaped, txt_records_str);
```
pwithnall: It is correct to advertise the service if `eos-update-server` is not running, since `eos-update-server` is socket-activated.

See my comments above for my thoughts so far about this. I haven’t made any changes to fix this yet, and will think about it some more tomorrow.

src/eos-autoupdater.c
```diff
@@ -30,13 +50,17 @@ typedef enum _UpdateStep {
...
+/* These must be kept in sync with #UpdateStep. */
+#define UPDATE_STEP_FIRST UPDATE_STEP_NONE
+#define UPDATE_STEP_LAST UPDATE_STEP_APPLY
+
 #define SEC_PER_DAY (3600ll * 24)
 
 /* This file is touched whenever the updater starts */
-#define UPDATE_STAMP_DIR        "/var/lib/eos-updater"
-#define UPDATE_STAMP_NAME       "eos-updater-stamp"
+static const char *UPDATE_STAMP_DIR = LOCALSTATEDIR "/eos-updater";
```
pwithnall: Nope: `debian/rules` passes `--localstatedir=/var/lib` to `configure`. IIRC this is the ‘other’ (non-GNU) default value for `localstatedir`. But if you can point me to a good precedent, I can easily move the `/lib` from `debian/rules` to `eos-autoupdater.c`.

src/eos-autoupdater.c
```diff
@@ -421,86 +512,145 @@ is_connected_through_mobile (void)
...
   GError *error = NULL;
-  gint update_interval;
+  guint update_interval_days;
   gboolean update_on_mobile;
   gboolean force_update = FALSE;
   GOptionContext *context;
 
   GOptionEntry entries[] = {
     { "force-update", 0, 0, G_OPTION_ARG_NONE, &force_update, "Force an update", NULL },
+    { "from-volume", 0, 0, G_OPTION_ARG_STRING, &volume_path, "Poll for updates from the volume", "PATH" },
```
pwithnall: It’s from commit 397b3040:
> Allow autoupdater to run updates from volume
>
> This adds the --from-volume flag that takes a string. Using this flag
> makes autoupdater to call PollVolume instead of Poll.

I have no idea of the motivation for this. Maybe to allow machines to be configured to always poll a mounted network volume (on a local server) for updates, rather than the network?

pwithnall:
Here’s one round of review fixes. I have not tested them thoroughly yet, but the code should be ready to review. I’ll test them tomorrow, and get the rest of the fixes done.

src/eos-updater-apply.c
pwithnall: Another alternative is to use [`ConditionNeedsUpdate=`](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#ConditionArchitecture=) to run a small service to update the Avahi `.service` file whenever `/usr` is updated with respect to `/var` and `/etc` (i.e. an update is applied). I think I prefer the `.path`-file-based option, though, since whether an update is applied is slightly decoupled from which objects and refs we actually have in `/ostree/repo` (and hence can serve to LAN peers).

src/eos-prepare-usb-update.c
pwithnall: Yeah. I guess we could do a `pull-local` and that should copy in the refs as `/refs/heads`, but that’s adding another code path which needs to be tested — whereas reusing the HTTP server is just existing code. Preparing a USB stick is not a time-critical operation, so let’s leave this as it is for now.

Added some clarifying comments in 82c2f39 and 6cedd29.

src/eos-prepare-usb-update.c
pwithnall: Do the mirroring options do anything aside from changing how refs are handled? I think the intention for the USB repo is that it can be used just like a remote, so the same fetch-phase code is used for pulling from all polled sources (main, LAN, USB).

src/eos-prepare-usb-update.c
pwithnall: Yeah, I don’t see how it can hurt. Let’s leave it in.

src/eos-prepare-volume.c
pwithnall: Fixed as 3a8c939.

src/eos-updater-fetch.c
pwithnall: Actually, I think the Kinvolk branch does this already, and we *don’t* want to enable `gpg-verify-summary`, because that means that pulling from a LAN peer requires the peer to have a valid `/ostree/repo/summary` file — which it might not have. The `summary` file is not pulled from the server by default, and even if it is, the user could locally nuke it using `ostree summary` (though that’s unlikely).

The code they have instead is in `eos-updater-poll-common.c`, in `commit_checksum_from_*()`, called by `fetch_commit_checksum()`. It works its way through the priority list of `refs.d`, `eos-summary`, original `summary`, and verifies them and tries to look up the given ref in them. The verification code looks the same as what `ostree.git` uses when you pass `gpg-verify-summary`.

The repository mirroring code currently copies the `refs.d` and `eos-summary` files (and their signatures). If no `eos-summary` file is available on the remote, but a normal `summary` file is, the `extensions->summary = …` assignment at the bottom of `commit_checksum_from_any_summary()` should ensure it gets copied to the mirror as `eos-summary`. Same for all the signatures.

pwithnall: (Spotted because the two commits I mentioned above break the unit tests; I think it’s better to revert them, but I’d appreciate your thoughts.)

src/eos-updater-apply.c
pwithnall: I think the `.path`-file based approach is probably the best way to go, with `PathChanged=/ostree/repo/refs/heads/ostree/1/1/0; PathChanged=/etc/eos-updater/eos-update-server.conf; PathChanged=/usr/share/eos-updater/etc/eos-update-server.conf`. It would run an `eos-updater-avahi` service to update the Avahi `.service` file.

`/etc/eos-updater/eos-update-server.conf` would be a key file containing a single option: whether to enable serving of updates over the LAN. We could then enable `eos-update-server.socket` by default. If the option is disabled, `eos-update-server` would bail as soon as it’s run; and the `eos-updater-avahi` would remove the Avahi `.service` file.

The problem I see with this approach is that `eos-updater-avahi.service` and `eos-update-server.socket` have to have the same enabled/disabled state. It’s no good if one of them is enabled and the other isn’t. We can’t trivially roll `eos-updater-avahi` into `eos-update-server`, because the latter may run for a very long time, and once it’s running, it’s running (so we can’t just update the Avahi `.service` file on startup). We’d end up reimplementing all the `.path` checks in code. I guess this can be solved by adding appropriate `Also=` and `Requires=` dependencies between the two units.

pwithnall: However, I don’t want to start implementing this without some feedback on the proposal, so I guess that will happen on Monday if you provide feedback tonight.

I don’t think this is a blocker for getting the branch merged though — it’s something I would work on as soon as it’s merged, but the current behaviour should be good enough for merging, right?

src/eos-updater-fetch.c
pwithnall: Reverted as 15cd6a1, b583027, 0931365.

src/eos-updater-avahi.c
pwithnall: Ah, also, Avahi does claim to use inotify to monitor its `/etc/avahi/services` directory.

pwithnall:
Pushed another round of fixes, which I think addresses all the comments except the Avahi `.service` stuff, which I’m going to punt to a follow-up task; and some of the USB updater stuff which is lower priority, and the questions are less serious.

src/eos-updater-apply.c
pwithnall: Filed https://phabricator.endlessm.com/T15544 as a follow-up task for this.

src/eos-updater-fetch.c
```diff
@@ -93,6 +95,21 @@ repo_pull (OstreeRepo *self,
...
+   * disabled. */
+  if (!ostree_repo_get_remote_boolean_option (self, remote_name,
+                                              "gpg-verify-summary", TRUE,
+                                              &gpg_verify_summary,
+                                              &child_error))
+    {
+      g_warning ("Failed to read gpg-verify-summary option from repository: %s",
+                 child_error->message);
+      gpg_verify_summary = TRUE;
+    }
```
dbnicholson: `ostree_repo_get_remote_boolean_option` won't return an error on the option not being, so I feel like this might mask real errors in the repo. But probably safest to keep going instead of bombing here.

src/eos-updater-poll-common.c
```diff
@@ -313,6 +313,8 @@ get_repo_pull_options (const gchar *url_override,
                          g_variant_new_variant (g_variant_new_int32 (OSTREE_REPO_PULL_FLAGS_COMMIT_ONLY)));
   g_variant_builder_add (&builder, "{s@v}", "refs",
                          g_variant_new_variant (g_variant_new_strv (&ref, 1)));
+  g_variant_builder_add (&builder, "{s@v}", "gpg-verify-summary",
+                         g_variant_new_variant (g_variant_new_boolean (TRUE)));
```
dbnicholson: Why isn't this doing the same "respect if the config has `gpg-verify-summary=false`" dance as the other commit?

src/eos-updater.c
```diff
@@ -219,6 +221,62 @@ listen_on_session_bus (void)
...
 }
 
+/* Remove our configuration files from /etc which are identical to the current
+ * versions installed in /usr/share or /usr/etc, which we embed as an MD5
+ * checksum. If we do this on all systems, we can eventually change the formats
+ * in /usr/etc without worrying about the new defaults being overwritten by
+ * stale files in /etc.
+ *
+ * This functionality can be removed after a few releases, once we’re confident
+ * all systems will have been upgraded. */
```
dbnicholson: For what it's worth, if you remove the file from /etc in the new ostree and the local version is unchanged from the old ostree, then it will be removed. It's only when there are local modifications that it will hang around.

That said, this is a nice thing for the service to do.

data/Makefile.am
```diff
@@ -7,7 +7,9 @@ eos_autoupdater_in = \
 
 systemdsystemunit_DATA = $(eos_autoupdater_in:%.in=%)
 
-sysconf_DATA = \
+# Example configuration files.
+sysconfexampledir = $(datadir)/$(PACKAGE_NAME)/etc
```
dbnicholson: I think `$(pkgdatadir)` is more concise than `$(datadir)/$(PACKAGE_NAME)`.

src/eos-autoupdater.c
```diff
@@ -61,6 +63,7 @@ static const char *UPDATE_STAMP_DIR = LOCALSTATEDIR "/eos-updater";
 static const char *UPDATE_STAMP_NAME = "eos-updater-stamp";
 
 static const char *CONFIG_FILE_PATH = SYSCONFDIR "/eos-updater.conf";
+static const char *STATIC_CONFIG_FILE_PATH = DATADIR "/" PACKAGE_NAME "/etc/eos-updater.conf";
```
dbnicholson: Bikeshedding - I don't think the additional `etc/` subdir is necessary. There's nothing in the pkgdatadir right now, and I can't see us accumulating enough files where the additional subdirectory is useful for disambiguation.

dbnicholson: Oh, and if you do want a separate subdir, I'd probably call it `conf` or `config` rather than carrying forward the `etc` misnomer.

src/eos-autoupdater.c
```diff
@@ -62,8 +62,9 @@ typedef enum _UpdateStep {
 static const char *UPDATE_STAMP_DIR = LOCALSTATEDIR "/eos-updater";
 static const char *UPDATE_STAMP_NAME = "eos-updater-stamp";
 
-static const char *CONFIG_FILE_PATH = SYSCONFDIR "/eos-updater.conf";
-static const char *STATIC_CONFIG_FILE_PATH = DATADIR "/" PACKAGE_NAME "/etc/eos-updater.conf";
+static const char *CONFIG_FILE_PATH = SYSCONFDIR "/" PACKAGE_NAME "/eos-autoupdater.conf";
```
dbnicholson: 👍

src/eos-prepare-volume.c
```diff
@@ -137,7 +137,11 @@ main (int argc,
     }
 
   sysroot = ostree_sysroot_new_default ();
-  if (!ostree_sysroot_load (sysroot, NULL, &error))
+
+  /* Lock the sysroot so it can’t be updated while we’re pulling from it. The
+   * lock is automatically released when we finalise the sysroot. */
+  if (!ostree_sysroot_lock (sysroot, &error) ||
```
dbnicholson: Just FYI, this will block indefinitely until the lock is acquired. I think that's completely fine right now, but I wanted to make sure you were aware of that. You can also use `ostree_sysroot_try_lock` for non-blocking that will fail immediately if the lock can't be acquired or `ostree_sysroot_lock_async` to have complete non-blocking control.

src/eos-updater-fetch.c
pwithnall: I think it was appropriate to put a `g_warning()` there, then. However, this code is dropped in a follow-up commit.

src/eos-updater-poll-common.c
pwithnall: Oops. However, this code is dropped in a follow-up commit.

src/eos-updater.c
pwithnall: Ah, I didn’t realise the three-way merge handled deletion as well. Nice.

data/Makefile.am
pwithnall: Fixed as cdd1888, 021155b.

src/eos-autoupdater.c
pwithnall: Fair point. Dropped the subdir in cdd1888, 021155b.

src/eos-prepare-volume.c
pwithnall: Yup. I’ve just pushed 8da448e to the branch to print a message if blocking.

src/eos-updater-poll-common.c
```diff
@@ -313,8 +313,6 @@ get_repo_pull_options (const gchar *url_override,
                          g_variant_new_variant (g_variant_new_int32 (OSTREE_REPO_PULL_FLAGS_COMMIT_ONLY)));
   g_variant_builder_add (&builder, "{s@v}", "refs",
                          g_variant_new_variant (g_variant_new_strv (&ref, 1)));
-  g_variant_builder_add (&builder, "{s@v}", "gpg-verify-summary",
-                         g_variant_new_variant (g_variant_new_boolean (TRUE)));
```
dbnicholson: Why did you revert this?

src/eos-updater-fetch.c
```diff
@@ -95,21 +93,6 @@ repo_pull (OstreeRepo *self,
...
-   * disabled. */
-  if (!ostree_repo_get_remote_boolean_option (self, remote_name,
-                                              "gpg-verify-summary", TRUE,
-                                              &gpg_verify_summary,
-                                              &child_error))
-    {
-      g_warning ("Failed to read gpg-verify-summary option from repository: %s",
-                 child_error->message);
-      gpg_verify_summary = TRUE;
-    }
```
dbnicholson: Why did you revert this?

src/eos-prepare-usb-update.c
```diff
@@ -418,6 +421,40 @@ run_pull_task_func (GTask *task,
...
+  summary_sig_path = g_build_filename (repo_path, "summary.sig", NULL);
+
+  if (!g_file_set_contents (summary_path,
+                            g_bytes_get_data (summary, NULL),
+                            g_bytes_get_size (summary),
+                            &local_error))
+    {
+      g_task_return_error (task, g_steal_pointer (&local_error));
+      return;
+    }
```
dbnicholson: I don't think this should be necessary when doing a mirror pull. OSTree will copy the summary and signature itself - https://github.com/endlessm/ostree/blob/master/src/libostree/ostree-repo-pull.c#L3129. And I think a mirror pull is what's wanted here - then the refs on the USB volume are under refs/heads and you don't need to do anything special with it.

src/eos-autoupdater.c
dbnicholson: Huh, I've never heard of `localstatedir` being anything but `/var` or `$prefix/var`. Certainly autoconf never does anything besides defaulting to `$prefix/var` - http://git.savannah.gnu.org/gitweb/?p=autoconf.git;a=blob;f=lib/autoconf/general.m4;hb=HEAD#l588. Can you point me to the description of the non-GNU localstatedir definition?

I'm not sure why we'd care about a non-GNU default, either. The intention for this code (and all Endless code as far as I know) has always been to follow the GNU/FHS standards. The code had unfortunately hardcoded `/var/lib/eos-updater` before, but to me the logical fix there is to pull in the `/var` prefix rather than assuming someone would pass in a non-standard `localstatedir` with `/lib` suffixed.

src/eos-updater-avahi.c
dbnicholson: OK, I didn't know avahi monitored that with inotify. I'll follow up with any further discussion about this handling in https://phabricator.endlessm.com/T15544.

src/eos-updater-fetch.c
dbnicholson: OK, I think I confused myself about the summary vs eos-summary handling when pulling from eos-update-server. I think you're right here and nothing is needed.

I'd just remove the commits and subsequent reverts. Since this code isn't pushed yet, that just adds noise to the history, IMO.

src/eos-updater-apply.c
dbnicholson: Yeah, I agree that reworking the details for how the server is handled can be deferred. However, I think my original comment that these updates should be non-fatal still stands. I'd hate for our users not getting updates because of a bug in a path for an orthogonal service.

src/eos-updater-poll.c
dbnicholson: OK.

src/eos-updater-fetch.c
pwithnall: As I said earlier, I intend to rebase and squash the fixups and reverts into the branch appropriately immediately before merging. I’ve deliberately kept them separate for the moment so github doesn’t helpfully delete the review comments and discussion from those commits.

src/eos-updater-avahi.c
pwithnall: 👍

src/eos-prepare-usb-update.c
dbnicholson: The mirroring mode does a couple things:

1. If not refs to pull are specified, then all refs on the remote are pulled. That's not (to my knowledge) the case here as we're specifying refs.

2. Refs are pulled into the local namespace (refs/heads) rather than the remote namespace (refs/remotes).

3. The summary file and signature are copied from upstream when they're available.

4. If you're mirroring into an archive-z2 repo, static deltas are skipped as you want to pull the exact objects rather than rebuilding them with deltas.

src/eos-autoupdater.c
pwithnall: Hrm, I think it would be better to use `sharedstatedir` here, and set it to `/var/lib` (which Debian does by default, so we don’t need to do anything). The FHS defines `/var/lib` as [variable state information](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s08.html), which is what we need here.

pwithnall: Fixed as 2de5a70.

src/eos-updater-apply.c
pwithnall: Ah, I missed that. Fixed as 8c7ad28.

src/eos-prepare-usb-update.c
dbnicholson: The most important thing here is for 2, which then won't require a smart server to sit in front of the repo and translate refs/remotes to refs/heads.

src/eos-updater-poll-common.c
dbnicholson: Saw your comment above now.

src/eos-updater-fetch.c
dbnicholson: Same.

src/eos-prepare-usb-update.c
pwithnall: # Doesn’t apply, as you say.
# Already handled by the HTTP server code, though it would be harmless to have the refs in `refs/heads` anyway, since the server code tries that first.
# Already handled by the `EosExtensions` stuff which creates `refs.d`. Again, it would be harmless for `summary{,.sig}` to be present.
# Could be useful.

I’ll add the option and give it a test.

src/eos-prepare-usb-update.c
dbnicholson: However, if the repo has the eos-summary and eos-summary.sig files, those should probably be preferred and put at summary and summary.sig locations.

src/eos-autoupdater.c
```diff
@@ -59,7 +59,7 @@ typedef enum _UpdateStep {
 #define SEC_PER_DAY (3600ll * 24)
 
 /* This file is touched whenever the updater starts */
-static const char *UPDATE_STAMP_DIR = LOCALSTATEDIR "/eos-updater";
+static const char *UPDATE_STAMP_DIR = SHAREDSTATEDIR "/eos-updater";
```
dbnicholson: I don't think this is right. The autoconf default for sharedstatedir is `$prefix/com` (http://git.savannah.gnu.org/gitweb/?p=autoconf.git;a=blob;f=lib/autoconf/general.m4;hb=HEAD#l587). Likewise, debhelper does no overriding of sharedstatedir (https://anonscm.debian.org/cgit/debhelper/debhelper.git/tree/Debian/Debhelper/Buildsystem/autoconf.pm#n29).

So, this does introduce a functional change moving the stamp directory from `/var/lib/eos-updater` to `/usr/com/eos-updater`.

dbnicholson:
I've finished another round of review. My only remaining comments:

* I think using `sharedstatedir` for the autoupdater stamp file location is wrong. See https://github.com/endlessm/eos-updater/pull/47/commits/2de5a704ea46601a3ce3924a60e5436b2569451b#r100831258.

* I think the USB volume should be prepped with a mirror pull so that the refs are in refs/heads.

* If the USB volume is prepped in mirror mode, then copying the summary file is unnecessary. However, if the source repo has eos-summary and eos-summary.sig, then those should probably be put in the standard location since that would carry through the upstream summary signature. Might be able to punt on this for now.

src/eos-autoupdater.c
dbnicholson: Fair, you overrode it in debian/rules. Still, I don't understand what's wrong with `LOCALSTATEDIR "/lib/eos-updater"`. Even dpkg and apt do it this way - https://anonscm.debian.org/cgit/dpkg/dpkg.git/tree/configure.ac#n55, https://anonscm.debian.org/cgit/apt/apt.git/tree/CMakeLists.txt#n195. They both use localstatedir of `/var` in debian/rules (dpkg manually, and apt via dh_auto_configure).

dbnicholson: Oh, I guess you didn't override it. I can't see how sharedstatedir is set to /var/lib anywhere in autoconf or debhelper.

dbnicholson: Are you sure you did this build with your latest changes? Here's what `./debian/rules build` says for me:
```
gcc -DHAVE_CONFIG_H -I. -I..  -DOSTREE_WITH_AUTOCLEANUPS -DSYSCONFDIR=\""/etc"\" -DSHAREDSTATEDIR=\""/usr/com"\" -DDATADIR=\""/usr/share"\" -DPKGDATADIR=\""/usr/share/eos-updater"\"   -Wdate-time -D_FORTIFY_SOURCE=2 -fno-strict-aliasing -Wall -Wextra -Wundef -Wnested-externs -Wwrite-strings -Wpointer-arith -Wmissing-declarations -Wmissing-prototypes -Wstrict-prototypes -Wredundant-decls -Wno-unused-parameter -Wno-missing-field-initializers -Wdeclaration-after-statement -Wformat=2 -Wold-style-definition -Wcast-align -Wformat-nonliteral -Wformat-security -Wsign-compare -Wstrict-aliasing -Wshadow -Winline -Wpacked -Wmissing-format-attribute -Wmissing-noreturn -Winit-self -Wmissing-include-dirs -Wunused-but-set-variable -Warray-bounds -Wimplicit-function-declaration -Wreturn-type -Wswitch-enum -Wswitch-default -Werror -Wno-suggest-attribute=format -Wno-error=unused-parameter -Wno-error=missing-field-initializers  -pthread -I/usr/include/gio-unix-2.0/ -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -pthread -I/usr/include/libsoup-2.4 -I/usr/include/libxml2 -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -pthread -I/usr/include/ostree-1 -I/usr/include/gio-unix-2.0/ -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include  -pthread -D_REENTRANT -I/usr/include/gio-unix-2.0/ -I/usr/include/uuid -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/eosmetrics-0  -g -O2 -fPIE -fstack-protector-strong -Wformat -Werror=format-security -c -o eos_updater-eos-updater.o `test -f 'eos-updater.c' || echo './'`eos-updater.c
```
Notice the `-DSHAREDSTATEDIR=\""/usr/com"\"` on the command line.

src/eos-prepare-usb-update.c
pwithnall: Fixed in eeaef32.

src/eos-autoupdater.c
pwithnall: Apparently, I didn’t built with my latest changes. That’s quite embarrassing. Let’s go with the `localstatedir` precedent then. Fixed as df5e7ba (I’ll rewrite the commit message when squashing it into the other commit).

pwithnall:
> I've finished another round of review. My only remaining comments:
> * I think using sharedstatedir for the autoupdater stamp file location is wrong. See 2de5a70#r100831258.

Fixed as df5e7ba (I’ll update the original commit message when squashing it in).

> * I think the USB volume should be prepped with a mirror pull so that the refs are in refs/heads.

Fixed as eeaef32.

> * If the USB volume is prepped in mirror mode, then copying the summary file is unnecessary. However, if the source repo has eos-summary and eos-summary.sig, then those should probably be put in the standard location since that would carry through the upstream summary signature. Might be able to punt on this for now.

Fixed as 72681a7.

dbnicholson:
All looks good to me now. Do you still want review from others or should we merge?

pwithnall:
> All looks good to me now. Do you still want review from others or should we merge?

I think it’s OK to merge now. I’ll squash the fixups, fix the commit messages, and merge it myself. Thanks for your patience. :-)

PR #50, opened 2017-02-14
T14894 attributes

pwithnall:
https://phabricator.endlessm.com/T14894


pwithnall:
> I'm curious what other files are immutable

It was each of the commit directories in the deployment, yes.

PR #51, opened 2017-02-16
T15545 configuration for builder

pwithnall:
Partial solution for T15545.

https://phabricator.endlessm.com/T15545


Makefile.am
```diff
@@ -10,3 +13,39 @@ AM_DISTCHECK_CONFIGURE_FLAGS = --with-systemdsystemunitdir='$${libdir}/systemd/s
 pkgconfigdir = $(libdir)/pkgconfig
 
 pkgconfig_DATA = eos-updater-0.pc
+
+# Utility library
+noinst_LTLIBRARIES += libeos-updater-util/libeos-updater-util-@EUU_API_VERSION@.la
```
cosimoc: Do you really need this to be versioned?

Makefile.am
```diff
@@ -10,3 +13,39 @@ AM_DISTCHECK_CONFIGURE_FLAGS = --with-systemdsystemunitdir='$${libdir}/systemd/s
...
+
+libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_CFLAGS = \
+	$(WARN_CFLAGS) \
+	$(CODE_COVERAGE_CFLAGS) \
+	$(GIO_CFLAGS) \
+	$(SOUP_CFLAGS) \
+	$(OSTREE_CFLAGS) \
```
cosimoc: Nitpick: alphabetical order

Makefile.am
```diff
@@ -10,3 +13,39 @@ AM_DISTCHECK_CONFIGURE_FLAGS = --with-systemdsystemunitdir='$${libdir}/systemd/s
...
+	-DPKGDATADIR=\""$(pkgdatadir)"\" \
+	$(NULL)
+libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_LDFLAGS = \
+	$(WARN_LDFLAGS) \
+	$(NULL)
+libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_LIBADD = \
+	$(CODE_COVERAGE_LIBS) \
+	$(GIO_LIBS) \
+	$(SOUP_LIBS) \
+	$(OSTREE_LIBS) \
```
cosimoc: Nitpick: alphabetical order

libeos-updater-util/tests/Makefile.am
```diff
@@ -0,0 +1,43 @@
+include $(top_srcdir)/glib-tap.mk
+
+installed_testdir = $(libexecdir)/installed-tests/libeos-updater-util-@EUU_API_VERSION@
+installed_test_metadir = $(datadir)/installed-tests/librhosydd-@EUU_API_VERSION@
```
cosimoc: I think there's a copy-paste typo here

libeos-updater-util/tests/config.c
```diff
@@ -0,0 +1,262 @@
...
+                                                        "key-file-unreadable",
+                                                        NULL);
+  g_file_set_contents (fixture->key_file_unreadable_path, "[Test]\nFile=3",
+                       -1, &error);
+  g_assert_no_error (error);
+  g_assert_cmpint (g_chmod (fixture->key_file_unreadable_path, 0200), ==, 0);
+
+  fixture->key_file_invalid_path = g_build_filename (fixture->tmp_dir,
+                                                     "key-file-invalid",
+                                              NULL);
```
cosimoc: Weird indentation

src/eos-autoupdater.c
```diff
@@ -66,6 +66,7 @@ static const char *UPDATE_STAMP_NAME = "eos-updater-stamp";
 static const char *CONFIG_FILE_PATH = SYSCONFDIR "/" PACKAGE "/eos-autoupdater.conf";
 static const char *OLD_CONFIG_FILE_PATH = SYSCONFDIR "/eos-updater.conf";
 static const char *STATIC_CONFIG_FILE_PATH = PKGDATADIR "/eos-autoupdater.conf";
+static const char *VAR_CONFIG_FILE_PATH = LOCALSTATEDIR "/local/" PACKAGE "/eos-autoupdater.conf";
```
ramcq: This should be /usr/local rather than /var - the fact that ends up in /var is an ostree implementation deal.

ramcq: So yeah it would be either /usr/local/etc/pkg-specific-path... or /usr/local/share/pkg-specific-...

cosimoc: @ramcq I am not sure the updater code should be looking in `/usr`. My understanding is that OSTree mounts (or used to mount) contents added to `/usr/local` into `/var/usrlocal` on the deployed system.

In general though, for similar situations we usually add directly files to subdirectories of `/var` from the image builder instead of relying on OSTree implementation details, so I think @pwithnall's approach is correct; I was only suggesting that we avoid `/var/local/$package_name` because that does not seem consistent with the layout of `/var/local`.

ramcq: /usr/local is purely a symlink to /var/local when you've booted from ostree - so this will never really make a difference in Endless - but the "correct" approach I was imagining was an XDG_DATA_DIR style search order like /etc/foo, /usr/local/etc/foo, /usr/etc/foo. Referring to /var for configuration just upsets my Debian sensibilities. It's ostree's business to move stuff around there.

You are totally right that, as /var/local is in fact just /usr/local relocated, it should have the same sub-heirachy as /usr. I suggest /usr/local/etc because we have the ostree convention of our "default" /etc being in /usr/etc, so searching /usr/local/etc isn't a bad shout.

ramcq: Actually I just realised the other one is in PKGDATADIR not /usr/etc, so the equiv /usr/local/share/... makes more sense.

Makefile.am
pwithnall: Strictly, no. However, it doesn’t hurt, and I think it’s better to add versioning (and all the other library trimmings) in from the beginning rather than adding them in half-way through the life of a library and omitting something. I plan to slowly add versioning to the other libraries in the project.

Makefile.am
pwithnall: Fair. I kept them in this order because they were copied verbatim from `src/Makefile.am`, but since enough of the other build rules changed in the transition, I can reorder this list. I will keep `WARN_CFLAGS` and `CODE_COVERAGE_CFLAGS` at the top, as it’s possible (though really really unlikely) that one of the dependency `_CFLAGS` variables would need to override them.

Makefile.am
pwithnall: OK. See my comments about `*_CFLAGS`.

libeos-updater-util/tests/Makefile.am
pwithnall: Whoops, fixed. This also means I haven’t updated the Debian packaging properly.

libeos-updater-util/tests/config.c
pwithnall: Thanks, fixed.

src/eos-autoupdater.c
pwithnall: This mess stems from the fact I have no Debian sensibilities and have so far avoided reading the FHS. I’ll change it to `/usr/local/share/eos-updater/*.conf`.

pwithnall:

src/eos-autoupdater.c
```diff
@@ -65,6 +66,7 @@ static const char *UPDATE_STAMP_NAME = "eos-updater-stamp";
 static const char *CONFIG_FILE_PATH = SYSCONFDIR "/" PACKAGE "/eos-autoupdater.conf";
 static const char *OLD_CONFIG_FILE_PATH = SYSCONFDIR "/eos-updater.conf";
 static const char *STATIC_CONFIG_FILE_PATH = PKGDATADIR "/eos-autoupdater.conf";
+static const char *LOCAL_CONFIG_FILE_PATH = PREFIX "/local/share/" PACKAGE "/eos-autoupdater.conf";
```
pwithnall: Using `PREFIX` means it works when build on Debian (`--prefix=/usr`) but when built under JHBuild (`--prefix=/`) we end up with `/local/share/eos-updater/…`, which isn’t great. I can’t think of a decent way of handling this without adding another `configure.ac` argument which essentially replicates `--prefix` and pretends you never changed it from `/usr/local` though.

src/eos-autoupdater.c
cosimoc: @ramcq I would still prefer not to depend on OSTree implementation detail here, take the most straightforward implementation and just use the exact same file path in the image builder and on the client side; but at the end of the day it's bikeshedding and don't feel like arguing it.
I will merge the PR using this path for now.

src/eos-autoupdater.c
cosimoc: Yeah, that's another reason why I don't think this file should end up inside `/usr`...

src/eos-autoupdater.c
ramcq: There is great precedent for checking both /usr/local and /usr - eg XDG_DATA_DIRS. Assuming /usr is read-only and using a non-standard /var path instead would be introducing dependencies on the ostree implementation detail. 

The assumption that the image builder is relying on, that /usr/local will end up in /var/local is the ostree-specific thing - but in fact the image builder also creates the symlink so the hook could also act on /usr/local and the file would end up in the right place.

But in the end... neither the updater nor the image builder can be expected to be that independent of ostree... so it doesn't matter all that much as long as it works!

PR #53, opened 2017-02-20
T15644 not ostree notification

pwithnall:
An update to make failure due to running eos-updater on a dev-converted system (i.e. one which doesn’t use OSTree and hence can’t be updated by eos-updater) more obvious, by giving it its own error code.

Also fix a few small issues with `eos-updater-ctl` I saw in the process.

https://phabricator.endlessm.com/T15644


pwithnall:
The eos-desktop part of this is [PR #234](https://github.com/endlessm/eos-desktop/pull/234).

PR #72, opened 2017-03-07
eos-updater-avahi: Correctly handle service resolution failure

pwithnall:
Previously, the code was assuming that service resolution always
succeeds, but that’s not the case. If it fails, the address is returned
as NULL.

See the example code:
http://www.avahi.org/doxygen/html/client-browse-services_8c-example.html#a21

https://phabricator.endlessm.com/T15948

### DIFF
--- a/src/eos-updater-avahi.c
+++ b/src/eos-updater-avahi.c
@@ -158,10 +158,22 @@ queue_callback (EosAvahiDiscoverer *discoverer)
                                                         "eos updater avahi callback");
 }
 
+static gboolean
+should_queue_callback (EosAvahiDiscoverer *discoverer)
+{
+  return (discoverer->state == EOS_AVAHI_RESOLVING_ONLY &&
+          discoverer->callback_id == 0 &&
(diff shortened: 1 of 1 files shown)


src/eos-updater-avahi.c
```diff
@@ -180,6 +184,31 @@ queue_error_callback (EosAvahiDiscoverer *discoverer,
...
+  va_list args;
+
+  if (discoverer->state != EOS_AVAHI_RESOLVING_ONLY)
+    return;
+
+  if (discoverer->callback_id > 0)
+    return;
+
+  if (g_hash_table_size (discoverer->discovered_services) > 0)
+    return;
```
cosimoc: These three conditions already exist in `maybe_queue_success_callback()`. If you refactor them in e.g. a `should_queue_callback()` function, this one could be just implemented as the composition of that and `queue_error_callback()`.

pwithnall: Good point. Fixed.

PR #75, opened 2017-03-08
T15975 autoupdater jitter

pwithnall:
https://phabricator.endlessm.com/T15975

### DIFF
--- a/data/eos-autoupdater.conf
+++ b/data/eos-autoupdater.conf
@@ -10,4 +10,5 @@
 #  3 - apply updates
 LastAutomaticStep=3
 IntervalDays=14
+RandomizedDelayDays=0
 UpdateOnMobile=false
--- a/src/eos-autoupdater.c
+++ b/src/eos-autoupdater.c
@@ -111,12 +111,15 @@ get_stamp_dir (void)
  * harm in the stamp file not being updated: it just means we’re going to check
  * again for updates sooner than otherwise. */
 static void
-update_stamp_file (guint randomized_delay_days)
+update_stamp_file (guint update_interval_days,
+                   guint randomized_delay_days)
 {
   const gchar *stamp_dir = get_stamp_dir ();
   g_autofree gchar *stamp_path = NULL;
   g_autoptr(GFile) stamp_file = NULL;
   g_autoptr(GError) error = NULL;
+  GTimeVal mtime;
+  g_autofree gchar *next_update = NULL;
 
   if (g_mkdir_with_parents (stamp_dir, 0755) != 0) {
     int saved_errno = errno;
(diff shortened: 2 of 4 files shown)


src/eos-autoupdater.c
```diff
@@ -140,6 +141,41 @@ update_stamp_file (void)
...
+                           "PRIORITY=%d", LOG_CRIT,
+                           "MESSAGE=Failed to get stamp file info: %s", error->message,
+                           NULL);
+          return;
+        }
+
+      g_get_current_time (&mtime);
+      actual_delay_days = g_random_int_range (0, randomized_delay_days + 1);
+      mtime.tv_sec += actual_delay_days * SEC_PER_DAY;
+      g_file_info_set_modification_time (file_info, &mtime);
```
cosimoc: As commented on Slack, let's add a comment here that we're setting an mtime in the future.

data/eos-autoupdater.conf
```diff
@@ -10,4 +10,5 @@
 #  3 - apply updates
 LastAutomaticStep=3
 IntervalDays=14
+RandomizedDelayDays=7
```
cosimoc: Instead of changing the default behavior, let's only add the delay in the deployments where we need it (i.e. those that we turn on LAN configuration for)

ramcq: Hmm... I'm not sure if the Random delay is so bad - we "hope" that we distribute our updates over a 2 week period, but we don't actually do anything to ensure this or counteract factors like computers which are off during the week, or off during the weekend, which will cause clustering. Generally, configuration changes between the default and the deployments introduce risk, because it's code paths we will inevitably test less. Is there anything to worry about having this switched on usually?

src/eos-autoupdater.c
pwithnall: Done.

data/eos-autoupdater.conf
pwithnall: Done. I’ve filed a PR for the configuration change as https://github.com/endlessm/eos-image-builder/pull/296.

pwithnall:

data/eos-autoupdater.conf
pwithnall: Mid-air collision. I missed @ramcq’s comment.

@ramcq: @cosimoc was concerned that normal-internet-connected computers would no longer immediately get updates after being installed.

ramcq: Ah... I see. Makes sense, unless we wanted always randomise the interval always, and have a separate preference as to whether we update "right away" on boot, or wait the random interval before the first update.

pwithnall: I can’t think of a situation where you’d want one and explicitly not want teh other.

cosimoc: I just don't want to introduce any policy changes for everyone as part of this. If you think there's merit in a different approach for regular users, please open a separate story.

PR #77, opened 2017-03-10
T15999 eos updater ctl enable server

pwithnall:
https://phabricator.endlessm.com/T15999


pwithnall:
Pushed an updated version which changes the `geteuid()` check.

PR #80, opened 2017-03-23
eos-repo-server: Reset timeout conditions on failed requests

pwithnall:
Previously they were reset only on successful requests, meaning that a
long-running timed out request in the middle of a legitimate stream of
update requests from a client would cause eos-update-server to exit.

https://phabricator.endlessm.com/T16270

### DIFF
--- a/src/eos-repo-server.c
+++ b/src/eos-repo-server.c
@@ -403,20 +403,29 @@ eos_filez_read_data_disconnect_and_clear_msg (EosFilezReadData *read_data)
 
 static void
 update_pending_requests (EosUpdaterRepoServer *server,
-                         guint new_value)
+                         gint                  delta)
 {
-  if (server->pending_requests == new_value)
+  GObject *obj = G_OBJECT (server);
+
+  g_assert (delta >= 0 || server->pending_requests >= (guint) -delta);
+  g_assert (delta <= 0 || server->pending_requests <= G_MAXUINT - delta);
+
+  if (delta == 0)
     return;
 
-  server->pending_requests = new_value;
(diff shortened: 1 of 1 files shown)


src/eos-repo-server.c
```diff
@@ -830,9 +831,6 @@ server_cb (SoupServer *soup_server,
   g_object_get (msg,
                 "status-code", &status_code,
                 NULL);
```
cosimoc: Does not look like `status_code` is used anymore.

src/eos-repo-server.c
```diff
@@ -403,20 +403,29 @@ eos_filez_read_data_disconnect_and_clear_msg (EosFilezReadData *read_data)
...
-                         guint new_value)
+                         gint                  delta)
 {
-  if (server->pending_requests == new_value)
+  GObject *obj = G_OBJECT (server);
+
+  g_assert (delta >= 0 || server->pending_requests >= (guint) -delta);
+  g_assert (delta <= 0 || server->pending_requests <= G_MAXUINT - delta);
+
+  if (delta == 0)
```
cosimoc: Any reason not to update `server->last_request_time` in this case?

src/eos-repo-server.c
pwithnall: Whoops.

src/eos-repo-server.c
pwithnall: Because the number of requests hasn’t changed, so there is no new ‘last request’ to change the time for.

pwithnall:
Updated branch to address review comments.

PR #82, opened 2017-03-31
eos-updater-avahi: Include scope ID in link-local IPv6 addresses

pwithnall:
Avahi doesn’t do this for us, and a link-local IPv6 address is useless
without a scope ID (since the same address can be used for multiple
machines on different interfaces; the scope ID disambiguates the
interface).

This fixes failures to poll and pull from LAN peers over IPv6 with the
error message “Failed to fetch latest commit from …: Could not connect:
Invalid argument”.

https://phabricator.endlessm.com/T16385

### DIFF
--- a/src/eos-updater-avahi.c
+++ b/src/eos-updater-avahi.c
@@ -37,6 +37,7 @@
 #include <avahi-common/strlst.h>
 #include <avahi-glib/glib-malloc.h>
 #include <avahi-glib/glib-watch.h>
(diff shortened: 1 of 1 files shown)


sgnn7:
@pwithnall LGTM though I'd add the upstream bug URI in the commit message.

pwithnall:
Updated to include the URI, thanks. Pushed.

PR #84, opened 2017-04-05
autoupdater: Port to libnm

krnowak:
The libnm-glib library is deprecated.

https://phabricator.endlessm.com/T15813

Unfortunately `make check` does not pass for me neither on my branch nor on the current master branch. Will need to investigate.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -56,7 +56,7 @@ AC_DEFINE_UNQUOTED([EOS_AVAHI_PORT], [$port], [Socket activation port to be used
 GLIB_REQUIRED_VERSION=2.50.0
 OSTREE_REQUIRED_VERSION=2016.15
 AVAHI_REQUIRED_VERSION=0.6.31
-NM_REQUIRED_VERSION=0.9.8.6
+NM_REQUIRED_VERSION=1.2.0
 
 PKG_CHECK_MODULES([GIO],
                   [gio-unix-2.0 >= $GLIB_REQUIRED_VERSION])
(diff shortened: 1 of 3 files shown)


src/eos-autoupdater.c
```diff
@@ -595,9 +595,12 @@ is_online (void)
       return TRUE;
     }
 
-  client = nm_client_new ();
+  client = nm_client_new (NULL, &error);
   if (!client)
-    return FALSE;
+    {
+      g_message ("Failed to get network manager client: %s", error->message);
```
pwithnall: s/network manager/NetworkManager/ (and similarly below)

configure.ac
```diff
@@ -68,7 +68,7 @@ PKG_CHECK_MODULES([SOUP],
                   [libsoup-2.4])
 
 PKG_CHECK_MODULES([EOS_AUTOUPDATER],
-                  [libnm-glib >= $NM_REQUIRED_VERSION
+                  [libnm >= $NM_REQUIRED_VERSION
```
pwithnall: This dependency information also needs to be changed in `debian/control`.

krnowak:
Updated. I bumped the libnm dep to 1.2 (that's why there is a check for libnm 1.6 in autoupdater), because eos3.1 seems to have that version still, while master has 1.6. Will this change land in eos3.1 too? (Likely seeing that the 3.1 is on par with master.)

Also, the failing (hanging actually) test is the one testing the LAN updates.

pwithnall:
> Also, the failing (hanging actually) test is the one testing the LAN updates.

Fixed as https://phabricator.endlessm.com/T16387#366105.

PR #87, opened 2017-04-06
Test log fixes

krnowak:
Just some minor fixes I have made during checking the hangs in the LAN update tests.

### DIFF
--- a/tests/spawn-utils.c
+++ b/tests/spawn-utils.c
@@ -51,7 +51,6 @@ cmd_result_ensure_ok (CmdResult *cmd,
   g_autoptr(GError) local_error = NULL;
   g_autofree gchar *dump = cmd_result_dump (cmd);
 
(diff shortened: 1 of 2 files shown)


tests/spawn-utils.c
```diff
@@ -121,6 +119,14 @@ cmd_async_result_free (CmdAsyncResult *cmd)
   g_free (cmd);
 }
 
+static gchar *
+join_envp_for_logging (const gchar * const *envp)
+{
+  g_autofree gchar *envp_joined = g_strjoinv ("\n - ", merged_env);
```
pwithnall: s/`merged_env`/`envp`/, I think

krnowak: Ew, seems like I actually forgot to compile-check it.

krnowak:
Updated.

PR #88, opened 2017-04-07
T15812 - drop XSLT rules

krnowak:
https://phabricator.endlessm.com/T15812


src/Makefile.am
```diff
@@ -156,25 +156,20 @@ eos_updater_SOURCES = \
 
 dist_man8_MANS += docs/eos-updater.8
 
+eos_updater_dbus_if := com.endlessm.Updater
```
pwithnall: Probably doesn’t make sense to have this as a variable. Changing the interface name would be a major change to all of the codebase, so factoring it out adds no value, and makes things less greppable (for example, `git grep com.endlessm.Updater.service` doesn’t work).

src/Makefile.am
```diff
@@ -156,25 +156,20 @@ eos_updater_SOURCES = \
...
+dbussystemservice_DATA = $(srcdir)/$(eos_updater_dbus_if).service
 
-CLEANFILES = $(BUILT_SOURCES) $(dbusconf_DATA) $(dbussystemservice_DATA)
+CLEANFILES = $(BUILT_SOURCES)
 
 EXTRA_DIST = \
-	eos-updater-ifname.xslt \
-	eos-updater-policy.xslt \
-	eos-updater-service.xslt \
 	eos-updater.xml \
```
pwithnall: It would be useful to also install `eos-updater.xml` as `/usr/share/dbus-1/interfaces/com.endlessm.Updater.xml`. Please do that as a separate commit.

src/Makefile.am
```diff
@@ -156,25 +156,20 @@ eos_updater_SOURCES = \
...
 
+eos_updater_dbus_if := com.endlessm.Updater
+
 dbusconfdir = $(datadir)/dbus-1/system.d
 dbussystemservicedir = $(datadir)/dbus-1/system-services
 
-dbusconf_DATA = $(eos_updater_dbus_if).conf
-dbussystemservice_DATA = $(eos_updater_dbus_if).service
+dbusconf_DATA = $(srcdir)/$(eos_updater_dbus_if).conf
+dbussystemservice_DATA = $(srcdir)/$(eos_updater_dbus_if).service
```
pwithnall: Shouldn’t `dbusconf_DATA` and `dbussystemservice_DATA` both have a `dist_` prefix? Please double-check that `make distcheck` and building the Debian package works after applying these commits. (As I said on Slack, I use [`build-snapshot`](https://git.apertis.org/cgit/apertis-customizations.git/tree/development/build-snapshot) to do both.)

debian/control
```diff
@@ -23,7 +23,6 @@ Build-Depends:
  libsoup2.4-dev,
  libsystemd-dev,
  ostree,
- xsltproc,
```
pwithnall: This commit can be squashed into the main one. No need to commit to the `debian/` directory separately.

krnowak: I was doing separate commits for changes in the debian directory for the sake of "debian: " prefix in the commit message. :)

src/Makefile.am
krnowak: Will do.

src/Makefile.am
krnowak: Hm hm hm, `make distcheck` passed here, I did run it before filing this PR. I think these things are distributed automatically if they are not built and installed somewhere, but that sounds overly smart/magic, so I'll check it.

src/Makefile.am
krnowak: Alright.

krnowak:
Updated.

pwithnall:
Lovely, thanks.

PR #89, opened 2017-04-13
T16551 server multiple repos

pwithnall:
https://phabricator.endlessm.com/T16551


libeos-update-server/server.c
```diff
@@ -0,0 +1,401 @@
...
+ * eus_server_disconnect:
+ * @self: an #EusServer
+ *
+ * Disconnect the server and all its repositories from the underlying
+ * #SoupServer and its socket. Cancel all pending requests and stop handling
+ * any new ones.
+ *
+ * This does not call soup_server_disconnect() on the underlying #SoupServer.
+ *
+ * This is called automatically when the #EusServer is disposed.
```
krnowak: This is actually a lie. I think that the call to the `eus_server_disconnect` function should happen in dispose before `GPtrArray` holding the repos is unreffed.

eos-update-server/docs/eos-update-server.8
```diff
@@ -60,7 +64,7 @@ zero or a negative number is provided, no timeout is implemented and
 .\"
 .IP "\fB\-r\fP, \fB\-\-serve\-remote=\fP"
 Name of the OSTree remote to serve from the local repository. The remote must
-be configured in the local repository. This is indented to be used for testing.
+be configured in the local repository. This is intended to be used for testing.
```
krnowak: I like the commit message for this fix. ;)

eos-update-server/docs/eos-update-server.conf.5
```diff
@@ -41,6 +49,21 @@ Whether to advertise and distribute updates over the local network. This may be
 \fItrue\fP or \fIfalse\fP. If \fItrue\fP, \fBeos\-update\-server\fP(8) and
 \fBeos\-updater\-avahi\fP(8) are enabled; otherwise, they will both refuse to
 advertise or distribute updates.
+\"
+.SH [Repository 0–9] SECTION OPTIONS
```
krnowak: Did you consider different naming of the sections? I'd like to propose the scheme similar to what git uses for the remotes:
```
[Repository "/ostree/repo"]
RemoteName="eos"

[Repository "/some/other/path"]
RemoteName="foo"
```

The upside of this scheme is that removing some repository section from the configuration file won't involve updating all the other repository sections (like it would do in the currently proposed scheme - write 10 repository sections and then remove `[Repository 0]`.

Later:
From the code I can see that the ordering is not really enforced. Basically index is used for setting the root path to `/<index>` for non-zero indexes and to an empty string for the zero index for backward compatibility. So instead we could have:
```
[Repository "main"]
Path=/ostree/repo
RemoteName=eos

[Repository "foo"]
Path=/some/other/path
RemoteName="foo"
```

so for the "main" repo server would set an empty root path and for other it would be `/<name>`. But it's probably bikeshedding at this point.

libeos-update-server/config.c
```diff
@@ -0,0 +1,213 @@
...
+      NULL
+    };
+  const gchar * const override_paths[] =
+    {
+      config_file_path,
+      NULL
+    };
+  g_auto(GStrv) groups = NULL;
+  gsize n_groups, i;
+  gboolean _advertise_updates;
```
krnowak: Cue in some language lawyer blathering about identifiers starting with an underline being reserved for C language/standard library.

Maybe remove the leading underline from the variable names and rename the function out parameters to something like `out_advertise_updates` or something?

eos-update-server/docs/eos-update-server.conf.5
krnowak: Also, if we decide to stay with the numbering then do you have any ideas how to make the `[Repository 0-9]` format clearer? Currently my first thought was "only max 10 repos are supported?".

eos-update-server/eos-update-server.c
```diff
@@ -448,6 +431,44 @@ start_listening (SoupServer *server,
...
+{
+  g_autoptr(EusRepo) eus_repo = NULL;
+  g_autoptr(GError) error = NULL;
+
+  if (!ostree_repo_open (repo, NULL, &error))
+    {
+      GFile *path = ostree_repo_get_path (repo);
+      g_autofree gchar *path_str = g_file_get_path (path);
+
+      message ("OSTree repository at ‘%s’ could not be opened: %s",
```
krnowak: I'd like to drop this `message` macro in favor of just `g_message`. Could you use it here and everywhere else instead?

eos-update-server/eos-repo-server.c
```diff
@@ -172,6 +178,25 @@ eos_updater_repo_server_set_property (GObject *object,
       g_set_object (&server->repo, g_value_get_object (value));
       break;
 
+    case PROP_ROOT_PATH:
+      g_clear_pointer (&server->root_path, g_free);
+      server->root_path = g_value_dup_string (value);
+
+      g_assert (server->root_path != NULL);
+
+      /* Add a missing leading slash if the root path is non-empty.. */
```
pwithnall: Fixed.

eos-update-server/eos-repo-server.c
```diff
@@ -231,6 +257,25 @@ eos_updater_repo_server_class_init (EosUpdaterRepoServerClass *repo_server_class
                                           G_PARAM_STATIC_STRINGS);
 
   /**
+   * EosUpdaterRepoServer:root-path:
+   *
+   * Root path to handle requests underneath. Any requests for paths not
+   * underneath this root will result in a HTTP 404 status code.
+   *
+   * It must be the empty string, or a string starting with a `/` and not
+   * ending in a `/`.
```
pwithnall: Changed the comment to `It should be the empty…`.

eos-update-server/eos-repo-server.c
```diff
@@ -214,34 +227,55 @@ eos_updater_repo_server_set_property (GObject *object,
...
-  server->pending_requests = 0;
-  server->last_request_time = 0;
-  g_clear_object (&server->cancellable);
-  g_clear_pointer (&server->cached_config, g_bytes_unref);
-  g_clear_object (&server->repo);
+  if (self->server != NULL)
+    {
+      soup_server_remove_handler (self->server, self->root_path);
+      g_clear_object (&self->server);
+    }
```
pwithnall: Will fix in a follow-up commit.

eos-update-server/eos-repo-server.c
```diff
@@ -214,34 +227,55 @@ eos_updater_repo_server_set_property (GObject *object,
...
 {
-  EosUpdaterRepoServer *server = EOS_UPDATER_REPO_SERVER (object);
+  EosUpdaterRepoServer *self = EOS_UPDATER_REPO_SERVER (object);
 
-  g_free (server->cached_repo_root);
-  g_free (server->remote_name);
-  g_free (server->root_path);
+  g_free (self->cached_repo_root);
+  g_free (self->remote_name);
+  g_free (self->root_path);
```
pwithnall: Will fix in a follow-up commit.

Makefile.am
```diff
@@ -121,6 +126,72 @@ CLEANFILES += $(eos_updater_avahi_in:%.in=%)
 
 dist_man8_MANS += eos-updater-avahi/docs/eos-updater-avahi.8
 
+# Server utility library
```
pwithnall: I find that more confusing, since it splits a single evaluation scope across multiple files, so the included `Makefile.am`s end up using rules which are defined in a different file.

libeos-update-server/server.c
```diff
@@ -0,0 +1,401 @@
...
+}
+
+static void
+eus_server_constructed (GObject *object)
+{
+  EusServer *self = EUS_SERVER (object);
+
+  G_OBJECT_CLASS (eus_server_parent_class)->constructed (object);
+
+  g_signal_connect (self->server, "request-read", (GCallback) request_read_cb, self);
```
pwithnall: Good catch; we should check. We can’t create a `SoupServer` ourselves, since the caller is the one who needs to call `soup_server_listen()` on it.

libeos-update-server/server.c
```diff
@@ -0,0 +1,401 @@
...
+static void
+eus_server_dispose (GObject *object)
+{
+  EusServer *self = EUS_SERVER (object);
+
+  self->pending_requests = 0;
+  self->last_request_time = 0;
+  g_clear_pointer (&self->repos, g_ptr_array_unref);
+
+  if (self->server != NULL)
```
pwithnall: Not correct: `dispose()` can be called multiple times, and it must not crash the second time it’s called. (That’s really unlikely to happen, as it only needs to happen when getting rid of refcount loops, but it’s best to write all `dispose()` implementations robustly.)

libeos-update-server/server.c
pwithnall: Whoops!

eos-update-server/eos-update-server.c
pwithnall: Let’s handle that in T15811; if that branch lands first, I’ll update this one to use `g_message()`.

libeos-update-server/config.c
pwithnall: Uuurgh, blah. Fixed.

eos-update-server/docs/eos-update-server.conf.5
pwithnall: I decided to go with numbering in the root path (rather than a string-based name) because the root path needs to be put in the DNS-SD records, and hence needs to be as small as possible. As you noticed, there is no ordering (or contiguity) enforced.

I’ll see if I can improve the documentation. I realise `[Repository 0–9]` is not particularly clear. I could go with `[Repository 0-9+]` (regex-style)?

pwithnall:

libeos-update-server/repo.c
```diff
@@ -162,62 +162,104 @@ eos_updater_repo_server_get_property (GObject *object,
...
+      /* Add a missing leading slash if the root path is non-empty. */
+      if (self->root_path[0] != '\0' && self->root_path[0] != '/')
+        {
+          g_autofree gchar *tmp = g_strconcat ("/", self->root_path, NULL);
+          g_free (self->root_path);
+          self->root_path = g_steal_pointer (&tmp);
+        }
+
+      /* Drop any trailing slash. */
+      if (strlen (self->root_path) > 1 &&
```
krnowak: Hopefully the compiler is smart enough to optmize all those calls to strlen to a single one.

libeos-update-server/server.c
krnowak: Oh, you are right.

Makefile.am
```diff
@@ -121,6 +126,72 @@ CLEANFILES += $(eos_updater_avahi_in:%.in=%)
...
+noinst_LTLIBRARIES += libeos-update-server/libeos-update-server-@EUS_API_VERSION@.la
+
+libeos_update_server_libeos_update_server_@EUS_API_VERSION@_la_CPPFLAGS = \
+	$(AM_CPPFLAGS) \
```
pwithnall: Updated to include `G_LOG_DOMAIN` here.

Makefile.am
```diff
@@ -121,6 +126,72 @@ CLEANFILES += $(eos_updater_avahi_in:%.in=%)
...
+	libeos-update-server/repo.h \
+	libeos-update-server/server.c \
+	libeos-update-server/server.h \
+	$(NULL)
+
+# eos-update-server program
+libexec_PROGRAMS += eos-update-server/eos-update-server
+
+eos_update_server_eos_update_server_CPPFLAGS = \
+	$(AM_CPPFLAGS) \
```
pwithnall: Done.

Makefile.am
```diff
@@ -121,6 +126,72 @@ CLEANFILES += $(eos_updater_avahi_in:%.in=%)
...
+	libeos-update-server/config.c \
+	libeos-update-server/config.h \
+	libeos-update-server/repo.c \
+	libeos-update-server/repo.h \
+	libeos-update-server/server.c \
+	libeos-update-server/server.h \
+	$(NULL)
+
+# eos-update-server program
+libexec_PROGRAMS += eos-update-server/eos-update-server
```
pwithnall: Yeah, the historic installation locations of all the binaries are all over the place. This branch doesn’t change the installation location for `eos-update-server`. I don’t place a high priority on changing those locations at the moment, as it would undoubtedly cause breakage.

eos-update-server/docs/eos-update-server.conf.5
```diff
@@ -21,10 +21,19 @@ eos\-update\-server.conf — Endless OS OSTree Server Configuration
...
-whose keys are described below.
+The configuration file contains a single mandatory section,
+\fI[Local Network Updates]\fP, and zero or more optional sections,
+\fI[Repository 0–65535]\fP, whose keys are described below. Each
+\fI[Repository 0–65535]\fP section describes a repository to serve; if there are no
+such sections, the default \fI/ostree/repo\fP repository is served with its
+\fIRemoteName=\fP set to the value of the \fB\-\-serve\-remote=\fP option passed
+to \fBeos\-update\-server\fP(8). Each section must be numbered uniquely; its
+index forms the root of the HTTP paths for requesting content from the
+repository. The indexes don’t have to be ordered or contiguous, and up to 65536
```
pwithnall: True, done.

eos-update-server/eos-update-server.c
```diff
@@ -492,25 +516,48 @@ main (int argc, char **argv)
...
     }
 
-  repo = eos_updater_local_repo ();
-  server = eos_updater_repo_server_new (repo,
-                                        options.served_remote,
-                                        NULL,
-                                        &error);
-  if (server == NULL)
+  /* Set up the server and repositories. */
+  soup_server = soup_server_new (NULL, NULL);
```
pwithnall: I don’t think so.

eos-update-server/eos-update-server.conf
```diff
@@ -3,3 +3,9 @@
 # and edit it.
 [Local Network Updates]
 AdvertiseUpdates=false
+
+# Default repository configuration. Add more [Repository 0-9] sections to
```
pwithnall: Whoops, fixed.

libeos-update-server/config.c
```diff
@@ -0,0 +1,213 @@
...
+      if (repository_path == NULL)
+        return FALSE;
+
+      remote_name = g_key_file_get_string (config, groups[i], REMOTE_NAME_KEY,
+                                           error);
+
+      if (remote_name == NULL)
+        return FALSE;
+
+      g_ptr_array_add (repository_configs,
```
pwithnall: True. Done.

pwithnall:
Updated.

eos-update-server/eos-update-server.c
krnowak: Well, reading the code for soup_server_new and g_object_new_valist makes me believe it should be fine, but I'm not going to get religious about it…

Makefile.am
krnowak: Uh, that seems to be still missing.

Makefile.am
krnowak: Fair enough, can be done later.

Makefile.am
pwithnall: I don’t know what happened there. Definitely added now.

eos-update-server/eos-update-server.c
pwithnall: You’re right; I should have checked. Fixed.

krnowak:
There seem to be still one G_LOG_DOMAIN define missing in the main Makefile.am file for the libeos-update-server library. With that fixes, this PR should be good to be merged.

eos-update-server/eos-update-server.c
pwithnall: Actually, no, that results in a compilation failure:
```
/opt/endless/source/eos-updater/eos-update-server/eos-update-server.c: In function ‘main’:
/opt/endless/source/eos-updater/eos-update-server/eos-update-server.c:520:3: error: not enough variable arguments to fit a sentinel [-Werror=format=]
   soup_server = soup_server_new (NULL);
```

krnowak: Meh, ok. The annotation in the libsoup is likely wrong (there is no such thing for g_object_new, which looks similar).

Thanks for checking!

PR #90, opened 2017-04-18
Convert to structured logging

krnowak:
https://phabricator.endlessm.com/T15811

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -49,6 +49,7 @@ AM_CPPFLAGS = \
 	-DDATADIR=\""$(datadir)"\" \
 	-DPKGDATADIR=\""$(pkgdatadir)"\" \
 	-DPREFIX=\""$(prefix)"\" \
+	-DG_LOG_USE_STRUCTURED \
 	$(NULL)
 
 # Utility library
@@ -63,6 +64,7 @@ libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_CFLAGS = \
 	$(NULL)
 libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_CPPFLAGS = \
 	$(AM_CPPFLAGS) \
+	-DG_LOG_DOMAIN=\""libeos-updater-util"\" \
 	$(NULL)
 libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_LDFLAGS = \
 	$(WARN_LDFLAGS) \
@@ -92,6 +94,7 @@ libexec_PROGRAMS += eos-updater-avahi/eos-updater-avahi
 
 eos_updater_avahi_eos_updater_avahi_CPPFLAGS = \
 	$(AM_CPPFLAGS) \
+	-DG_LOG_DOMAIN=\""eos-updater-avahi"\" \
 	$(NULL)
 eos_updater_avahi_eos_updater_avahi_CFLAGS = \
 	$(WARN_CFLAGS) \
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -32,7 +32,7 @@ BUILT_SOURCES = \
 	eos-updater-generated.c \
 	$(NULL)
 
-libeos_updater_dbus_la_CPPFLAGS = $(common_cppflags)
+libeos_updater_dbus_la_CPPFLAGS = $(common_cppflags) -DG_LOG_DOMAIN=\""eos-updater-dbus"\"
 libeos_updater_dbus_la_CFLAGS = $(WARN_CFLAGS) $(CODE_COVERAGE_CFLAGS) $(GIO_CFLAGS)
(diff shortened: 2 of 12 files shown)


Makefile.am
```diff
@@ -63,6 +64,7 @@ libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_CFLAGS = \
 	$(NULL)
 libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_CPPFLAGS = \
 	$(AM_CPPFLAGS) \
+	-DG_LOG_DOMAIN=\""LibEosUpdaterUtil"\" \
```
pwithnall: Nitpick: I’d prefer this to be `libeos-updater-util` to match the library filename.

Makefile.am
```diff
@@ -92,6 +94,7 @@ libexec_PROGRAMS += eos-updater-avahi/eos-updater-avahi
 
 eos_updater_avahi_eos_updater_avahi_CPPFLAGS = \
 	$(AM_CPPFLAGS) \
+	-DG_LOG_DOMAIN=\""EosUpdaterAvahi"\" \
```
pwithnall: Nitpick: Similarly, I’d prefer this to be `eos-updater-avahi`, so it matches the binary name.

src/Makefile.am
```diff
@@ -32,7 +32,7 @@ BUILT_SOURCES = \
 	eos-updater-generated.c \
 	$(NULL)
 
-libeos_updater_dbus_la_CPPFLAGS = $(common_cppflags)
+libeos_updater_dbus_la_CPPFLAGS = $(common_cppflags) -DG_LOG_DOMAIN=\""EosUpdaterDBus"\"
```
pwithnall: Nitpick: `libeos-updater-dbus`.

Similarly for the other domains in the patch (I won’t leave more comments about this).

src/eos-autoupdater.c
```diff
@@ -105,6 +106,86 @@ get_stamp_dir (void)
...
+   * G_GNUC_PRINTF annotation that has a zero as the second parameter,
+   * so it suggests to use this attribute. */
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wsuggest-attribute=format"
+  gint message_length = g_vasprintf (&message, format, args);
+#pragma GCC diagnostic pop
+  const GLogField fields[] = {
+    { "MESSAGE", message, message_length },
+    { "MESSAGE_ID", msgid, EOS_UPDATER_MSGID_LENGTH },
+    { "PRIORITY", numeric_log_level, 1 },
```
pwithnall: Nitpick: It’s possible someone might (in future) use a log level like `"10"` which is not a single byte. I’d use `-1` for the length here in that case, which means it will be calculated using `strlen()` before being outputted.

src/eos-autoupdater.c
```diff
@@ -105,6 +106,86 @@ get_stamp_dir (void)
...
+    /* strlen(G_LOG_DOMAIN) should be folded into a constant at
+     * compilation time */
+    { "GLIB_DOMAIN", G_LOG_DOMAIN, (G_LOG_DOMAIN != NULL) ? strlen(G_LOG_DOMAIN) : 0 },
```
pwithnall: This `GLogField` needs to be included conditionally — you can’t set the `length` to `0` to pretend the field doesn’t exist. `0` indicates that a field contains an arbitrary pointer value which should be carried around unevaluated.

*Update*: Ah, you do exactly that below. Excellent.

Just use `-1` for the length; it will be calculated using `strlen()` nearer output time.

krnowak: I was rather wondering if I should rather drop all this dance around `G_LOG_DOMAIN` being NULL, since we set it during compilation, so the field could be always included. But for now I'll make it simpler like you suggested.

src/eos-autoupdater.c
krnowak: I had a look here - https://git.gnome.org/browse/glib/tree/glib/gmessages.c#n1402

It tells something about the priority being only 1 byte only with a following nul byte, but that sounds bogus - the link to the syslog manpage states no such restriction.

Will fix it to `-1`.

src/eos-autoupdater.c
pwithnall: > I was rather wondering if I should rather drop all this dance around `G_LOG_DOMAIN` being `NULL`, since we set it during compilation

That would be a very good option. Add an assertion which checks that it’s set (and is non-`NULL` and non-`""`).

Makefile.am
krnowak: Done.

Makefile.am
krnowak: Done.

src/Makefile.am
krnowak: Done. For the rest too.

src/eos-autoupdater.c
krnowak: Went with `G_STATIC_ASSERT` checks for `G_LOG_DOMAIN`.

PR #91, opened 2017-04-18
eos-autoupdater: Don’t normally write stamp file on first boot

pwithnall:
Only write it if a randomised delay is configured, otherwise our first
check for updates will be delayed by the IntervalDays period. When we do
write it, discount the normal IntervalDays period so that the update
will be performed after the RandomizedDelayDays period *only*.

https://phabricator.endlessm.com/T16643

### DIFF
--- a/src/eos-autoupdater.c
+++ b/src/eos-autoupdater.c
@@ -192,8 +192,9 @@ info (const gchar *msgid,
  * harm in the stamp file not being updated: it just means we’re going to check
  * again for updates sooner than otherwise. */
 static void
-update_stamp_file (guint update_interval_days,
-                   guint randomized_delay_days)
+update_stamp_file (guint64 last_successful_update_secs,
+                   guint   update_interval_days,
+                   guint   randomized_delay_days)
 {
   const gchar *stamp_dir = get_stamp_dir ();
   g_autofree gchar *stamp_path = NULL;
@@ -201,6 +202,7 @@ update_stamp_file (guint update_interval_days,
   g_autoptr(GError) error = NULL;
   GTimeVal mtime;
   g_autofree gchar *next_update = NULL;
+  g_autoptr(GFileInfo) file_info = NULL;
 
   if (g_mkdir_with_parents (stamp_dir, 0755) != 0) {
     int saved_errno = errno;
@@ -212,7 +214,8 @@ update_stamp_file (guint update_interval_days,
     return;
   }
 
-  g_get_current_time (&mtime);
+  mtime.tv_sec = last_successful_update_secs;
+  mtime.tv_usec = 0;
 
   stamp_path = g_build_filename (stamp_dir, UPDATE_STAMP_NAME, NULL);
   stamp_file = g_file_new_for_path (stamp_path);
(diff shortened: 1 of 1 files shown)


src/eos-autoupdater.c
```diff
@@ -212,7 +213,8 @@ update_stamp_file (guint update_interval_days,
     return;
   }
 
-  g_get_current_time (&mtime);
+  mtime.tv_sec = last_successful_update_secs;
```
krnowak: Hm, we should add an `else` case to the `if (randomized_delay > 0)` that sets the mtime of the stamp file to `last_successful_update_secs`, otherwise the mtime of the stamp will be now when there is no randomized delay, instead of `last_successful_update_secs`, no? Earlier it wasn't a problem, I think, because g_get_current_time() was called around the same time when the contents of the timestamp were updated.

src/eos-autoupdater.c
```diff
@@ -599,10 +601,29 @@ is_time_to_update (guint update_interval_days,
...
+     * update. */
+    if (randomized_delay_days > 0)
+      {
+        guint64 last_successful_update_secs;
+
+        g_debug ("Not time to update, due to stamp file not being present, but %s is set to %u days.",
+                 RANDOMIZED_DELAY_KEY, randomized_delay_days);
+        last_successful_update_secs = g_get_real_time () / G_USEC_PER_SEC;
+        if (last_successful_update_secs >= update_interval_days * SEC_PER_DAY)
+          last_successful_update_secs -= update_interval_days * SEC_PER_DAY;
```
krnowak: This probably will break in the following scenario (let's assume that the current interval is one week):
- eos-autoupdater runs for the first time, creates a stamp file with mtime set to now minus the interval, so mtime is basically from week ago
- I modify the configuration file, so the interval is two times longer (from one week to two weeks)
- eos-autoupdater runs again, finds out that the next update should be perfomed next week, skip updates

If you deem this scenario to be too strained, then fine, the code looks ok. Otherwise you could consider doing something like:
```
if (stamp_file_does_not_exist)
  {
    if (randomized_delay > 0)
      {
        now = g_get_real_time ();
        first_update_time  = now + randomized_delay_days;
        /* special case - normally the stamp is empty,
         * but we write a byte to let eos-autoupdater
         * know that we should update ASAP + randomized delay
         * without looking at interval days */
        fwrite ("1", 1, 1, stamp_file);
        set_mtime (stamp_file, first_update_time);
        is_time_to_update = FALSE;
      }
    else
      {
        is_time_to_update = TRUE;
      }
  }
else
  {
    if (length_of_stamp_file > 0)
      {
        is_time_to_update = mtime_of_stamp_file < g_get_real_time ();
      }
    else
      {
        is_time_to_update = mtime_of_stamp_file + interval_days < g_get_real_time ();
      }
  }
```

(the pseudocode above blatantly ignores the issues with time units)

pwithnall: > If you deem this scenario to be too strained, then fine, the code looks ok

I think that scenario is too strained. If the user edits the configuration, any changes to make the `IntervalDays` shorter will result in an update being correctly run the next time `eos-autoupdater` is triggered. Any edits to make it longer will result in an additional delay before the first update, as you say, but that’s still in the same direction as what the user wants (delaying updates for a longer period). It’s unlikely to impact on dev testing, since `IntervalDays` is normally edited down to `0` (not up) by developers testing `eos-updater`. I don’t think the extra complexity is worth it.

src/eos-autoupdater.c
pwithnall: I’m not sure what you mean. In the `(randomized_delay_days == 0)` case, an update is scheduled immediately, and the `.stamp` file is not written until the update has completed successfully. What problem are you predicting here?

src/eos-autoupdater.c
cosimoc: I'm not sure I completely understand the issue here; since we're setting the mtime of the stamp file `IntervalDays` in the past, if the interval was originally one week and then the user changed it to two weeks, would we not start an update one week after the change? Would you expect that the update happens two weeks after the change at that point instead?

pwithnall: I think the issue is that Krzesimir expects the next update to be started immediately (i.e. sometime shortly after the first boot), since no update has yet been done, despite changing the `IntervalDays` in the configuration file.

krnowak: Yes. To be exact - that the update should take place after the random days delay (which means immediately if the delay is zero) ignoring the `IntervalDays` for the first update.

src/eos-autoupdater.c
krnowak: What I meant here is that I understood that the `last_successful_update_secs` parameter tells us what the mtime of the timestamp should be, possibly altered a bit when the random delay is greater than zero. So we have 4 scenarios:
- `last_successful_update_secs` is now and random delay is 0 - timestamp of the file is now, so this is fine
- `last_successful_update_secs` is now and random delay is > 0 - timestamp is some random days in future
- `last_successful_update_secs` is 5 days ago and random delay is 0 - timestamp is now, so that's probably wrong
- `last_successful_update_secs` is 5 days ago and random delay is > 0 - timestamp is 5 days ago + random days, so that's fine.
(I'm not considering `last_successful_update_secs` being in future, but that's basically the same case as it being in the past.)

I'm actually predicting no problem at all here, since the third scenario does not happen at all so far - in this situation we perform the update immediately. It just looked wrong and likely not futureproof if we ever change the update logic again.

pwithnall: Ah, I see what you mean now. Thanks for clarifying. I’ll fix that.

src/eos-autoupdater.c
pwithnall: As I said above, I don’t think this is a problem we need to worry about.

PR #92, opened 2017-04-19
eos-updater: Handle G_IO_ERROR_FAILED when running in FBE test mode

pwithnall:
When test mode is enabled in the first boot environment, overlay file
systems are mounted on top of everything (including /ostree), which
breaks OSTree. (This is expected.)

This causes libostree to return an error from ostree_sysroot_load(),
which we were not correctly handling, and hence were passing through the
wrong error domain over eos-updater’s D-Bus interface. This caused an
error message to appear in the UI when it wasn’t supposed to.

https://phabricator.endlessm.com/T16574

### DIFF
--- a/src/eos-updater-poll.c
+++ b/src/eos-updater-poll.c
@@ -268,13 +268,16 @@ metadata_fetch (GTask *task,
 
   /* Check we’re not on a dev-converted system. */
   deployment = eos_updater_get_booted_deployment (&error);
-  if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
+  if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND) ||
+      g_error_matches (error, G_IO_ERROR, G_IO_ERROR_FAILED))
     {
       g_task_return_new_error (task, EOS_UPDATER_ERROR,
                                EOS_UPDATER_ERROR_NOT_OSTREE_SYSTEM,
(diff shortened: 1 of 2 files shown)


wjt:
Is there an argument that we should adjust the test mode overlays to match the overlays used for live boots, which do not cause this problem?

cosimoc:
This looks reasonable to me for now, so I merged it -- another option would be to try and explicitly detect test mode, since this error could also possibly have other causes.

pwithnall:
> another option would be to try and explicitly detect test mode, since this error could also possibly have other causes.

IIRC from looking at the libostree code, I don’t think there’s anything else obvious which could cause `G_IO_ERROR_FAILED` there actually. In any case, any I/O failure of `ostree_sysroot_load()` is a pretty good indication that the system is not (currently) an OSTree deployment, so I think the error we return is always appropriate.

> Is there an argument that we should adjust the test mode overlays to match the overlays used for live boots, which do not cause this problem? I guess live boot mode is explicitly detected. What is the failure that libostree hits?

Potentially. libostree hits the failure that `ostree=` is specified on the kernel command line, but the file system layout doesn’t look like a deployment, because `eos-test-mode` has overmounted overlay file systems everywhere. (At least, I assume that’s the cause of the problem — I have not investigated which mount breaks things.) This results in the error `Unexpected state: ostree= kernel argument found, but / is not a deployment root`.

I’ll look at the difference between test and live mode before closing the task.

PR #94, opened 2017-04-26
util: Add functions for parsing the entire string as a number

krnowak:
That thing got copy-pasted all over the place and I was about to do it
once again. Better just split it to function and add test to verify I
did not break anything.

PR #95, opened 2017-04-28
new DNS-SD records

krnowak:
This is a work in progress, not tested or anything, but can serve as a base for API review. It is based on #94, so please ignore the string to number conversion functions. I still have to review this code against the new IETF drafts.

(Edit: Pressed the submit too quickly)

Added three functions:

- generate an ostree avahi service file
  - it follows a bit an API convention in ostree, so it takes a GVariant being a vardict for options
- delete an ostree avahi service file
  - it deletes it for a specific repository index
- delete all ostree avahi service files but the ones in the whitelist
  - this is to be used in the situation when we change the config in a way that we remove some repositories, so we can also remove all the obsolete files without having to remember what files there were.
  - makes me wonder if this is really necessary - probably we could remove the whitelist, so all the ostree avahi service files would be removed and then we could recreate them based on a new configuration

https://phabricator.endlessm.com/T16589

### DIFF
--- a/eos-updater-avahi/docs/eos-updater-avahi.8
+++ b/eos-updater-avahi/docs/eos-updater-avahi.8
@@ -117,6 +117,13 @@ Describes the DNS\-SD record for Avahi to advertise to the local network. No
 advertisement will be made if this file is not present. This is the file which
 \fBeos\-updater\-avahi\fP adds, updates and removes when run.
 .\"
+.IP \fI/etc/avahi/services/eos\-ostree\-updater\-[0–65535].service\fP 4
+.IX Item "/etc/avahi/services/eos-ostree-updater\-[0–65535].service"
+Similarly to the \fI/etc/avahi/services/eos\-updater.service\fP file,
+it describes the DNS\-SD record for Avahi, but a different format is
+used. Many files can exist, one for each advertised repository. These
+files are managed by \fBeos\-updater\-avahi\fP.
+.\"
 .IP \fI/etc/eos\-updater/eos\-update\-server.conf\fP 4
 .IX Item "/etc/eos\-updater/eos\-update\-server.conf"
 .IX Item "/usr/local/share/eos\-updater/eos\-update\-server.conf"
--- a/eos-updater-avahi/eos-updater-avahi.c
+++ b/eos-updater-avahi/eos-updater-avahi.c
@@ -18,15 +18,188 @@
  *
  * Authors:
  *  - Philip Withnall <withnall@endlessm.com>
+ *  - Krzesimir Nowak <krzesimir@kinvolk.io>
  */
 
 #include <glib.h>
 #include <libeos-update-server/config.h>
 #include <libeos-updater-util/avahi-service-file.h>
+#include <libeos-updater-util/extensions.h>
 #include <libeos-updater-util/ostree.h>
 #include <locale.h>
 #include <ostree.h>
 #include <stdlib.h>
+
+static gboolean
+get_refs (OstreeRepo     *repo,
+          gchar        ***out_refs,
+          GCancellable   *cancellable,
+          GError        **error)
+{
+  g_autoptr(GHashTable) refs = NULL;
+  GHashTableIter iter;
+  gchar *key;
+  gchar *value;
+  g_autoptr(GPtrArray) refs_array = NULL;
+
+  if (!ostree_repo_list_refs (repo,
+                              NULL, /* no prefix filtering */
+                              &refs,
+                              cancellable,
+                              error))
+    return FALSE;
+
+  if (g_hash_table_size (refs) == 0)
+    {
+      g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                           "no refs to advertise");
+      return FALSE;
+    }
+  refs_array = g_ptr_array_new_full (g_hash_table_size (refs) + 1, g_free);
+  g_hash_table_iter_init (&iter, refs);
+  while (g_hash_table_iter_next (&iter, (gpointer *)&key, (gpointer *)&value))
+    {
+      g_autofree gchar *refspec = g_steal_pointer (&key);
+      g_autofree gchar *ref = NULL;
+
+      g_hash_table_iter_steal (&iter);
+      g_free (g_steal_pointer (&value));
+      if (!ostree_parse_refspec (refspec, NULL, &ref, error))
+        return FALSE;
+      g_ptr_array_add (refs_array, g_steal_pointer (&ref));
+    }
+  g_ptr_array_add (refs_array, NULL);
+
+  g_assert (out_refs);
+  *out_refs = (gchar **)g_ptr_array_free (g_steal_pointer (&refs_array), FALSE);
+  return TRUE;
+}
+
+static gboolean
+get_raw_summary_timestamp_from_metadata (GBytes    *summary,
+                                         gboolean  *out_found,
+                                         guint64   *out_timestamp,
+                                         GError   **error)
+{
+  g_autoptr(GVariant) summary_variant = g_variant_ref_sink (g_variant_new_from_bytes (OSTREE_SUMMARY_GVARIANT_FORMAT, summary, FALSE));
+  g_autoptr(GVariantIter) iter = NULL;
+  g_autofree gchar *key = NULL;
+  g_autoptr(GVariant) value = NULL;
+
+  g_variant_get (summary_variant, OSTREE_SUMMARY_GVARIANT_STRING, NULL, &iter);
+  while (g_variant_iter_next (iter, "{sv}", &key, &value))
+    {
+      if (g_strcmp0 (key, "ostree.summary.last-modified") == 0)
+        break;
+
+      g_free (g_steal_pointer (&key));
+      g_variant_unref (g_steal_pointer (&value));
+    }
+
+  g_assert (out_found != NULL);
+  g_assert (out_timestamp != NULL);
+  if (value == NULL)
+    {
+      *out_found = FALSE;
+      *out_timestamp = 0;
+      return TRUE;
+    }
+  if (!g_variant_is_of_type (value, G_VARIANT_TYPE_UINT64))
+    {
+      g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                           "\"ostree.summary.last-modified\" "
+                           "metadata has an invalid type");
+      return FALSE;
+    }
+  *out_found = TRUE;
+  *out_timestamp = GUINT64_FROM_BE (g_variant_get_uint64 (value));
+  return TRUE;
+}
+
+static gboolean
+bad_timestamp (guint64   modification_time_secs,
+               GError  **error)
+{
+  g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+               "Invalid timestamp %" G_GUINT64_FORMAT,
+               modification_time_secs);
+  return FALSE;
+}
+
+static gboolean
+get_summary_timestamp_from_guint64 (guint64     modification_time_secs,
+                                    GDateTime **out_summary_timestamp,
+                                    GError    **error)
+{
+  g_autoptr(GDateTime) summary_timestamp = NULL;
+
+  if (modification_time_secs > G_MAXINT64)
+    return bad_timestamp (modification_time_secs, error);
+
+  summary_timestamp = g_date_time_new_from_unix_utc ((gint64)modification_time_secs);
+  if (summary_timestamp == NULL)
+    return bad_timestamp (modification_time_secs, error);
+
+  g_assert (out_summary_timestamp);
+  *out_summary_timestamp = g_steal_pointer (&summary_timestamp);
+  return TRUE;
+}
+
+static gboolean
+get_summary_timestamp (OstreeRepo    *repo,
+                       GDateTime    **out_summary_timestamp,
+                       GCancellable  *cancellable,
+                       GError       **error)
+{
+  // FIXME: Find the summary without using eos specific bits to ease
+  // the moving of this code to ostree.
+  g_autoptr(EosExtensions) extensions = eos_extensions_new_from_repo (repo,
+                                                                      cancellable,
+                                                                      error);
+  gboolean found;
+  guint64 raw_timestamp;
+
+  if (extensions == NULL)
+    return FALSE;
+
+  if (!get_raw_summary_timestamp_from_metadata (extensions->summary, &found, &raw_timestamp, error))
+    return FALSE;
+
+  if (!found)
+    raw_timestamp = extensions->summary_modification_time_secs;
+
+  return get_summary_timestamp_from_guint64 (raw_timestamp, out_summary_timestamp, error);
+}
+
+static gboolean
+get_refs_and_summary_timestamp (OstreeSysroot   *sysroot,
+                                gchar         ***out_refs,
+                                GDateTime      **out_summary_timestamp,
+                                GCancellable    *cancellable,
+                                GError         **error)
+{
+  g_autoptr(OstreeRepo) repo = NULL;
+  g_auto(GStrv) refs = NULL;
+  g_autoptr(GDateTime) summary_timestamp = NULL;
+
+  if (!ostree_sysroot_get_repo (sysroot,
+                                &repo,
+                                cancellable,
+                                error))
+    return FALSE;
+
+  if (!get_refs (repo, &refs, cancellable, error))
+    return FALSE;
+
+  if (!get_summary_timestamp (repo, &summary_timestamp, cancellable, error))
+    return FALSE;
+
+  g_assert (out_refs);
+  g_assert (out_summary_timestamp);
+  *out_refs = g_steal_pointer (&refs);
+  *out_summary_timestamp = g_steal_pointer (&summary_timestamp);
+  return TRUE;
+}
 
 static gboolean
 update_service_file (gboolean       advertise_updates,
@@ -106,6 +279,16 @@ update_service_file (gboolean       advertise_updates,
   /* Apply the policy. */
   if (!delete)
     {
+      g_autoptr(GDateTime) summary_timestamp = NULL;
+      g_auto(GStrv) refs = NULL;
+
+      if (!get_refs_and_summary_timestamp (sysroot,
+                                           &refs,
+                                           &summary_timestamp,
+                                           cancellable,
+                                           error))
+        return FALSE;
+
       if (!eos_avahi_service_file_generate (avahi_service_directory,
                                             commit_ostree_path,
                                             commit_date_time,
@@ -116,13 +299,33 @@ update_service_file (gboolean       advertise_updates,
                                          NULL);
           return FALSE;
         }
+      if (!eos_ostree_avahi_service_file_generate (avahi_service_directory,
+                                                   (const gchar *const *) refs,
+                                                   summary_timestamp,
+                                                   NULL, /* no options, use defaults */
+                                                   cancellable,
+                                                   error))
+        {
+          eos_ostree_avahi_service_file_delete (avahi_service_directory,
+                                                0,
+                                                cancellable,
+                                                NULL);
+          eos_avahi_service_file_delete (avahi_service_directory, cancellable,
+                                         NULL);
+          return FALSE;
+        }
 
       return TRUE;
     }
   else
     {
-      return eos_avahi_service_file_delete (avahi_service_directory,
-                                            cancellable, error);
+      if (!eos_avahi_service_file_delete (avahi_service_directory,
+                                          cancellable, error))
+        return FALSE;
+      return eos_ostree_avahi_service_file_delete (avahi_service_directory,
+                                                   0,
+                                                   cancellable,
+                                                   error);
     }
 }
 
--- a/eos-updater-avahi/eos-updater-avahi.c
+++ b/eos-updater-avahi/eos-updater-avahi.c
@@ -31,22 +31,22 @@
 #include <stdlib.h>
 
 static gboolean
-get_refs (OstreeRepo     *repo,
-          gchar        ***out_refs,
-          GCancellable   *cancellable,
-          GError        **error)
+get_refs (OstreeRepo            *repo,
+          OstreeCollectionRef ***out_refs,
+          GCancellable          *cancellable,
+          GError               **error)
 {
   g_autoptr(GHashTable) refs = NULL;
   GHashTableIter iter;
-  gchar *key;
+  OstreeCollectionRef *key;
   gchar *value;
   g_autoptr(GPtrArray) refs_array = NULL;
 
-  if (!ostree_repo_list_refs (repo,
-                              NULL, /* no prefix filtering */
-                              &refs,
-                              cancellable,
-                              error))
+  if (!ostree_repo_list_collection_refs (repo,
+                                         /* match_collection_id: */ NULL,
+                                         &refs,
+                                         cancellable,
+                                         error))
     return FALSE;
 
   if (g_hash_table_size (refs) == 0)
@@ -55,23 +55,18 @@ get_refs (OstreeRepo     *repo,
                            "no refs to advertise");
       return FALSE;
     }
-  refs_array = g_ptr_array_new_full (g_hash_table_size (refs) + 1, g_free);
+  refs_array = g_ptr_array_new_full (g_hash_table_size (refs) + 1, ostree_collection_ref_free);
   g_hash_table_iter_init (&iter, refs);
   while (g_hash_table_iter_next (&iter, (gpointer *)&key, (gpointer *)&value))
     {
-      g_autofree gchar *refspec = g_steal_pointer (&key);
-      g_autofree gchar *ref = NULL;
-
       g_hash_table_iter_steal (&iter);
       g_free (g_steal_pointer (&value));
-      if (!ostree_parse_refspec (refspec, NULL, &ref, error))
-        return FALSE;
-      g_ptr_array_add (refs_array, g_steal_pointer (&ref));
+      g_ptr_array_add (refs_array, key);
     }
   g_ptr_array_add (refs_array, NULL);
 
   g_assert (out_refs);
-  *out_refs = (gchar **)g_ptr_array_free (g_steal_pointer (&refs_array), FALSE);
+  *out_refs = (OstreeCollectionRef **)g_ptr_array_free (g_steal_pointer (&refs_array), FALSE);
   return TRUE;
 }
 
@@ -172,14 +167,14 @@ get_summary_timestamp (OstreeRepo    *repo,
 }
 
 static gboolean
-get_refs_and_summary_timestamp (OstreeSysroot   *sysroot,
-                                gchar         ***out_refs,
-                                GDateTime      **out_summary_timestamp,
-                                GCancellable    *cancellable,
-                                GError         **error)
+get_refs_and_summary_timestamp (OstreeSysroot         *sysroot,
+                                OstreeCollectionRef ***out_refs,
+                                GDateTime            **out_summary_timestamp,
+                                GCancellable          *cancellable,
+                                GError               **error)
 {
   g_autoptr(OstreeRepo) repo = NULL;
-  g_auto(GStrv) refs = NULL;
+  g_auto(OstreeCollectionRefv) refs = NULL;
   g_autoptr(GDateTime) summary_timestamp = NULL;
 
   if (!ostree_sysroot_get_repo (sysroot,
@@ -280,7 +275,7 @@ update_service_file (gboolean       advertise_updates,
   if (!delete)
     {
       g_autoptr(GDateTime) summary_timestamp = NULL;
-      g_auto(GStrv) refs = NULL;
+      g_auto(OstreeCollectionRefv) refs = NULL;
 
       if (!get_refs_and_summary_timestamp (sysroot,
                                            &refs,
@@ -300,7 +295,7 @@ update_service_file (gboolean       advertise_updates,
           return FALSE;
         }
       if (!eos_ostree_avahi_service_file_generate (avahi_service_directory,
-                                                   (const gchar *const *) refs,
+                                                   refs,
                                                    summary_timestamp,
                                                    NULL, /* no options, use defaults */
                                                    cancellable,
--- a/libeos-update-server/repo.c
+++ b/libeos-update-server/repo.c
@@ -625,9 +625,19 @@ path_is_handled_as_is (const gchar *requested_path)
     }
 
   if (g_str_has_prefix (requested_path, "/deltas/") ||
-      g_str_has_prefix (requested_path, "/extensions/") ||
-      g_str_equal (requested_path, "/summary") ||
-      g_str_equal (requested_path, "/summary.sig"))
+      g_str_has_prefix (requested_path, "/extensions/"))
+    return TRUE;
+
+  return FALSE;
+}
+
+static gboolean
+path_is_summary (const gchar *path)
+{
+  if (g_str_equal (path, "/summary") ||
+      g_str_equal (path, "/summary.sig") ||
+      g_str_equal (path, "/extensions/eos/eos-summary") ||
+      g_str_equal (path, "/extensions/eos/eos-summary.sig"))
     return TRUE;
 
   return FALSE;
@@ -747,6 +757,40 @@ handle_config (EusRepo     *self,
 }
 
 static void
+handle_summary (EusRepo     *self,
+                SoupMessage *msg,
+                const gchar *requested_path)
+{
+  g_autofree gchar *eos_raw_path = NULL;
+
+  if (g_str_has_prefix (requested_path, "/summary"))
+    {
+      g_autofree gchar *raw_path = g_build_filename (self->cached_repo_root, requested_path, NULL);
+      g_autofree gchar *eos_summary_path = NULL;
+      gboolean served = FALSE;
+
+      if (!serve_file_if_exists (msg,
+                                 self->cached_repo_root,
+                                 raw_path,
+                                 self->cancellable,
+                                 &served))
+        return;
+      if (served)
+        return;
+
+      eos_summary_path = g_strconcat ("/extensions/eos/eos-", requested_path + 1, NULL);
+      eos_raw_path = g_build_filename (self->cached_repo_root, eos_summary_path, NULL);
+      g_debug ("no %s, falling back to %s", requested_path, eos_summary_path);
+    }
+  else
+    {
+      eos_raw_path = g_build_filename (self->cached_repo_root, requested_path, NULL);
+    }
+
+  serve_file (msg, self->cached_repo_root, eos_raw_path, self->cancellable);
+}
+
+static void
 handle_refs_heads (EusRepo     *self,
                    SoupMessage *msg,
                    const gchar *requested_path)
@@ -820,8 +864,10 @@ handle_path (EusRepo     *self,
     handle_objects_filez (self, msg, path);
   else if (path_is_handled_as_is (path))
     handle_as_is (self, msg, path);
-  else if (g_strcmp0 (path, "/config") == 0)
+  else if (g_str_equal (path, "/config"))
     handle_config (self, msg);
+  else if (path_is_summary (path))
+    handle_summary (self, msg, path);
   else if (g_str_has_prefix (path, "/refs/heads/"))
     handle_refs_heads (self, msg, path);
   else
--- a/libeos-updater-util/avahi-service-file.c
+++ b/libeos-updater-util/avahi-service-file.c
@@ -76,6 +76,7 @@ check_record_size (const gchar  *key,
 
 static gboolean
 handle_text_record (GString      *records_str,
+                    gsize        *total_size,
                     const gchar  *key,
                     GVariant     *text_value,
                     GError      **error)
@@ -91,6 +92,11 @@ handle_text_record (GString      *records_str,
   if (!check_record_size (key, record_size, error))
     return FALSE;
 
+  /* txt records are pascal strings, so one byte for length and then
+   * payload.
+   */
+  g_assert (total_size != NULL);
+  *total_size += 1 + record_size;
   escaped_key = g_markup_escape_text (key, key_length);
   escaped_value = g_markup_escape_text (value_string, value_string_length);
   g_string_append_printf (records_str,
@@ -102,6 +108,7 @@ handle_text_record (GString      *records_str,
 
 static gboolean
 handle_binary_record (GString      *records_str,
+                      gsize        *total_size,
                       const gchar  *key,
                       GVariant     *binary_value,
                       GError      **error)
@@ -119,6 +126,11 @@ handle_binary_record (GString      *records_str,
   if (!check_record_size (key, record_size, error))
     return FALSE;
 
+  /* txt records are pascal strings, so one byte for length and then
+   * payload.
+   */
+  g_assert (total_size != NULL);
+  *total_size += 1 + record_size;
   escaped_key = g_markup_escape_text (key, key_length);
   encoded_value = g_base64_encode (value_data, value_data_length);
   escaped_value = g_markup_escape_text (encoded_value, -1);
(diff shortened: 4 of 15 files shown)


libeos-updater-util/avahi-service-file.c
```diff
@@ -110,35 +179,35 @@ generate_avahi_service_template_to_file (GFile *path,
...
+  g_variant_builder_add (&records_builder, "(sv)",
+                         "eos_txt_version",
+                         g_variant_new_string ("1"));
+  g_variant_builder_add (&records_builder, "(sv)",
+                         eos_avahi_v1_ostree_path,
+                         g_variant_new_string (ostree_path));
+  g_variant_builder_add (&records_builder, "(sv)",
+                         eos_avahi_v1_head_commit_timestamp,
+                         g_variant_new_string (timestamp_str));
+  txt_records = g_variant_ref_sink (g_variant_builder_end (&records_builder));
```
pwithnall: Would it make more sense to have `generate_avahi_service_template_to_file()` do the `ref_sink` call on `txt_records`?

libeos-updater-util/avahi-service-file.c
```diff
@@ -257,3 +326,426 @@ eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+  return index;
+}
+
+static guint16
+get_avahi_service_port (GVariantDict *options_dict)
+{
+  // TODO: Should we store the port number in the configuration
+  // instead having it as the compile-time constant? I can't remember
+  // why I opted for this solution (was it lack of the configuration
+  // file?)
```
pwithnall: Yeah, moving it to a config file would be good. That can be done as a separate commit/PR though, at a fairly low priority.

libeos-updater-util/avahi-service-file.c
```diff
@@ -257,3 +326,426 @@ eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+
+  if (!ostree_repo_list_refs (repo,
+                              NULL, /* no prefix filtering */
+                              &refs,
+                              cancellable,
+                              error))
+    return FALSE;
+
+  g_hash_table_iter_init (&iter, refs);
+  while (g_hash_table_iter_next (&iter, &key, NULL))
```
pwithnall: Would be nice to give `key` its proper type (`const gchar *`) rather than it remaining a mysterious `gpointer`. That means adding a cast to the `g_hash_table_iter_next()` call though.

libeos-updater-util/avahi-service-file.c
```diff
@@ -257,3 +326,426 @@ eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+  g_assert (g_variant_type_equal (g_variant_get_type (summary_timestamp_variant),
+                                  EOS_OSTREE_AVAHI_V1_SUMMARY_TIMESTAMP_VARIANT_TYPE));
+  repository_index_variant = variant_sink_new ("q", repository_index);
+  g_assert (g_variant_type_equal (g_variant_get_type (repository_index_variant),
+                                  EOS_OSTREE_AVAHI_V1_REPOSITORY_INDEX_VARIANT_TYPE));
+
+  service_file = get_ostree_service_file (avahi_service_directory,
+                                          repository_index);
+
+  return generate_ostree_avahi_v1_service_file_from_variants (service_file,
```
pwithnall: You could drop `variant_sink_new()` and have `generate_ostree_avahi_v1_service_file_from_variants()` do the `ref_sink` calls.

libeos-updater-util/avahi-service-file.c
```diff
@@ -257,3 +326,426 @@ eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+
+#define variant_sink_new(format_string, ...) (g_variant_ref_sink (g_variant_new (format_string, __VA_ARGS__)))
+
+static GVariant *
+get_bloom_filter_variant (OstreeRepo *repo,
+                          GVariantDict *options_dict,
+                          GCancellable *cancellable,
+                          GError **error)
+{
+  guint32 bloom_size = 255 - strlen(EOS_OSTREE_AVAHI_V1_REFS_BLOOM_FILTER_FIELD) - 2;
```
pwithnall: A comment explaining the maths here would be useful for future reference.

libeos-updater-util/avahi-service-file.c
```diff
@@ -257,3 +326,426 @@ eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+  guint8 bloom_k = 1;
+  guint8 bloom_hash_id = EOS_OSTREE_AVAHI_STR_BLOOM_HASH_ID;
```
pwithnall: It would probably also be useful to add a comment identifying which parameters need to be kept stable otherwise binary compatibility of the bloom filter between server and client will be broken. (`bloom_size`, `bloom_k` and `bloom_hash_id`.)

libeos-updater-util/avahi-service-file.c
```diff
@@ -257,3 +326,426 @@ eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+  OstreeBloomHashFunc bloom_hash_func;
+  g_autoptr(OstreeBloom) filter = NULL;
+  g_autoptr(GBytes) sealed_filter = NULL;
+  g_autoptr(GVariant) bloom_bits_variant = NULL;
+
+  g_variant_dict_lookup (options_dict, EOS_OSTREE_AVAHI_OPTION_BLOOM_SIZE_U,
+                         "u", &bloom_size);
+  g_variant_dict_lookup (options_dict, EOS_OSTREE_AVAHI_OPTION_BLOOM_K_Y,
+                         "y", &bloom_k);
```
pwithnall: You need to validate that `bloom_k > 0`.

libeos-updater-util/avahi-service-file.c
```diff
@@ -257,3 +326,426 @@ eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+  guint32 bloom_size = 255 - strlen(EOS_OSTREE_AVAHI_V1_REFS_BLOOM_FILTER_FIELD) - 2;
+  guint8 bloom_k = 1;
+  guint8 bloom_hash_id = EOS_OSTREE_AVAHI_STR_BLOOM_HASH_ID;
+  OstreeBloomHashFunc bloom_hash_func;
+  g_autoptr(OstreeBloom) filter = NULL;
+  g_autoptr(GBytes) sealed_filter = NULL;
+  g_autoptr(GVariant) bloom_bits_variant = NULL;
+
+  g_variant_dict_lookup (options_dict, EOS_OSTREE_AVAHI_OPTION_BLOOM_SIZE_U,
+                         "u", &bloom_size);
```
pwithnall: You need to validate that `bloom_size > 0`.

libeos-updater-util/avahi-service-file.c
```diff
@@ -257,3 +326,426 @@ eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+    default:
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                   "unknown bloom hash function id %u", bloom_hash_id);
+      return NULL;
+    }
+
+  filter = ostree_bloom_new (bloom_size, bloom_k, bloom_hash_func);
+  if (!fill_filter_with_refs (filter, repo, cancellable, error))
+    return FALSE;
+  sealed_filter = ostree_bloom_seal (filter);
```
pwithnall: OOI, what do you think of the `OstreeBloom` API? The API’s not stable yet, so if you can think of any improvements, please say, and I can get them into the libostree branch before it’s merged.

libeos-updater-util/avahi-service-file.c
```diff
@@ -257,3 +326,426 @@ eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+                                  EOS_OSTREE_AVAHI_VERSION_VARIANT_TYPE));
+  refs_bloom_filter_variant = get_bloom_filter_variant (repo,
+                                                        options_dict,
+                                                        cancellable,
+                                                        error);
+  if (refs_bloom_filter_variant == NULL)
+    {
+      return FALSE;
+    }
+  g_assert (g_variant_type_equal (g_variant_get_type (refs_bloom_filter_variant),
```
pwithnall: Can simplify this a bit by using `g_variant_is_of_type()`.

libeos-updater-util/avahi-service-file.c
```diff
@@ -257,3 +326,426 @@ eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+                                                    repo,
+                                                    &options_dict,
+                                                    cancellable,
+                                                    error);
+
+    default:
+      break;
+    }
+
+  g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
```
pwithnall: Move this into the `default:` case for symmetry with the other case handling.

libeos-updater-util/avahi-service-file.c
```diff
@@ -257,3 +326,426 @@ eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+      if (eos_string_to_unsigned (repository_index_str,
+                                  0,
+                                  G_MAXUINT16,
+                                  &repository_index,
+                                  NULL))
+        {
+          gsize idx;
+
+          for (idx = 0; idx < indices_length; ++idx)
+            if (idx == repository_index)
```
pwithnall: This `for; if` construct could just be replaced by `return (repository_index >= indices_length)`, right?

libeos-updater-util/avahi-service-file.h
```diff
@@ -44,4 +46,42 @@ gboolean eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
                                         GCancellable  *cancellable,
                                         GError       **error);
 
+/* new DNS-SD records format for ostree */
```
pwithnall: Eventually it would be good to have some documentation for the format somewhere (in a comment, probably), to save people having to read through the code and work out all the invariants.

libeos-updater-util/avahi-service-file.h
```diff
@@ -44,4 +46,42 @@ gboolean eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+#define EOS_OSTREE_AVAHI_V1_REPOSITORY_INDEX_VARIANT_TYPE G_VARIANT_TYPE_UINT16
+
+#define EOS_OSTREE_AVAHI_OPTION_FORCE_VERSION_Y "force-version"
+#define EOS_OSTREE_AVAHI_OPTION_BLOOM_HASH_ID_Y "bloom-hash-id"
+#define EOS_OSTREE_AVAHI_OPTION_BLOOM_K_Y "bloom-k"
+#define EOS_OSTREE_AVAHI_OPTION_BLOOM_SIZE_U "bloom-size"
+#define EOS_OSTREE_AVAHI_OPTION_REPO_INDEX_Q "repository-index"
+#define EOS_OSTREE_AVAHI_OPTION_PORT_Q "port"
+
+#define EOS_OSTREE_AVAHI_STR_BLOOM_HASH_ID 1
```
pwithnall: Do all these `#define`s have to be in the header file? They could be private in the C file, right? No reason to de-encapsulate them.

libeos-updater-util/avahi-service-file.c
pwithnall: Unless the `idx` is actually supposed to be used as `indices_to_spare[idx]`?

libeos-updater-util/avahi-service-file.h
```diff
@@ -44,4 +46,42 @@ gboolean eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+gboolean eos_ostree_avahi_service_file_generate (const gchar   *avahi_service_directory,
+                                                 OstreeRepo    *repo,
+                                                 GVariant      *options,
+                                                 GCancellable  *cancellable,
+                                                 GError       **error);
+gboolean eos_ostree_avahi_service_file_delete (const gchar   *avahi_service_directory,
+                                               guint16        repository_index,
+                                               GCancellable  *cancellable,
+                                               GError       **error);
+gboolean eos_ostree_avahi_service_file_cleanup_directory (const gchar   *avahi_service_directory,
```
pwithnall: I think this `cleanup_directory()` API is too complex. There should be no harm in having a `delete_all()` API and then having the caller re-generate the repositories it wants to keep. If that causes unnecessary churn in Avahi’s DNS-SD broadcasts, then I think that’s a bug in Avahi which needs to be fixed (it should probably wait a second after service files are modified before propagating those changes to the network, in case more service file modifications are coming; I don’t know if it does that already).

libeos-updater-util/extensions.h
```diff
@@ -75,6 +75,7 @@ struct _EosExtensions
 
   GBytes *summary;
   GBytes *summary_sig;
+  guint64 summary_modification_time_unix;
```
pwithnall: I would call this `summary_modification_time_secs` to make it clear what the units are. Maybe add a comment after: `/* since the Unix epoch */`.

libeos-updater-util/avahi-service-file.c
```diff
@@ -38,65 +41,131 @@ const gchar * const eos_avahi_v1_ostree_path = "eos_ostree_path";
...
-txt_records_to_string (const gchar **txt_records)
+txt_records_to_string (GVariant *txt_records)
 {
+  GVariantIter iter;
   g_autoptr(GString) str = NULL;
-  gsize idx;
+  GVariant *txt_value;
+  const gchar *txt_key;
+
+  g_assert (g_variant_type_equal (g_variant_get_type (txt_records),
```
krnowak: This is a static function, so I think that g_assert is fine. If it fires, then it means I have made a mistake somewhere, as txt records are constructed entirely in this file.

Thanks for the hint about `g_variant_is_of_type` - I can see it is not exactly the same as using `g_variant_type_equal`, but it shouldn't matter in our case.

libeos-updater-util/avahi-service-file.c
```diff
@@ -38,65 +41,131 @@ const gchar * const eos_avahi_v1_ostree_path = "eos_ostree_path";
...
+        {
+          gsize txt_value_data_length;
+          gconstpointer txt_value_data = g_variant_get_fixed_array (txt_value,
+                                                                    &txt_value_data_length,
+                                                                    1);
+          g_autofree gchar *txt_value_encoded = NULL;
+          g_autofree gchar *txt_value_escaped = NULL;
+
+          if (txt_value_data_length + strlen (txt_key) + 1 > 255)
+            g_warning ("The TXT record with key %s is longer than 255 bytes, it will be truncated!",
```
krnowak: Hmm, bloom size is one of the options we pass to the generator function, so adding an assertion here would mean that the users of the `eos_ostree_avahi_service_file_generate` function will need to do the math themselves (this 255 - 3 - strlen(key)). Unless we implement this `rb0`, `rb1`, `rbN` thing.

libeos-updater-util/avahi-service-file.c
```diff
@@ -38,65 +41,131 @@ const gchar * const eos_avahi_v1_ostree_path = "eos_ostree_path";
...
+                                  txt_value_escaped);
+        }
     }
 
+  // TODO: compute the total size of txt records and complain loudly
+  // if they are longer then certain sizes:
+  //
+  // - more than 256 bytes, because some crappy hardware did not work
+  //   properly with longer records (likely not applicable for us and
+  //   that would limit the size of the bloom filter severely)
```
krnowak: I don't think so. At least - not on this level, maybe in avahi. But I wrote this after reading [a paragraph in section 6.2 in rfc 6763](https://tools.ietf.org/html/rfc6763#section-6.2).

I believe the symptoms would be simply truncated DNS-SD records sent over the network. I think that we would need to simply state that we don't support crappy hardware (after checking if the hardware we want to run this code on is indeed not crappy ;) )

libeos-updater-util/avahi-service-file.c
krnowak: I'd say no, even if it means that we need to worry about sinking the gvariant ourselves and general verbosity of the code. My impression is that functions that actually store the gvariant should call `ref_sink`. In our case the functions simply read contents of the gvariant.

libeos-updater-util/avahi-service-file.c
krnowak: Sure, will keep this TODO there then.

libeos-updater-util/avahi-service-file.c
krnowak: Alright.

libeos-updater-util/avahi-service-file.c
krnowak: Heh, and I got the math wrong here, I think. I need to subtract one byte more (1 for bloom k, one for hash id, one for `=`).

libeos-updater-util/avahi-service-file.c
krnowak: I'm not sure I understand what you mean here. The default values here can change anytime - the server should handle the change just fine. Or did you mean the `(yyay)` signature?

libeos-updater-util/avahi-service-file.c
krnowak: Sure.

libeos-updater-util/avahi-service-file.c
krnowak: Sure.

libeos-updater-util/avahi-service-file.c
krnowak: The `ostree_bloom_add_element` function was a bit confusing in the beginning as it was just taking a gpointer and no length. That is - it was until I understood that specifying the hash function for the bloom filter also specifies the element type (in our case - nul-terminated string).

libeos-updater-util/avahi-service-file.c
krnowak: Yes.

libeos-updater-util/avahi-service-file.c
krnowak: As stated above - I'm not sure that the function that does not store the gvariant and does not modify it should ever sink or ref it.

libeos-updater-util/avahi-service-file.c
krnowak: The other case handling does it in a different way only because there was some more code after `switch`. I can move this inside the default - I hope that the compiler won't start complaining about missing `return`.

libeos-updater-util/avahi-service-file.c
krnowak: Oh, right, thanks for spotting. Yeah, the second comment is right. But all this will be likely gone anyway.

libeos-updater-util/avahi-service-file.h
krnowak: Will do.

libeos-updater-util/avahi-service-file.h
krnowak: Maybe not all the defines, but the `EOS_OSTREE_AVAHI_OPTION` ones could stay here, so the user can use them when specifying options for `eos_ostree_avahi_service_file_generate` instead of writing "repository-index" themselves.

libeos-updater-util/avahi-service-file.h
krnowak: Right, I also had doubts about this API. I will make it simpler.

Also, avahi currently uses inotify (or kqueue on BSDs I think) to watch the services directory. I don't see any timeout here, so I think avahi reloads the config immediately.

libeos-updater-util/extensions.h
krnowak: Ok.

libeos-updater-util/avahi-service-file.c
pwithnall: As I understand things, the main thinking behind when to use `ref_sink` is to allow `g_variant_new()` to be used inline where appropriate (which means for most functions which take a `GVariant` as a parameter, like here). For example, I think it would be reasonable to want to call the following (perhaps in a unit test):

`generate_avahi_service_template_to_file (…, g_variant_new_parsed ("@a{sv} { 'blah': … }"), …);`

libeos-updater-util/avahi-service-file.c
pwithnall: The alternative is the callers triggering a `g_warning()` if they don’t do the maths themselves, which is just as bad. Code should be able to work out if it’s going to trigger a warning. I generally fix this by adding a `check` method which just does the validation (and returns a `GError` on failure), which the caller can call before calling this function.

libeos-updater-util/avahi-service-file.c
pwithnall: Is this somewhere where a `G_STATIC_ASSERT` to check the total length of something (not necessarily this variable) would be useful to catch mistakes?

/me is a big fan of `G_STATIC_ASSERT`.

libeos-updater-util/avahi-service-file.c
pwithnall: Sorry, you’re right. I was reading through this too fast. Ignore my comment. :-)

libeos-updater-util/avahi-service-file.c
pwithnall: Yup; it’s just the same as `GHashTable` in this respect. I’ll try and make sure that’s covered in the documentation.

libeos-updater-util/avahi-service-file.c
pwithnall: The compiler won’t start complaining; it should be able to figure out that there’s a `return` on all code paths from the `switch`.

krnowak:
Updated, but I did not even try to compile it. Not everything was addressed yet (mostly documenting the txt messages format and the stuff eos extensions file).

libeos-updater-util/avahi-service-file.c
krnowak: I rewrote the code a bit, so now it returns a GError if the record is too long. Also added a `check_options` function that does some basic checks.

libeos-updater-util/avahi-service-file.c
krnowak: I added another option, so you can specify which size level you don't want to break.

libeos-updater-util/avahi-service-file.c
krnowak: Added the sinking to `generate_avahi_service_template_to_file`.

libeos-updater-util/avahi-service-file.c
krnowak: I like static assert too, but nothing comes to my mind about how we could use it in this case.

libeos-updater-util/avahi-service-file.c
krnowak: Maybe consider also adding an API for getting the `k` parameter and a hash function from `OstreeBloom`. I could use it here.

libeos-updater-util/avahi-service-file.c
krnowak: Added sinking to `variant_to_binary_variant` which gets called by `generate_ostree_avahi_v1_service_file_from_variants` for all the variants it takes.

libeos-updater-util/avahi-service-file.c
krnowak: Indeed, it did not complain.

krnowak:
Now it builds, added a bit of the TXT records format docs. There are some TODOs left, some of which I will address tomorrow. Tests are still to be written.

libeos-updater-util/avahi-service-file.c
pwithnall: I’ve added `ostree_bloom_get_size()`, `ostree_bloom_get_k()` and `ostree_bloom_get_hash_func()` in the upstream branch.

libeos-updater-util/avahi-service-file.h
```diff
@@ -44,4 +46,160 @@ gboolean eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+ * EOS_OSTREE_AVAHI_SIZE_LEVEL_ABSOLUTELY_LAX:
+ *
+ * TODO
+ */
+#define EOS_OSTREE_AVAHI_SIZE_LEVEL_ABSOLUTELY_LAX ((guchar)6)
+
+gboolean eos_ostree_avahi_service_file_check_options (GVariant  *options,
+                                                      GError   **error);
+gboolean eos_ostree_avahi_service_file_generate (const gchar   *avahi_service_directory,
+                                                 OstreeRepo    *repo,
```
krnowak: What do you think about replacing the `OstreeRepo *repo` parameter with a string list describing refs and a GDateTime describing repo's summary timestamp? I remember I tried to avoid using ostree API in previous version of DNS-SD records generation API. This would be also tons easier to test. Not sure how to pass them - should them be a part `GVariant *options` (say, "advertised-refs" being an `as` variant and "summary-timestamp" being a `t` variant or rather as explicit parameters to the function. I'm leaning to the latter, as refs and timestamp are not optional and have no sane default values we could use when the user does not specify "advertised-refs" or "summary-timestamp" options.

krnowak: OTOH, passing OstreeRepo allows us to extract some more or just different data from the repo for the future versions of TXT records. Passing refs and summary timestamp explicitly can be risky, because we may possibly not use them in some future version of TXT records.

So maybe just stuffing everything to the options variant is not an entirely bad idea in order to avoid both using ostree API and having useless parameters in future.

pwithnall: I’d go with explicit list-of-refs and timestamp arguments. An options `GVariant` is too flexible (type safety is lost; as you say, both of those parameters are required). Taking an `OstreeRepo` would be slightly nicer, but way too hard to test.

If more data is needed for future versions of the TXT records, we can always add a new API and deprecate this one.

I think when this code is moved into OSTree, we might change the API to take an `OstreeRepo` instead, since it should be easier to test from inside OSTree. But that’s something to worry about later.

krnowak: Ah, I went with the options-only thing. Will rework it later then.

krnowak:
Updated to fix two bugs in eos-updater-avahi (I forgot to steal the autoptr variable for out parameter and forgot to terminate the string array with NULL).

eos-updater-avahi/docs/eos-updater-avahi.8
```diff
@@ -117,6 +117,13 @@ Describes the DNS\-SD record for Avahi to advertise to the local network. No
 advertisement will be made if this file is not present. This is the file which
 \fBeos\-updater\-avahi\fP adds, updates and removes when run.
 .\"
+.IP \fI/etc/avahi/services/eos\-ostree\-updater\-[0\-65535].service\fP 4
+.IX Item "/etc/avahi/services/eos-ostree-updater-[0-65535].service"
+Similarly to the eos-updater.service file in the same directory, it
```
pwithnall: `eos-updater.service` needs to be formatted like other filenames in the man page, and the hyphen in it escaped.

eos-updater-avahi/docs/eos-updater-avahi.8
```diff
@@ -117,6 +117,13 @@ Describes the DNS\-SD record for Avahi to advertise to the local network. No
 advertisement will be made if this file is not present. This is the file which
 \fBeos\-updater\-avahi\fP adds, updates and removes when run.
 .\"
+.IP \fI/etc/avahi/services/eos\-ostree\-updater\-[0\-65535].service\fP 4
+.IX Item "/etc/avahi/services/eos-ostree-updater-[0-65535].service"
```
pwithnall: Hyphens need to be escaped.

eos-updater-avahi/docs/eos-updater-avahi.8
```diff
@@ -117,6 +117,13 @@ Describes the DNS\-SD record for Avahi to advertise to the local network. No
...
+Similarly to the eos-updater.service file in the same directory, it
+describes the DNS-SD record for Avahi, but a different format is
```
pwithnall: Hyphen in ‘DNS-SD’ needs to be escaped.

eos-updater-avahi/docs/eos-updater-avahi.8
```diff
@@ -117,6 +117,13 @@ Describes the DNS\-SD record for Avahi to advertise to the local network. No
 advertisement will be made if this file is not present. This is the file which
 \fBeos\-updater\-avahi\fP adds, updates and removes when run.
 .\"
+.IP \fI/etc/avahi/services/eos\-ostree\-updater\-[0\-65535].service\fP 4
```
pwithnall: Use an en-dash in the numeric range: 0–65535. Same in the line below.

eos-updater-avahi/eos-updater-avahi.c
```diff
@@ -23,12 +23,120 @@
...
+  return TRUE;
+}
+
+static gboolean
+get_summary_timestamp (OstreeRepo    *repo,
+                       GDateTime    **out_summary_timestamp,
+                       GCancellable  *cancellable,
+                       GError       **error)
+{
+  // TODO: Find the summary without using eos specific bits to ease
```
pwithnall: Make this a `FIXME`, since I’ve been trying to follow the convention in eos-updater that `TODO` is something to fix before a patch is merged, and `FIXME` is something longer-term to fix eventually.

eos-updater-avahi/eos-updater-avahi.c
```diff
@@ -23,12 +23,120 @@
...
+    }
+  summary_timestamp = g_date_time_new_from_unix_utc ((gint64)extensions->summary_modification_time_secs);
+  if (summary_timestamp == NULL)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                   "Invalid timestamp %" G_GUINT64_FORMAT,
+                   extensions->summary_modification_time_secs);
+      return FALSE;
+    }
+  // TODO: Try getting the timestamp from summary metadata first.
```
pwithnall: This needs to be fixed. [libostree now adds it to the summary file](https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-repo.c#L4565)

libeos-updater-util/avahi-service-file.h
```diff
@@ -44,4 +44,241 @@ gboolean eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
                                         GCancellable  *cancellable,
                                         GError       **error);
 
+/* new DNS-SD records format for ostree */
+/* TODO: gtkdocize it? */
```
pwithnall: Drop this `TODO`. As long as the content’s there, there’s no need to gtkdocize it yet.

libeos-updater-util/avahi-service-file.h
```diff
@@ -44,4 +44,241 @@ gboolean eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+ *   - key: "ri"
+ *   - type: "q" (uint16)
+ *   - contents: it gets appended to the host's URL as "/%u"; 0 is a special
+ *               case - it is not appended, the host's URL serves as a home
+ *               of the repository
+ *
+ * - summary timestamp,
+ *   - key: "st",
+ *   - type: "t" (uint64)
+ *   - contents: a unix utc timestamp of the summary, ideally telling when
```
pwithnall: Should mention that it’s in seconds (not in milliseconds, for example).

pwithnall: Need to specify its endianness, too.

libeos-updater-util/avahi-service-file.h
```diff
@@ -44,4 +44,241 @@ gboolean eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+ * - refs bloom filter, a Bloom filter that contains all the refs the host has
+ *   - key: "rb"
+ *   - type: "(yyay)" (tuple containing a byte, a byte and an array of bytes)
+ *   - contents: first byte is the "k" parameter, second byte is the
+ *               hash id, an array of bytes are the bloom filter bits
+ *
+ * - repository index, a number identifying an OSTree repository
+ *   - key: "ri"
+ *   - type: "q" (uint16)
+ *   - contents: it gets appended to the host's URL as "/%u"; 0 is a special
```
pwithnall: Need to specify its endianness.

libeos-updater-util/avahi-service-file.h
```diff
@@ -44,4 +44,241 @@ gboolean eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+ *
+ * Default value of this option (if not overridden) is
+ * %EOS_OSTREE_AVAHI_STR_BLOOM_HASH_ID.
+ */
+#define EOS_OSTREE_AVAHI_OPTION_BLOOM_HASH_ID_Y "bloom-hash-id"
+/**
+ * EOS_OSTREE_AVAHI_OPTION_BLOOM_K_Y:
+ *
+ * Specifies the k parameter for the bloom filter. It roughly
+ * translates to how many times an element will be hashed before using
```
pwithnall: Roughly? It means this *exactly*.

libeos-updater-util/avahi-service-file.h
```diff
@@ -44,4 +44,241 @@ gboolean eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+ */
+#define EOS_OSTREE_AVAHI_OPTION_PORT_Q "port"
+/**
+ * EOS_OSTREE_AVAHI_OPTION_TXT_RECORDS_SIZE_LEVEL_Y:
+ *
+ * Specifies the size limit generated TXT records can have.
+ * Possible values:
+ *
+ * - %EOS_OSTREE_AVAHI_SIZE_LEVEL_CUSTOM
+ * - %EOS_OSTREE_AVAHI_SIZE_LEVEL_SUPPORT_CRAPPY_HARDWARE
```
pwithnall: Maybe s/CRAPPY/FAULTY/

libeos-updater-util/avahi-service-file.h
```diff
@@ -44,4 +44,241 @@ gboolean eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+
+/**
+ * EOS_OSTREE_AVAHI_SIZE_LEVEL_CUSTOM:
+ *
+ * A possible value for the
+ * %EOS_OSTREE_AVAHI_OPTION_TXT_RECORDS_SIZE_LEVEL_Y option. It means
+ * that the size limit is specified in the
+ * %EOS_OSTREE_AVAHI_OPTION_TXT_RECORDS_CUSTOM_SIZE_T option.
+ */
+#define EOS_OSTREE_AVAHI_SIZE_LEVEL_CUSTOM ((guint8)0)
```
pwithnall: Why are the `EOS_OSTREE_AVAHI_SIZE_LEVEL_*` values not part of an enum?

libeos-updater-util/extensions.c
```diff
@@ -414,6 +415,22 @@ get_summary (EosExtensions *extensions,
...
+                                    &local_error);
+  if (summary_info == NULL)
+    {
+      if (!g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
+        {
+          g_propagate_error (error, g_steal_pointer (&local_error));
+          return FALSE;
+        }
+    }
+  extensions->summary_modification_time_secs = g_file_info_get_attribute_uint64 (summary_info,
```
pwithnall: It’s possible for `summary_info` to be `NULL` at this point, if the error was `G_IO_ERROR_NOT_FOUND` above.

libeos-updater-util/avahi-service-file.c
```diff
@@ -37,67 +39,355 @@ const gchar * const EOS_UPDATER_AVAHI_SERVICE_TYPE = "_eos_updater._tcp";
...
+handle_text_record (GString      *records_str,
+                    gsize        *total_size,
+                    const gchar  *key,
+                    GVariant     *text_value,
+                    GError      **error)
+{
+  gsize value_string_length;
+  const gchar *value_string = g_variant_get_string (text_value,
+                                                    &value_string_length);
+  gsize key_length = strlen(key);
```
pwithnall: Add a space before `(`.

libeos-updater-util/avahi-service-file.c
```diff
@@ -37,67 +39,355 @@ const gchar * const EOS_UPDATER_AVAHI_SERVICE_TYPE = "_eos_updater._tcp";
...
+                      gsize        *total_size,
+                      const gchar  *key,
+                      GVariant     *binary_value,
+                      GError      **error)
+{
+  gsize value_data_length;
+  gconstpointer value_data = g_variant_get_fixed_array (binary_value,
+                                                        &value_data_length,
+                                                        1);
+  gsize key_length = strlen(key);
```
pwithnall: Missing a space before the `(`.

libeos-updater-util/avahi-service-file.c
```diff
@@ -37,67 +39,355 @@ const gchar * const EOS_UPDATER_AVAHI_SERVICE_TYPE = "_eos_updater._tcp";
...
+            return FALSE;
+          }
+
+        break;
+      }
+
+    case EOS_OSTREE_AVAHI_SIZE_LEVEL_SUPPORT_CRAPPY_HARDWARE:
+      if (total_size > 256)
+        {
+          g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
```
pwithnall: No point in having separate error strings for all of these cases. You could put the size limits in a `const gsize[]` array indexed by the enum value, then factor out the error handling using a single error string like “TXT records of size %blah break the limit of %blah bytes.”.

libeos-updater-util/avahi-service-file.c
pwithnall: → Make it a `FIXME`.

libeos-updater-util/avahi-service-file.c
```diff
@@ -248,12 +554,697 @@ eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+                          guint32       *out_bloom_size,
+                          GError       **error)
+{
+  /* 255 bytes is a maximum size of the key=value TXT record pair.  We
+   * subtract the length of the key name, then 1 byte for =, 1 byte
+   * for bloom k and 1 byte for hash id. There is no space reserved
+   * for the array of bytes being the bloom filter bits, because it is
+   * the last member of the variant tuple and it is treated specially.
+   */
+  guint32 max_bloom_size = 255 - strlen(EOS_OSTREE_AVAHI_V1_REFS_BLOOM_FILTER_FIELD) - 3;
```
pwithnall: Missing a space before `(`.

eos-updater-avahi/docs/eos-updater-avahi.8
krnowak: Done.

eos-updater-avahi/docs/eos-updater-avahi.8
krnowak: Done.

eos-updater-avahi/docs/eos-updater-avahi.8
krnowak: I changed it to a full path, so I dropped the "in the same directory" part.

eos-updater-avahi/docs/eos-updater-avahi.8
krnowak: Done.

eos-updater-avahi/eos-updater-avahi.c
krnowak: Alright.

eos-updater-avahi/eos-updater-avahi.c
krnowak: Fixed.

libeos-updater-util/avahi-service-file.h
krnowak: Ok.

libeos-updater-util/avahi-service-file.h
krnowak: Alright, it's big endian from now on. I updated the two places where we create gvariants for repo index and summary timestamp. Will file a PR for your ostree branch later with the changes.

libeos-updater-util/avahi-service-file.h
krnowak: Added.

krnowak: Done. It's big-endian.

libeos-updater-util/avahi-service-file.h
krnowak: I wrote roughly because I thought we do something more than just hashing the element `k` times. Fixed.

libeos-updater-util/avahi-service-file.h
krnowak: Thanks for the suggestion. Applied.

libeos-updater-util/avahi-service-file.h
krnowak: Uh, I think it was because enums are a bit underspecified (what size are they?), so they may need casting when passed to varargs functions like `g_variant_new`.

libeos-updater-util/extensions.c
krnowak: Oops, fixed.

libeos-updater-util/avahi-service-file.c
krnowak: Fixed.

libeos-updater-util/avahi-service-file.c
krnowak: Fixed.

libeos-updater-util/avahi-service-file.c
krnowak: Fixed.

libeos-updater-util/avahi-service-file.c
krnowak: Fixed. I didn't bother with `const gsize[]` array, because it wouldn't work for custom size, and for lax validation I just do `return TRUE`.

krnowak:
Just pushed the fixes, more may follow up during testing.

libeos-updater-util/avahi-service-file.h
pwithnall: IIRC an enum is defined as the smallest signed int type which will contain all the members’ values (I might be wrong here). I agree it’s not immediately obvious though. I’d prefer to put all these values in an enum, cast when passing it to GVariant functions, and add a `G_STATIC_ASSERT (sizeof (TheEnum) <= sizeof (guint8))` or so.

pwithnall: I’m wrong, enums are defined as ints, with some caveats:
http://stackoverflow.com/questions/366017/what-is-the-size-of-an-enum-in-c

krnowak:
Pushed some fixes, I still need to make some proper commits out of them. I still did not address the enum-instead-of-defines review issue. At least this code works together with my ostree changes in https://github.com/pwithnall/ostree/pull/2.

libeos-update-server/repo.c
```diff
@@ -814,14 +861,18 @@ handle_path (EusRepo     *self,
...
   if (strstr (path, "..") != NULL)
     soup_message_set_status (msg, SOUP_STATUS_FORBIDDEN);
   else if (g_str_has_prefix (path, "/objects/") && g_str_has_suffix (path, ".filez"))
     handle_objects_filez (self, msg, path);
   else if (path_is_handled_as_is (path))
     handle_as_is (self, msg, path);
-  else if (g_strcmp0 (path, "/config") == 0)
+  else if (g_str_equal (path, "/config"))
     handle_config (self, msg);
+  else if (path_is_summary(path))
```
pwithnall: Missing space before `(`.

krnowak:
Updated. I hopefully addressed all the review issues, rebased it on top of current master, updated the ostree bloom implementation to the newest one and prepared the commits.

krnowak:
The only issue left is this eos-updater-avahi.py test that fails for me. Will investigate it tomorrow.

pwithnall:
This looks good, though given the change in architecture being [discussed on the mailing list](https://mail.gnome.org/archives/ostree-list/2017-May/msg00027.html) we can’t merge this until the new architecture has been finalised, implemented in OSTree, and this PR tweaked to make use of it. :-(

krnowak:
I updated it with a most likely broken commit, but the bulk of work should be there. I copied the bloom hash func for OstreeCollectionRef to ostree bloom files. I also copied the ostree-ref.[ch] and modified them slightly to avoid including libglnx and ostree-repo-private.h headers.

mwleeds:
I pushed a fixed version of this to the branch "mwleeds/T16589" that compiles and passes the tests. So that's ready for CR.

libeos-updater-util/avahi-service-file.c
```diff
@@ -248,12 +518,699 @@ eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+ * @refs_to_advertise: an array of refs to advertise
+ * @summary_timestamp: timestamp of the repo summary
+ * @options: (nullable): vardict #GVariant
+ * @cancellable: (nullable): a #GCancellable
+ * @error: return location for a #GError
+ *
+ * Create a `.service` file in @avahi_service_directory for the updater. This
+ * instructs Avahi to advertise a DNS-SD service for the updater, with TXT
+ * records indicating this machine has the refs for @ostree_path available with
+ * a commit at @head_commit_timestamp.
```
pwithnall: This documentation seems outdated — `ostree_path` and `head_commit_timestamp` are no longer arguments to this function. We now have `refs_to_advertise`.

libeos-updater-util/avahi-service-file.c
```diff
@@ -248,12 +518,699 @@ eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+ * should be something that is provided by the source of the summary
+ * (like metadata in the summary). As a fallback, a modification time
+ * of the locally stored summary file could be used, but it is rather
+ * fragile.
+ *
+ * @options can contain various options, which are dependent on the
+ * version of DNS-SD records. For the details, start reading about the
+ * %EOS_OSTREE_AVAHI_OPTION_FORCE_VERSION_Y option. If @options is
+ * %NULL, default values will be used instead. Default values are
+ * described in each options' documentation.
```
pwithnall: s/each options'/each option’s/

libeos-updater-util/avahi-service-file.h
```diff
@@ -44,4 +46,197 @@ gboolean eos_avahi_service_file_delete (const gchar   *avahi_service_directory,
...
+ * Default value of this option (if not overridden) is
+ * %EOS_OSTREE_AVAHI_BLOOM_HASH_ID_OSTREE_COLLECTION_REF.
+ */
+#define EOS_OSTREE_AVAHI_OPTION_BLOOM_HASH_ID_Y "bloom-hash-id"
+/**
+ * EOS_OSTREE_AVAHI_OPTION_BLOOM_K_Y:
+ *
+ * Specifies the k parameter for the bloom filter. It translates to
+ * how many times an element will be hashed before using it to set a
+ * bit in the bloom filter.
```
pwithnall: > It translates to how many times an element will be hashed before using it to set a bit in the bloom filter.

That’s not really correct. The `k` parameter determines how many different hash functions are used. That does mean ‘how many times an element will be hashed’, but the bits which each hash function invocation returns are all set in the bloom filter; not a single bit.

pwithnall:
> I pushed a fixed version of this to the branch "mwleeds/T16589" that compiles and passes the tests. So that's ready for CR.

Since that branch isn’t part of this PR, I can’t easily review it. I think you should be able to force-push to Krzesimir’s branch and push your fixups there so they get included in this PR. Could you try that?

I’ll add comments on specific commits in your branch, which we can hopefully then address in this PR:
 - https://github.com/endlessm/eos-updater/commit/5526c5c06d74f11fa07fa128f72540801e99d182#commitcomment-23636281
 - https://github.com/endlessm/eos-updater/commit/a409c2bd67c5cba76caa22897cdd6f977a2f6d45#commitcomment-23636298
 - https://github.com/endlessm/eos-updater/commit/44895e714465ff1e8abc9bcd04698ee5467b75d2#commitcomment-23636313
 - https://github.com/endlessm/eos-updater/commit/e487d0c2d3d82c7bb37512c0eae367a5e17e7cdd#commitcomment-23636346
 - https://github.com/endlessm/eos-updater/commit/e487d0c2d3d82c7bb37512c0eae367a5e17e7cdd#commitcomment-23636346

krnowak:
> Since that branch isn’t part of this PR, I can’t easily review it. I think you should be able to force-push to Krzesimir’s branch and push your fixups there so they get included in this PR. Could you try that?

That won't work unless @mwleeds is the maintainer of the project, no?

libeos-updater-util/avahi-service-file.h
krnowak: https://github.com/endlessm/eos-updater/pull/95#discussion_r116693111 ;)

mwleeds:
@krnowak right I can't push to your `eos-updater` fork, so I opened #101 instead

libeos-updater-util/avahi-service-file.h
pwithnall: Heh, snookered.

pwithnall:
> @krnowak right I can't push to your eos-updater fork, so I opened #101 instead

Thanks, let’s continue the review there. I’ll close this. Thanks for your updates as well, Krzesimir!

libeos-updater-util/avahi-service-file.c
```diff
@@ -37,67 +39,319 @@ const gchar * const EOS_UPDATER_AVAHI_SERVICE_TYPE = "_eos_updater._tcp";
...
 static gchar *
-txt_records_to_string (const gchar **txt_records)
+txt_records_to_string (GVariant      *txt_records,
+                       GVariantDict  *options_dict,
+                       GError       **error)
 {
+  GVariantIter iter;
   g_autoptr(GString) str = NULL;
-  gsize idx;
+  GVariant *txt_value;
```
mwleeds: Similarly for `txt_key`

PR #97, opened 2017-06-01
 Follow ref redirects in commit metadata 

dsd:
https://phabricator.endlessm.com/T17339


src/eos-updater-poll-common.c
```diff
@@ -742,14 +751,59 @@ fetch_latest_commit (OstreeRepo *repo,
...
+
+  if (!ostree_repo_load_variant (repo,
+                                 OSTREE_OBJECT_TYPE_COMMIT,
+                                 checksum,
+                                 &commit,
+                                 error))
+    return FALSE;
+
+  /* If this is a redirect commit, follow it and fetch the new ref instead */
+  g_variant_get_child (commit, 0, "@a{sv}", &metadata);
```
pwithnall: Nitpick: This could be simplified to `metadata = g_variant_get_child_value (commit, 0)` (which is equivalent).

src/eos-updater-poll-common.c
```diff
@@ -742,14 +751,59 @@ fetch_latest_commit (OstreeRepo *repo,
...
+  rebase = g_variant_lookup_value (metadata, "ostree.endoflife-rebase", G_VARIANT_TYPE_STRING);
+
+  if (rebase)
```
pwithnall: Nitpick: The code style in eos-updater is to be explicit about `NULL` comparisons, so this should be `if (rebase != NULL)`.

PR #99, opened 2017-06-12
Add ostree-tests as a build dependency

GeorgesStavracas:
The tests of eos-updater package depend on
ostree-trivial-httpd, which is provided by
ostree-test package.

This commit adds ostree-test as a build-time
dependency, so that Jenkins build passes all
the tests.

https://phabricator.endlessm.com/T17204

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -23,6 +23,7 @@ Build-Depends:
  libsoup2.4-dev,
  libsystemd-dev,
  ostree,
+ ostree-tests (>= 2017.6)
 
 Package: eos-updater
 Section: misc


debian/control
```diff
@@ -20,6 +20,7 @@ Build-Depends:
  libgsystem-dev,
  libnm-dev (>= 1.2.0),
  libostree-dev (>= 2016.15+dev24.a109440),
+ ostree-tests (>= 2017.6),
```
dbnicholson: Nitpick - can you put this at the end after `ostree` to maintain the alphabetical order?

debian/control
```diff
@@ -23,6 +23,7 @@ Build-Depends:
  libsoup2.4-dev,
  libsystemd-dev,
  ostree,
+ ostree-tests (>= 2017.6)
```
pwithnall: Nitpick: Can you leave the trailing comma on the list so we don’t get git diff noise when another dependency is added afterwards in future?

pwithnall:
> Might also be needed as a runtime dependency of eos-updater-tests, but I'll leave that to @pwithnall. I can't tell if the installed tests actually run `trivial-httpd` or not.

It doesn’t need to be a runtime dependency — the tests which spawn it are uninstalled-only at the moment.

debian/control
pwithnall: Looks like your merge raced with my review. Never mind.

dbnicholson: Sorry, I was thinking the same thing but decided not to nitpick further. Will push a trivial followup.

dbnicholson: Oh, @GeorgesStavracas already did it.

PR #100, opened 2017-06-20
src: Add a new error code for errors during fetching updates

pwithnall:
Override all unknown errors while fetching with the new code. This
should make failures a bit easier to read.

https://phabricator.endlessm.com/T17616


src/eos-updater-fetch.c
```diff
@@ -43,6 +43,14 @@ content_fetch_finished (GObject *object,
 
   if (error)
     {
+      if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_FAILED))
+        {
+          g_autofree gchar *old_message = error->message;
```
krnowak: You want to either steal it or duplicate it, otherwise `old_message` will become a dangling pointer after the next line. And you should get a crash after the double free.

pwithnall: Ouch, that’s embarrassing.

PR #101, opened 2017-08-14
new DNS-SD records (continued from #95)

mwleeds:
This is a continuation of PR #95 since I don't have write access to the branch on that one. After CR/QA, the fixup commits need to be squashed and the commit message on f3cf6b58b6c81f82b04a33de023fc0ea026b502b should be updated, then this can be merged.

### DIFF
--- a/libeos-updater-util/avahi-service-file.c
+++ b/libeos-updater-util/avahi-service-file.c
@@ -76,6 +76,7 @@ check_record_size (const gchar  *key,
 
 static gboolean
 handle_text_record (GString      *records_str,
+                    gsize        *total_size,
                     const gchar  *key,
                     GVariant     *text_value,
                     GError      **error)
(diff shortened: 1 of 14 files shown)


mwleeds:
I've addressed @pwithnall's review comments, and I've seen `ostree` accept the records generated by this code, so perhaps this could be merged.

libeos-updater-util/avahi-service-file.c
```diff
@@ -255,8 +255,8 @@ txt_records_to_string (GVariant      *txt_records,
 {
   GVariantIter iter;
   g_autoptr(GString) str = NULL;
-  GVariant *txt_value;
-  const gchar *txt_key;
+  g_autoptr(GVariant) txt_value = NULL;
+  g_autofree gchar *txt_key = NULL;
```
pwithnall: No, `txt_key` should remain as `const`, since the GVariant format string used for it is `&s`, not `s`.

pwithnall:
Apart from the one remaining review comment, this looks good to me. When you fix that one, please squash all the fixups (`git rebase -i --autosquash`), since github doesn’t do that automatically for us when we merge. Then I should be able to just hit the merge button on the branch.

However, note that it can’t be merged until https://phabricator.endlessm.com/T16919 is fixed and we have the updated libostree package in EOS OBS, since we depend on the fix for https://github.com/ostreedev/ostree/pull/1069, right?

mwleeds:
Alright I did the squashing and cleaned up Krnowak's last commit. So now this has to wait for ostreedev/ostree#1069 and `T16919`.

PR #102, opened 2017-08-17
euu/config: Expand config API to support per-key fallbacks

pwithnall:
This includes two trivial commits to fix some compiler warnings (one potentially impactful) which I spotted while working on this.

https://phabricator.endlessm.com/T18308


cosimoc:
@mwleeds has a lot on his plate and was happy with my review, so merging this.

PR #103, opened 2017-08-21
T18308 — fix tests to work with uninstalled config files

pwithnall:
Also improve the debug output from one of the integration tests to make this actually diagnosable in future.

Requesting review from @cosimoc and @mwleeds, but feel free to split it or delegate it amongst yourselves as per your current workloads. Thanks!

https://phabricator.endlessm.com/T18308

### DIFF
--- a/libeos-updater-util/config.c
+++ b/libeos-updater-util/config.c
@@ -24,6 +24,7 @@
  *  - Philip Withnall <withnall@endlessm.com>
  */
 
+#include <gio/gio.h>
 #include <glib.h>
 #include <glib-object.h>
 #include <libeos-updater-util/config.h>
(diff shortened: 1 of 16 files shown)


libeos-updater-util/config.c
```diff
@@ -108,25 +125,69 @@ euu_config_file_set_property (GObject      *object,
...
+  EuuConfigFile *self = EUU_CONFIG_FILE (object);
+  g_autoptr(GBytes) bytes = NULL;
+  g_autoptr(GError) error = NULL;
+
+  /* Chain up. */
+  G_OBJECT_CLASS (euu_config_file_parent_class)->constructed (object);
+
+  /* Load the default config file from the given resource. It’s a fatal error
+   * if this fails. We load this in the constructor to ensure we fail early,
+   * rather than conditionally on accessing something from the config file. */
```
cosimoc: You could also make this implement `GInitable` and enforce it that way -- but this is OK too.

pwithnall: I don’t think so. `GInitable` is for objects whose initialisation can fail at runtime. The whole point of having the defaults in a `GResource` is it means that they *can’t* fail at runtime (if they do, it’s a programmer error).

PR #104, opened 2017-09-08
eos-updater-poll-common: Remove unused variables

mwleeds:


pwithnall:
Looks good, although it would be best to wait for review (even on trivial patches) before pushing. :-)

PR #106, opened 2017-09-20
Build system fixes and improvements

pwithnall:
A few small fixes I needed to make while doing some QA locally.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -61,6 +61,7 @@ AM_CPPFLAGS = \
 noinst_LTLIBRARIES += libeos-updater-util/libeos-updater-util-@EUU_API_VERSION@.la
 
 libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_CFLAGS = \
+	$(AM_CFLAGS) \
 	$(CODE_COVERAGE_CFLAGS) \
 	$(WARN_CFLAGS) \
 	$(GIO_CFLAGS) \
@@ -104,6 +105,7 @@ eos_updater_avahi_eos_updater_avahi_CPPFLAGS = \
 	-DG_LOG_DOMAIN=\""eos-updater-avahi"\" \
 	$(NULL)
 eos_updater_avahi_eos_updater_avahi_CFLAGS = \
+	$(AM_CFLAGS) \
 	$(WARN_CFLAGS) \
 	$(CODE_COVERAGE_CFLAGS) \
 	$(EOS_UPDATER_AVAHI_CFLAGS) \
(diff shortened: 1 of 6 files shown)


Makefile.am
```diff
@@ -62,6 +62,7 @@ noinst_LTLIBRARIES += libeos-updater-util/libeos-updater-util-@EUU_API_VERSION@.
 
 libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_CFLAGS = \
 	$(AM_CFLAGS) \
+	$(STD_CFLAGS) \
```
mwleeds: Would it make sense to do `AM_CFLAGS = $(STD_CFLAGS) $(CODE_COVERAGE_CFLAGS) $(WARN_CFLAGS)` to reduce clutter?

libeos-updater-util/ostree-bloom.c
```diff
@@ -398,6 +398,12 @@ ostree_bloom_get_hash_func (OstreeBloom *bloom)
   return bloom->hash_func;
 }
 
+/* Ignore some warnings from the SipHash code rather than modifying it. */
+_Pragma ("GCC diagnostic push")
+_Pragma ("GCC diagnostic ignored \"-Wswitch-default\"")
+_Pragma ("GCC diagnostic ignored \"-Wimplicit-fallthrough\"")
```
mwleeds: This causes a warning to print on gcc versions less than 7. That should be fine; just noting it.

mwleeds:
I'm not sure why GitHub is marking one of my review comments as outdated; it's not.

libeos-updater-util/ostree-bloom.c
pwithnall: What warning? (I don’t have an old enough version of gcc to test with.) We should be able to add in a version check around the diagnostic line which is causing the warning.

Makefile.am
pwithnall: Yes and no. It would reduce clutter, but I find it makes it harder to work out which set of flags apply to a particular target, hence I prefer to be more explicit about it (unless dealing with a lot of very similar targets, like test binaries, where we can guarantee more uniformity). It is a bit of personal preference though.

libeos-updater-util/ostree-bloom.c
mwleeds:
```
libeos-updater-util/ostree-bloom.c:404:1: warning: unknown option after ‘#pragma GCC diagnostic’ kind [-Wpragmas]
 _Pragma ("GCC diagnostic ignored \"-Wimplicit-fallthrough\"")
 ^~~~~~~
```

pwithnall: Thanks. I’ve pushed an update which disables `-Wpragmas` too.

PR #107, opened 2017-09-20
fetch: Output slightly more information on a fetch failure

pwithnall:
This information is reported in the eos-updater D-Bus interface, but
it’s useful to have it in the journal too.

https://phabricator.endlessm.com/T19272

### DIFF
--- a/src/eos-updater-fetch.c
+++ b/src/eos-updater-fetch.c
@@ -208,7 +208,7 @@ content_fetch (GTask *task,
   goto cleanup;
 
  error:
-  g_message ("Fetch returning ERROR");
+  g_message ("Fetch: error: %s", error->message);
   g_task_return_error (task, error);
 
  cleanup:


src/eos-updater-fetch.c
```diff
@@ -208,7 +208,7 @@ content_fetch (GTask *task,
   goto cleanup;
 
  error:
-  g_message ("Fetch returning ERROR");
+  g_message ("Fetch: error: %s", error->message);
```
mwleeds: I think it will be printed when `eos_updater_set_error` gets called by `content_fetch_finished` but I suppose it doesn't hurt to print it twice.

pwithnall: Indeed. The alternative is to drop this message entirely. Printing ‘there was an error’ but not printing any details of the error is a bit rubbish. I’d be happy to drop this message if you think that’s a better approach?

PR #109, opened 2017-09-25
Fixes for compiling/analysing with Clang

pwithnall:
https://phabricator.endlessm.com/T19388

### DIFF
--- a/libeos-updater-util/util.c
+++ b/libeos-updater-util/util.c
@@ -180,7 +180,6 @@ eos_updater_save_or_delete  (GBytes *contents,
                              GError **error)
 {
   g_autoptr(GFile) target = g_file_get_child (dir, filename);
-  g_autoptr(GFile) target_parent = g_file_get_parent (target);
 
   if (contents == NULL)
     return delete_files_and_empty_parents (dir, target, cancellable, error);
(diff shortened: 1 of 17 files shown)


src/eos-prepare-usb-update.c
```diff
@@ -329,7 +331,9 @@ pull_ready (GObject *object,
             GAsyncResult *res,
             gpointer pull_data_ptr)
 {
-  g_autoptr(EosPullData) pull_data = EOS_PULL_DATA (pull_data_ptr);
+  g_autoptr(EosPullData) pull_data = pull_data_ptr;
+
+  g_return_if_fail (EOS_IS_PULL_DATA (pull_data));
```
smspillaz: Surely we want to run the `EOS_IS_PULL_DATA` check here before assigning the autoptr? (if the pointed-to object is not a `GObject`, calling `g_object_unref` on it will below up .... that said, we would have already hit programmer error in this case...)

libeos-updater-util/util.c
```diff
@@ -385,7 +385,8 @@ static void
 quit_clear_user_data (EosQuitFile *quit_file)
 {
   gpointer user_data = g_steal_pointer (&quit_file->user_data);
-  GDestroyNotify notify = g_steal_pointer (&quit_file->notify);
+  GDestroyNotify notify = quit_file->notify;
+  quit_file->notify = NULL;
```
smspillaz: Can't just explicitly cast it instead?

pwithnall: No, clang still complains.

src/eos-prepare-usb-update.c
pwithnall: If the check fails, we’re hosed anyway.

PR #110, opened 2017-09-30
T19388: Fix OSTree bloom hashing (and tests) on ARM

pwithnall:
T19388 turns out to have been caused by an implicit cast from a 64-bit integer to a 32-bit one. Some of the top 32 bits turned out to actually contain useful information.

https://phabricator.endlessm.com/T19388

### DIFF
--- a/eos-update-server/eos-update-server.c
+++ b/eos-update-server/eos-update-server.c
@@ -91,7 +91,7 @@ local_port_goption (const gchar *option_name,
                    "Invalid port number %s: %s", value, local_error->message);
       return FALSE;
     }
-  options->local_port = number;
+  options->local_port = (guint16) number;
   return TRUE;
 }
 
@@ -242,7 +242,7 @@ timeout_data_setup_timeout (TimeoutData *data)
 {
   clear_source (&data->timeout_id);
   if (data->timeout_seconds > 0)
-    data->timeout_id = g_timeout_add_seconds (data->timeout_seconds, timeout_cb, data);
+    data->timeout_id = g_timeout_add_seconds ((guint) data->timeout_seconds, timeout_cb, data);
 }
 
 static const gchar *
@@ -269,7 +269,7 @@ timeout_data_maybe_setup_quit_file (TimeoutData *data,
 {
   const gchar *filename = quit_file_name ();
   g_autoptr(EosQuitFile) quit_file = NULL;
-  gint timeout_seconds = 5;
+  guint timeout_seconds = 5;
 
   if (filename == NULL)
     return TRUE;
@@ -284,7 +284,7 @@ timeout_data_maybe_setup_quit_file (TimeoutData *data,
     return FALSE;
 
   data->quit_file = g_steal_pointer (&quit_file);
-  data->quit_file_timeout_seconds = timeout_seconds;
+  data->quit_file_timeout_seconds = (gint) timeout_seconds;
   return TRUE;
 }
 
--- a/eos-update-server/eos-update-server.c
+++ b/eos-update-server/eos-update-server.c
@@ -102,7 +102,6 @@ serve_remote_goption (const gchar *option_name,
                       GError **error)
 {
   Options* options = options_ptr;
-  g_autoptr(GError) local_error = NULL;
   g_autofree gchar *remote = NULL;
   g_autofree gchar *test_refspec = g_strdup_printf ("%s:test", value);
 
--- a/eos-updater-avahi/eos-updater-avahi.c
+++ b/eos-updater-avahi/eos-updater-avahi.c
@@ -91,8 +91,13 @@ get_raw_summary_timestamp_from_metadata (GBytes    *summary,
       return TRUE;
     }
 
+  /* FIXME: Disable diagnostics for GUINT64_TO_BE():
+   * https://bugzilla.gnome.org/show_bug.cgi?id=788384 */
+_Pragma ("GCC diagnostic push")
+_Pragma ("GCC diagnostic ignored \"-Wsign-conversion\"")
   *out_found = TRUE;
   *out_timestamp = GUINT64_FROM_BE (g_variant_get_uint64 (value));
+_Pragma ("GCC diagnostic pop")
   return TRUE;
 }
 
@@ -222,7 +227,7 @@ update_service_file (gboolean       advertise_updates,
 
   if (commit_checksum != NULL)
     {
-      commit_date_time = g_date_time_new_from_unix_utc (commit_timestamp);
+      commit_date_time = g_date_time_new_from_unix_utc ((gint64) commit_timestamp);
       formatted_commit_date_time = g_date_time_format (commit_date_time,
                                                        "%FT%T%:z");
     }
--- a/libeos-update-server/server.c
+++ b/libeos-update-server/server.c
@@ -255,15 +255,20 @@ update_pending_requests (EusServer *self,
   GObject *obj = G_OBJECT (self);
 
   g_assert (delta >= 0 || self->pending_requests >= (guint) -delta);
-  g_assert (delta <= 0 || self->pending_requests <= G_MAXUINT - delta);
+  g_assert (delta <= 0 || self->pending_requests <= G_MAXUINT - (guint) delta);
 
   if (delta == 0)
     return;
 
-  g_debug ("%s: Updating from %u to %u", G_STRFUNC, self->pending_requests,
-           self->pending_requests + delta);
+  g_debug ("%s: Updating from %u by %d", G_STRFUNC, self->pending_requests, delta);
 
-  self->pending_requests += delta;
+  /* Silence a -Wconversion warning. @delta may be negative, but
+   * @pending_requests may be outside the range we can losslessly cast to
+   * (gint), so we have to do everything in terms of (guint)s. */
+  if (delta > 0)
+    self->pending_requests += (guint) delta;
+  else
+    self->pending_requests -= (guint) -delta;
   self->last_request_time = g_get_monotonic_time ();
 
   g_object_freeze_notify (obj);
--- a/libeos-updater-util/avahi-service-file.c
+++ b/libeos-updater-util/avahi-service-file.c
@@ -1144,7 +1144,6 @@ iterate_and_remove_ostree_service_files (GFileEnumerator  *enumerator,
     {
       GFileInfo *file_info;
       GFile* file;
-      g_autoptr(GError) local_error = NULL;
       g_autoptr(GMatchInfo) match_info = NULL;
       g_autofree gchar *matched_repository_index = NULL;
       const gchar *filename;
--- a/libeos-updater-util/avahi-service-file.c
+++ b/libeos-updater-util/avahi-service-file.c
@@ -99,8 +99,8 @@ handle_text_record (GString      *records_str,
    */
   g_assert (total_size != NULL);
   *total_size += 1 + record_size;
-  escaped_key = g_markup_escape_text (key, key_length);
-  escaped_value = g_markup_escape_text (value_string, value_string_length);
+  escaped_key = g_markup_escape_text (key, (gssize) key_length);
+  escaped_value = g_markup_escape_text (value_string, (gssize) value_string_length);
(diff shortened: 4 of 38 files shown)


pwithnall:
This shouldn’t be merged yet, since I still need to do a bit more work to get eos-updater fully compiling with `-Wconversion`. What’s here already is ready to review, though, since the remaining work is additive.

pwithnall:
Finished — `make check` now passes in my ARM chroot. Ready for review and merge.

libeos-updater-util/ostree-bloom.c
```diff
@@ -275,11 +275,11 @@ ostree_bloom_maybe_contains (OstreeBloom   *bloom,
 
   for (i = 0; i < bloom->k; i++)
     {
-      gsize idx;
+      guint64 idx;
 
       idx = bloom->hash_func (element, i);
 
-      if (!ostree_bloom_get_bit (bloom, idx % (bloom->n_bytes * 8)))
+      if (!ostree_bloom_get_bit (bloom, (gsize) (idx % (bloom->n_bytes * 8))))
```
smspillaz: I'm guessing that we assume here that `bloom->n_bytes * 8` will never overflow here (perhaps a range check in `ostree_bloom_new` is necessary?)

eos-update-server/eos-update-server.c
```diff
@@ -242,7 +242,7 @@ timeout_data_setup_timeout (TimeoutData *data)
 {
   clear_source (&data->timeout_id);
   if (data->timeout_seconds > 0)
-    data->timeout_id = g_timeout_add_seconds (data->timeout_seconds, timeout_cb, data);
+    data->timeout_id = g_timeout_add_seconds ((guint) data->timeout_seconds, timeout_cb, data);
```
smspillaz: It seems like `TimeoutData::seconds` is never set to a negative value anywhere - can we just make it `guint`?

eos-updater-avahi/eos-updater-avahi.c
```diff
@@ -222,7 +227,7 @@ update_service_file (gboolean       advertise_updates,
 
   if (commit_checksum != NULL)
     {
-      commit_date_time = g_date_time_new_from_unix_utc (commit_timestamp);
+      commit_date_time = g_date_time_new_from_unix_utc ((gint64) commit_timestamp);
```
smspillaz: Unfortunate that `g_date_time_new_from_unix_utc` takes `gint64` here instead of `guint64`. Should we do a range check to make sure that we don't end up with a negative value and instead wrap around?

libeos-update-server/server.c
```diff
@@ -255,15 +255,14 @@ update_pending_requests (EusServer *self,
...
 
   if (delta == 0)
     return;
 
-  g_debug ("%s: Updating from %u to %u", G_STRFUNC, self->pending_requests,
-           self->pending_requests + delta);
+  g_debug ("%s: Updating from %u by %d", G_STRFUNC, self->pending_requests, delta);
 
-  self->pending_requests += delta;
+  self->pending_requests += (guint) delta;
```
smspillaz: It looks like `delta` can be negative here here (based on the `g_assert` above) but it is cast to a `guint`

libeos-updater-util/avahi-service-file.c
```diff
@@ -99,8 +99,8 @@ handle_text_record (GString      *records_str,
    */
   g_assert (total_size != NULL);
   *total_size += 1 + record_size;
-  escaped_key = g_markup_escape_text (key, key_length);
-  escaped_value = g_markup_escape_text (value_string, value_string_length);
+  escaped_key = g_markup_escape_text (key, (gssize) key_length);
```
smspillaz: It might be worth bounds-checking `key_length` and `value_string_length` here as they come from an uncontrolled input.

tests/test-update-from-main.c
```diff
@@ -47,6 +47,15 @@ test_update_from_main (EosUpdaterFixture *fixture,
...
   g_autoptr(GVariant) main_source_variant = NULL;
 
+  /* We could get OSTree working by setting OSTREE_BOOTID, but shortly
+   * afterwards we hit unsupported syscalls in qemu-user when running in an
+   * ARM chroot (for example), so just bail. */
+  if (!eos_test_has_ostree_boot_id ())
+    {
+      g_test_skip ("OSTree will not work without a boot ID");
+      return;
+    }
```
smspillaz: I wonder if we can use a GTestFixture to implement this for all tests as opposed to just some of them (mental note that I need to do this in my own branch too)

eos-updater-avahi/eos-updater-avahi.c
pwithnall: I’m not too fussed about the difference between `guint64` and `gint64` for timestamps. We’re going to hit the year 2038 problem for `gint32` in 20 years; the problems with 64-bit integers are waaaaaay off.

libeos-updater-util/avahi-service-file.c
pwithnall: They’re checked as part of `record_size` by the `check_record_size()` call above.

tests/test-update-from-main.c
pwithnall: We can’t, but we don’t want to. I want as many tests as possible to run in an ARM chroot, and to only skip the ones which will spuriously fail.

We can’t use `g_test_skip()` from a fixture setup function, but that’s something which would ideally be [fixed in GLib](https://bugzilla.gnome.org/show_bug.cgi?id=788419).

libeos-update-server/server.c
pwithnall: I *think* this will do what we want, since it’s just making the cast which was previously implicit, explicit. In two’s-complement terms, the addition will deliberately wrap around and give the right answer. However, I will double-check.

eos-update-server/eos-update-server.c
pwithnall: It comes from the `--timeout` command line option (see `options_init()`), which can legitimately be negative.

libeos-updater-util/ostree-bloom.c
pwithnall: Ah, indeed. I thought we had one of those, but I was wrong. I’ll push a fixup. Thanks for checking thoroughly. :-)

pwithnall: Fixup pushed. We don’t need it on `eos3.3` since it doesn’t fix any demonstrable bugs.

libeos-update-server/server.c
pwithnall: Yes, this does do what we want, although I agree that it’s quite counterintuitive. I’ll add a comment.

pwithnall: Done.

pwithnall:
All comments answered or addressed with additional commits. Ready for review again, thanks.

eos-update-server/eos-update-server.c
smspillaz: Ah, got it (-1 means no timeout, which is reasonable).

eos-updater-avahi/eos-updater-avahi.c
smspillaz: Yeah, fair enough

tests/test-update-from-main.c
smspillaz: Ah, I see. Yes, it would be nice if that could be fixed in GLib, indeed.

mwleeds:
@pwithnall do these not show up for you?
```
libeos-updater-util/util.c:485:26: warning: implicit conversion loses integer precision: 'gulong' (aka 'unsigned long') to 'guint'
      (aka 'unsigned int') [-Wshorten-64-to-32]
  quit_file->signal_id = g_signal_connect (quit_file->monitor,
                       ~ ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/glib-2.0/gobject/gsignal.h:475:5: note: expanded from macro 'g_signal_connect'
    g_signal_connect_data ((instance), (detailed_signal), (c_handler), (data), NULL, (GConnectFlags) 0)
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

```
eos-autoupdater.c:285:21: warning: implicit conversion changes signedness: 'gint' (aka 'int') to 'UpdateStep' (aka 'enum _UpdateStep')
      [-Wsign-conversion]
  UpdateStep step = GPOINTER_TO_INT (step_data);
             ~~~~   ^~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/lib/x86_64-linux-gnu/glib-2.0/include/glibconfig.h:87:29: note: expanded from macro 'GPOINTER_TO_INT'
#define GPOINTER_TO_INT(p)      ((gint)  (glong) (p))
                                 ^~~~~~~~~~~~~~~~~~~
```

```
eos-test-utils.c:1260:61: warning: implicit conversion loses integer precision: 'gsize' (aka 'unsigned long') to 'guint' (aka 'unsigned int')
      [-Wshorten-64-to-32]
  g_autoptr(GPtrArray) source_strs = g_ptr_array_sized_new (n_sources);
                                     ~~~~~~~~~~~~~~~~~~~~~  ^~~~~~~~~
```

pwithnall:
Hmm, weird. They don’t show up with the ARM chroot compiler, but do show up when I rebuild from scratch on x86-64. I thought I’d done that. Good catch!

pwithnall:
Fixes pushed. Ready for re-review.

smspillaz:
(last few commits are an r+ from me)

pwithnall:
For fun, I just tried compiling with `CC=clang` on x86-64, and found a few more `-Wconversion` failures. Fixes pushed for them.

src/eos-autoupdater.c
```diff
@@ -249,8 +251,8 @@ update_stamp_file (guint64 last_successful_update_secs,
 
   if (randomized_delay_days > 0)
     {
-      gint32 actual_delay_days = g_random_int_range (0, randomized_delay_days + 1);
-      mtime.tv_sec += actual_delay_days * SEC_PER_DAY;
+      gint32 actual_delay_days = g_random_int_range (0, (gint32) randomized_delay_days + 1);
```
mwleeds: I think this could be negative if for example the user sets the delay to `G_MAXUINT32 - 1`, although that shouldn't happen in practice. And I think it would only result in checking for updates more often than intended.

libeos-update-server/server.c
```diff
@@ -255,15 +255,17 @@ update_pending_requests (EusServer *self,
...
     return;
 
-  g_debug ("%s: Updating from %u to %u", G_STRFUNC, self->pending_requests,
-           self->pending_requests + delta);
+  g_debug ("%s: Updating from %u by %d", G_STRFUNC, self->pending_requests, delta);
 
-  self->pending_requests += delta;
+  /* Silence a -Wconversion warning. @delta may be negative, in which case the
+   * beauty of two’s-complement will give us the result we’d expect if the
+   * (guint) cast was not explicit. */
```
mwleeds: I'm surprised we're relying on technically undefined behavior but I guess if that's how practically every compiler and architecture works it's fine.

pwithnall: Ooops, you’re right, C doesn’t technically have to be implemented using two’s-complement. I’ll rework the commit to not rely on that. This is getting ridiculous though.

src/eos-autoupdater.c
pwithnall: `randomized_delay_days` is limited to [0, (`G_MAXINT32` / `SEC_PER_DAY`) - 1] on input, so should always be safe to convert to `gint32` and add 1 to.

libeos-update-server/server.c
pwithnall: Reworked in commit dc692c96cbd0a2c3e2ebafa3b6b0676a58f201bd.

src/eos-autoupdater.c
mwleeds: Oops, I missed that

PR #114, opened 2017-10-18
lib/extensions: Drop support for refs.d directories

pwithnall:
With the new unsigned summary support and ostree-metadata branch, we
don’t need refs.d directories. They were never supported by our
infrastructure anyway, so were never used in practice.

https://phabricator.endlessm.com/T15926

---

This depends on https://github.com/endlessm/eos-updater/pull/113 being merged first, since it’s stacked on top of that. The only relevant commit in this PR is the top-most one.

### DIFF
--- a/libeos-updater-util/extensions.c
+++ b/libeos-updater-util/extensions.c
@@ -27,187 +27,10 @@
 #include <string.h>
 
 static void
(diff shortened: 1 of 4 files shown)


pwithnall:
Rebased and merged, thanks.

PR #116, opened 2017-10-24
Port eos-updater to use OstreeRepoFinder

pwithnall:
A whole bunch of commits necessary for this to work, and to get the tests working again.

https://phabricator.endlessm.com/T19293


src/eos-updater-poll-common.c
```diff
@@ -372,32 +427,27 @@ fetch_latest_commit (OstreeRepo *repo,
...
-      options = get_repo_pull_options (url_override, ref);
-      if (!ostree_repo_pull_with_options (repo,
-                                          remote_name,
-                                          options,
-                                          NULL,
-                                          cancellable,
-                                          error))
-        return FALSE;
-
       g_clear_pointer (&checksum, g_free);
```
smspillaz: Hmm, clearing the checksum outparam here as a way of indicating that we hit a rebase feels like a non-intuitive design to me - perhaps we could have a function that just returns the ref to "follow" if we need to follow a new one and then another one which actually fetches the checksum.

src/eos-updater-fetch.c
```diff
@@ -165,6 +248,31 @@ content_fetch (GTask *task,
...
+  if (data->results != NULL)
+    {
+      g_message ("Fetch: using results %p", data->results);
+
+      if (content_fetch_new (updater, data, task_context, cancellable, &error))
+        {
+          g_message ("Fetch: finished pulling using libostree P2P code");
+          g_task_return_boolean (task, TRUE);
+
+          goto cleanup;
```
smspillaz: I wonder if instead of using `goto` here, we can split the content-fetch-via-ostree and content-fetch-via-old-code parts into separate functions and dispatch from here.

src/eos-updater-poll-common.c
```diff
@@ -652,12 +659,30 @@ eos_update_info_to_string (EosUpdateInfo *update)
...
   timestamp_str = g_date_time_format (timestamp, "%FT%T%:z");
 
-  return g_strdup_printf ("%s, %s, %s, %s\n   %s",
+  results_string = g_string_new ("");
+  for (i = 0; update->results != NULL && update->results[i] != NULL; i++)
+    {
+      const OstreeRepoFinderResult *result = update->results[i];
+
+      if (i != 0)
+        g_string_append (results_string, "\n   ");
```
smspillaz: If we want all the lines of this to be indented, then the whitespace padding should probably go on the `g_string_append_printf` line below.

src/eos-updater-poll.c
```diff
@@ -203,6 +203,192 @@ get_fetchers (SourcesConfig *config,
...
+    {
+      switch (g_array_index (config->download_order, EosUpdaterDownloadSource, i))
+        {
+        case EOS_UPDATER_DOWNLOAD_MAIN:
+          g_ptr_array_add (finders, ostree_repo_finder_config_new ());
+          break;
+
+        case EOS_UPDATER_DOWNLOAD_LAN:
+          finder_avahi = ostree_repo_finder_avahi_new (context);
+          g_ptr_array_add (finders, g_object_ref (finder_avahi));
```
smspillaz: I think `finder_avahi` will be leaked here if we have multiple lan sources - it makes more sense to me to just pass `ostree_repo_finder_avahi_new` directly to `g_ptr_array_add`

smspillaz: Oh, hmm, I guess its only possible to have one `finder_avahi`. Maybe an assertion that it isn't already set?

src/eos-updater-poll.c
```diff
@@ -203,6 +203,192 @@ get_fetchers (SourcesConfig *config,
...
+        }
+    }
+
+  g_ptr_array_add (finders, NULL);  /* NULL terminator */
+
+  /* TODO: Stop this at some point; think of a better way to store it and
+   * control its lifecycle. */
+  if (finder_avahi != NULL)
+    ostree_repo_finder_avahi_start (OSTREE_REPO_FINDER_AVAHI (finder_avahi),
+                                    &local_error);
```
smspillaz: Does this return a `gboolean`? If so, can we check the result of that instead of checking `local_error` below?

src/eos-updater-poll.c
```diff
@@ -203,6 +203,192 @@ get_fetchers (SourcesConfig *config,
...
+
+  if (local_error != NULL)
+    {
+      g_warning ("Avahi finder failed; removing it: %s", local_error->message);
+      g_ptr_array_remove (finders, finder_avahi);
```
smspillaz: Maybe instead of adding `finder_avahi` to the array above and then removing it here, it might be better to capture `finder_avahi` in the variable above and then add it to the array if `ostree_repo_finder_avahi_start` succeeded.

src/eos-updater-poll.c
```diff
@@ -203,6 +203,192 @@ get_fetchers (SourcesConfig *config,
...
+  g_auto(OstreeRepoFinderResultv) results = NULL;
+  g_autoptr(EosUpdateInfo) info = NULL;
+  g_autofree gchar *checksum = NULL;
+  g_autoptr(GVariant) commit = NULL;
+  g_autofree gchar *booted_refspec = NULL, *new_refspec = NULL;
+  g_autoptr(OstreeCollectionRef) booted_collection_ref = NULL, new_collection_ref = NULL;
+  const gchar *upgrade_refspec;
+  const OstreeCollectionRef *upgrade_collection_ref;
+  g_autoptr(GPtrArray) finders = NULL;  /* (element-type OstreeRepoFinder) */
+  g_autoptr(OstreeRepoFinderAvahi) finder_avahi = NULL;
```
smspillaz: It looks like this is unused

src/eos-updater-poll.c
```diff
@@ -203,6 +203,192 @@ get_fetchers (SourcesConfig *config,
...
+  if (local_error != NULL)
+    {
+      g_warning ("Avahi finder failed; removing it: %s", local_error->message);
+      g_ptr_array_remove (finders, finder_avahi);
+      g_clear_object (&finder_avahi);
+      g_clear_error (&local_error);
+    }
+
+  if (out_finder_avahi != NULL)
+    *out_finder_avahi = g_steal_pointer (&finder_avahi);
```
smspillaz: Since `finder_avahi` does not appear to be used below, this outparam can probably be removed.

src/eos-updater-poll.c
```diff
@@ -203,6 +203,192 @@ get_fetchers (SourcesConfig *config,
...
+  /* Check whether the commit is a redirection; if so, fetch the new ref and
+   * check again. */
+  do
+    {
+      const OstreeCollectionRef *refs[] = { NULL, NULL };
+      g_autoptr(GVariant) pull_options = NULL;
+      g_autoptr(GAsyncResult) find_result = NULL, pull_result = NULL;
+      g_auto(GVariantBuilder) builder = G_VARIANT_BUILDER_INIT (G_VARIANT_TYPE ("a{sv}"));
+
+      refs[0] = upgrade_collection_ref;
```
smspillaz: Can we initialise `refs` directly with `{ upgrade_collection_ref, NULL }`?

src/eos-updater-poll.c
```diff
@@ -240,14 +428,49 @@ metadata_fetch (GTask *task,
...
-                       config.download_order);
+  /* Do we want to use the new libostree code for P2P, or fall back on the old
+   * eos-updater code?
+   * FIXME: Eventually drop the old code. See:
+   * https://phabricator.endlessm.com/T19606 */
+  if (use_new_code)
+    {
+      info = metadata_fetch_new (data->repo, &config, task_context, cancellable, &error);
+
+      if (error != NULL)
```
smspillaz: should we check `info == NULL` as opposed to `error != NULL` ?

src/eos-updater-poll.c
```diff
@@ -203,6 +203,192 @@ get_fetchers (SourcesConfig *config,
...
+  if (!is_checksum_an_update (repo, checksum, &commit, error))
+    return NULL;
+
+  if (commit == NULL)
+    return NULL;
+
+  info = eos_update_info_new (checksum, commit,
+                              upgrade_refspec, booted_refspec,
+                              NULL, g_steal_pointer (&results));
+  metrics_report_successful_poll (info);
```
smspillaz: I think this function is not defined in this commit (it is in the next one)

eos-updater-avahi/eos-updater-avahi.c
```diff
@@ -78,9 +78,20 @@ get_raw_summary_timestamp_from_metadata (GBytes    *summary,
...
+  g_autoptr(GVariant) additional_metadata = NULL;
   g_autoptr(GVariant) value = NULL;
 
-  value = g_variant_lookup_value (summary_variant, "ostree.summary.last-modified", G_VARIANT_TYPE_UINT64);
+  if (!g_variant_is_normal_form (summary_variant))
+    {
+      *out_found = FALSE;
+      *out_timestamp = 0;
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND,
+                   "Corrupt summary file");
```
smspillaz: Is `G_IO_ERROR_NOT_FOUND` the best error code for this? I've noticed that callers will often check `NOT_FOUND` as an indicator that some file needs to be created. Perhaps `G_IO_ERROR_FAILED` (or our own error code)?

src/eos-updater-poll.c
```diff
@@ -132,6 +135,15 @@ read_config (const gchar *config_file_path,
...
     return FALSE;
 
+  /* FIXME: For the moment, this is undocumented and hidden. It can also be set
+   * via the PollVolume() D-Bus method. It must be non-empty if set. */
+  sources_config->override_uris = euu_config_file_get_strv (config,
+                                                            DOWNLOAD_GROUP,
+                                                            "OverrideUris", NULL,
+                                                            error);
+  if (sources_config->override_uris != NULL && sources_config->override_uris[0] == NULL)
+    g_clear_pointer (&sources_config->override_uris, g_strfreev);
```
smspillaz: Not sure I understand this - why do we get the strv from the config file and then free it straight away?

src/eos-updater-poll-common.c
pwithnall: Yeah, it would be nice to rework this a bit. I’d prefer to do that as a follow-up commit, since I was just extending the existing behaviour rather than introducing new behaviour here.

src/eos-updater-fetch.c
pwithnall: My thinking was to try and minimise the number of changes to the old code path, at least to begin with, to reduce the chance of accidentally breaking it. I might do a follow-up commit to refactor this a bit.

src/eos-updater-poll-common.c
pwithnall: The initial indentation is part of the final `g_strdup_printf()` call (with the `return`), below. It would be simpler to refactor as you suggest though.

src/eos-updater-poll.c
pwithnall: It does not. I could change it to in upstream OSTree, but I don’t want to block this branch on getting another change into OSTree (even if it is trivial).

src/eos-updater-poll.c
pwithnall: Oops, I didn’t check compilability between commits after rebasing and squashing.

eos-updater-avahi/eos-updater-avahi.c
pwithnall: Callers should do whatever’s documented for the method they’re calling. I don’t think we’re returning `G_IO_ERROR_NOT_FOUND` from elsewhere in this function, so it’s distinct. I’ll have a think about whether another code is more appropriate.

src/eos-updater-poll.c
pwithnall: We only free it if it’s empty. These two lines basically mean we can treat (`override_uris == NULL`) and (`override_uris` is empty) as identical cases in the rest of the code, rather than having to check for both.

pwithnall: I could add a comment.

src/eos-updater-fetch.c
pwithnall: Pushed a follow-up commit.

src/eos-updater-poll-common.c
pwithnall: Pushed a follow-up commit.

src/eos-updater-poll.c
smspillaz: Makes sense

src/eos-updater-poll-common.c
pwithnall: Fixup pushed.

eos-updater-avahi/eos-updater-avahi.c
smspillaz: I think the main problem is that its easy for `G_IO_ERROR_NOT_FOUND` to collide:

    gboolean some_func (..., GError **error)
      {
        if (!g_file_get_contents ("some-file", &contents, &error))
          return FALSE;

        if (!get_raw_summary_timestamp_from_metadata(..., &error))
          return FALSE;
      }

Callers of `some_func` might respond to `G_IO_ERROR_NOT_FOUND` by creating the file requested by `g_file_get_contents` and trying again (even though the issue was actually with`get_raw_summary_timestamp_from_metadata`)

src/eos-updater-poll.c
smspillaz: Ah, got it - I must have read this wrong.

src/eos-updater-poll.c
pwithnall: `strv_to_download_order()` checks that there are no duplicated sources, but I’ll add an assertion to make it clearer.

src/eos-updater-poll.c
pwithnall: I can’t see any advantage in doing that; both approaches are ugly, as far as I see. The whole code needs refactoring, but I didn’t want to do that until the old code paths have been dropped. I’ll add a `FIXME` about that.

src/eos-updater-poll.c
pwithnall: Oops, I need to stop it searching once polling is complete.

src/eos-updater-poll.c
pwithnall: Fixup pushed to actually use `finder_avahi`.

src/eos-updater-poll.c
pwithnall: Fixup pushed.

src/eos-updater-poll.c
pwithnall: `FIXME` fixup pushed.

src/eos-updater-poll.c
pwithnall: Assertion added in a fixup.

src/eos-updater-poll.c
pwithnall: Indeed, this was left over from an earlier iteration of the code where that wasn’t possible. Fixup pushed.

src/eos-updater-poll.c
pwithnall: Nope, since it’s permitted for `metadata_fetch_new()` to return `NULL` and not set an error, in the case that no results were found.

eos-updater-avahi/eos-updater-avahi.c
pwithnall: Yeah, I’m satisfied that `G_IO_ERROR_NOT_FOUND` is appropriate here. If a caller wants to create the ‘missing’ file, they’ll call `ostree_repo_regenerate_summary()` which doesn’t care whether the `summary` file exists already.

src/eos-updater-poll.c
pwithnall: Pushed as a fixup.

pwithnall:
All review comments so far addressed, apart from the one about compile errors in intermediate commits. I’ll handle that one after squashing everything down after review.

src/eos-updater-poll-common.c
```diff
@@ -603,79 +376,87 @@ fetch_latest_commit (OstreeRepo *repo,
...
 
-      g_clear_pointer (&checksum, g_free);
-      if (!fetch_commit_checksum (repo,
-                                  cancellable,
-                                  remote_name,
-                                  ref,
-                                  url_override,
-                                  &checksum,
-                                  error))
+      if (!parse_latest_commit (repo, refspec, &redirect_followed, &checksum,
```
mwleeds: Aren't we leaking `checksum` here in the case that a redirect is followed?

src/eos-updater-poll.c
```diff
@@ -162,105 +135,253 @@ read_config (const gchar *config_file_path,
...
+        g_main_context_iteration (context, TRUE);
+
+      if (!ostree_repo_pull_from_remotes_finish (repo, pull_result, error))
+        return NULL;
+
+      g_clear_pointer (&new_refspec, g_free);
+      g_clear_pointer (&new_collection_ref, ostree_collection_ref_free);
+
+      /* Parse the commit and check there’s no redirection to a new ref. */
+      if (!parse_latest_commit (repo, upgrade_refspec, &redirect_followed,
```
mwleeds: Similar to the other case, I think we're leaking `checksum`

src/eos-updater-poll-common.c
pwithnall: Oops, yes. I’ll push a fixup. Well spotted.

src/eos-updater-poll.c
pwithnall: Yup, fixup will be pushed shortly.

src/eos-updater-poll-common.c
```diff
@@ -996,23 +798,12 @@ run_fetchers (EosMetadataFetchData *fetch_data,
       g_autofree gchar *booted_ref = NULL;
       g_autoptr(GError) metrics_error = NULL;
```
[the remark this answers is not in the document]
pwithnall: Yup, I’ll push a fixup shortly.

pwithnall:
Thanks for the reviews. I’ve squashed everything in, and tested with `git for-each master.. make` that it builds fine in all the intermediate states. (I don’t expect the test suite to pass in each intermediate state, so I haven’t checked that.)

PR #118, opened 2017-10-27
debian: Add missing Python dependencies for eos-updater-prepare-volume

pwithnall:
Seems like dh_python is not doing its job and listing the Python
dependencies automatically. List the dependencies for
eos-updater-prepare-volume and eos-updater-ctl manually.

https://phabricator.endlessm.com/T19293

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -59,6 +59,10 @@ Architecture: any
 Multi-arch: no
 Depends:
  eos-updater (= ${binary:Version}),
+ gir1.2-flatpak-1.0,
+ gir1.2-glib-2.0,
+ gir1.2-ostree-1.0,
+ python3-gi,
  ${misc:Depends},
  ${python3:Depends},
  ${shlibs:Depends},


dbnicholson:
Just FYI, dh_python handles runtime python dependencies. There's no automatic build dependency generator.

pwithnall:
> Just FYI, dh_python handles runtime python dependencies. There's no automatic build dependency generator.

I guess the issue here is that the runtime dependencies for the Python scripts have to be a subset of the build-time dependencies, since the Python scripts are run in the build-time unit tests. Do you think the patch is correct, or is there a better way of including the runtime dependencies into the build-time ones automatically (without this duplication)?

Thanks for the FYI. :-)

dbnicholson:
I don't believe there's any way to automatically generate build time dependencies in debian. It's more or less expected that the source portion of the control file is static so that it can be analyzed directly without running external processes. There are some ways to hook into dpkg to make adjustments to the fields in the dsc file (which uses the control file as source), but that happens at the end of the build after the build dependencies have been analyzed. And I don't think it would be kosher to have the Build-Depends different between the dsc and debian/control file.

So, as far as I know, this is the correct and only way to to do this.

pwithnall:
(Follow-up which does this in #119.)

PR #121, opened 2017-11-02
Fix trying to list directories in eos-update-server

pwithnall:
A couple of minor fixes for problems spotted while starting to work on https://phabricator.endlessm.com/T19367.

### DIFF
--- a/libeos-update-server/repo.c
+++ b/libeos-update-server/repo.c
@@ -659,6 +659,7 @@ serve_file_if_exists (SoupMessage *msg,
   g_autoptr(GBytes) file_bytes = NULL;
   g_autoptr(GError) error = NULL;
   g_autoptr(SoupBuffer) buffer = NULL;
+  GFileType file_type;
 
   /* Security check to ensure we don’t get tricked into serving files which
    * are outside the document root. This canonicalises the paths but does not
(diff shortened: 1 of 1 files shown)


libeos-update-server/repo.c
```diff
@@ -680,18 +681,33 @@ serve_file_if_exists (SoupMessage *msg,
...
+
+      /* mmap() can legitimately fail if the underlying file system doesn’t
+       * support it, which can happen if we’re using an overlayfs. Fall back to
+       * reading in the file. */
+      g_clear_error (&error);
+      if (!g_file_get_contents (raw_path, (gchar **) &contents, &contents_len, &error))
+        {
+          g_warning ("Failed to load ‘%s’: %s", raw_path, error->message);
+          soup_message_set_status (msg, SOUP_STATUS_INTERNAL_SERVER_ERROR);
+          return FALSE;
```
mwleeds: We should have `*served = FALSE;` here, just to be safe even though all the callers initialize `served`.

pwithnall: I don’t think so — the code didn’t used to do this, and the convention in GLib is for all out-parameters to remain unchanged if the function fails. Since we can guarantee all the callers behave properly here (thanks for checking), I don’t want to include extra code to deal with the non-existent programmer error of if they don’t.

mwleeds: Oh, I wasn't aware of that convention

PR #123, opened 2017-11-06
libeos-update-server: Support /refs/mirrors/ requests

mwleeds:
When peers request collection refs from each other, they can use the
format /refs/mirrors/$collection_id/$ref to avoid having to assume that
the remote name is the same on both machines. This commit makes
eos-update-server handle these requests either directly if such a file
exists or by redirecting them to /refs/remotes/$remote_name/$ref if
the remote $remote_name has a collection ID that matches $collection_id.

https://phabricator.endlessm.com/T18718

### DIFF
--- a/libeos-update-server/repo.c
+++ b/libeos-update-server/repo.c
@@ -914,33 +914,34 @@ handle_refs_mirrors (EusRepo     *self,
   remotes = ostree_repo_remote_list (self->repo, NULL);
   if (remotes != NULL)
     {
-      for (i = 0; i < g_strv_length (remotes); ++i)
+      guint remotes_len = g_strv_length (remotes);
+      for (i = 0; i < remotes_len; ++i)
         {
-            g_autofree gchar *remote_collection_id = NULL;
-            gboolean served = FALSE;
+          g_autofree gchar *remote_collection_id = NULL;
 
-            ostree_repo_get_remote_option (self->repo, remotes[i], "collection-id", NULL, &remote_collection_id, &error);
-            if (local_error != NULL)
-              {
-                g_warning ("Error getting collection ID for remote %s: %s", remotes[i], error->message);
-                g_clear_error (&error);
-                continue;
-              }
-            if (remote_collection_id == NULL || g_strcmp0 (remote_collection_id, collection_id) != 0)
+          ostree_repo_get_remote_option (self->repo, remotes[i], "collection-id", NULL, &remote_collection_id, &error);
+          if (error != NULL)
+            {
+              g_warning ("Error getting collection ID for remote %s: %s", remotes[i], error->message);
+              g_clear_error (&error);
               continue;
+            }
+          if (remote_collection_id == NULL || g_strcmp0 (remote_collection_id, collection_id) != 0)
+            continue;
 
-            g_clear_pointer (&raw_path, g_free);
-            raw_path = g_build_filename (self->cached_repo_root,
-                                         "refs",
-                                         "remotes",
-                                         self->remote_name,
-                                         collection_ref,
-                                         NULL);
+          g_clear_pointer (&raw_path, g_free);
+          raw_path = g_build_filename (self->cached_repo_root,
+                                       "refs",
+                                       "remotes",
+                                       self->remote_name,
+                                       collection_ref,
+                                       NULL);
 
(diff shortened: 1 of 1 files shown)


libeos-update-server/repo.c
```diff
@@ -862,6 +862,92 @@ handle_refs_heads (EusRepo     *self,
...
+    {
+      g_debug ("Invalid /refs/mirrors/ request: missing ref");
+      soup_message_set_status (msg, SOUP_STATUS_BAD_REQUEST);
+      return;
+    }
+
+  remotes = ostree_repo_remote_list (self->repo, NULL);
+  if (remotes != NULL)
+    {
+      for (i = 0; i < g_strv_length (remotes); ++i)
```
smspillaz: `g_strv_length` here is O(N) - it would make more sense to cache this value above as opposed to recomputing it every time.

libeos-update-server/repo.c
```diff
@@ -862,6 +862,92 @@ handle_refs_heads (EusRepo     *self,
...
+              {
+                g_warning ("Error getting collection ID for remote %s: %s", remotes[i], error->message);
+                g_clear_error (&error);
+                continue;
+              }
+            if (remote_collection_id == NULL || g_strcmp0 (remote_collection_id, collection_id) != 0)
+              continue;
+
+            g_clear_pointer (&raw_path, g_free);
+            raw_path = g_build_filename (self->cached_repo_root,
```
smspillaz: Instead of having to clear and overload `raw_path` here, would it make more sense to assign this to a different name at the scope of this block with a `g_autofree` annotation?

mwleeds: That would work too but this way there's symmetry with how `handle_refs_heads` works.

libeos-update-server/repo.c
mwleeds: good catch

libeos-update-server/repo.c
```diff
@@ -862,6 +862,93 @@ handle_refs_heads (EusRepo     *self,
...
+  collection_id = g_strdup (requested_path + prefix_len);
+  *strstr (collection_id, "/") = '\0';
+  if (!ostree_validate_collection_id (collection_id, &error))
+    {
+      g_debug ("Invalid /refs/mirrors/ request: %s", error->message);
+      soup_message_set_status (msg, SOUP_STATUS_BAD_REQUEST);
+      return;
+    }
+
+  collection_ref = requested_path + prefix_len + strlen(collection_id) + 1; /* e.g eos2/i386 */
```
pwithnall: Nitpick: Missing space before `(`.

libeos-update-server/repo.c
```diff
@@ -862,6 +862,93 @@ handle_refs_heads (EusRepo     *self,
...
+  if (collection_ref == NULL || *collection_ref == '\0')
```
pwithnall: It’s not possible for `collection_ref == NULL` to ever be true.

libeos-update-server/repo.c
```diff
@@ -862,6 +862,93 @@ handle_refs_heads (EusRepo     *self,
...
+    {
+      g_debug ("Invalid /refs/mirrors/ request: missing ref");
+      soup_message_set_status (msg, SOUP_STATUS_BAD_REQUEST);
+      return;
+    }
+
+  remotes = ostree_repo_remote_list (self->repo, NULL);
+  if (remotes != NULL)
+    {
+      guint remotes_len = g_strv_length (remotes);
```
pwithnall: You can get `remotes_len` directly out of `ostree_repo_remote_list()` as its second argument.

libeos-update-server/repo.c
```diff
@@ -862,6 +862,93 @@ handle_refs_heads (EusRepo     *self,
...
+
+          served = FALSE;
+          if (!serve_file_if_exists (msg, self->cached_repo_root, raw_path, self->cancellable, &served) || served)
+            return;
+
+          g_debug ("Failed to find file ‘%s’, trying next remote", raw_path);
+        }
+    }
+
+  g_warning ("Requested ref ‘%s’ not found in any remote", requested_path);
```
pwithnall: Missing `soup_message_set_status (msg, SOUP_STATUS_NOT_FOUND)`?

mwleeds: good catch

mwleeds:
Thanks for the reviews; I addressed all the comments

PR #126, opened 2017-11-20
eos-updater-avahi: Drop support for old-style DNS-SD record format

pwithnall:
No client code exists in the wild any more which will read it, so drop
it. No systems were ever deployed which used the old format, and any
modern versions of eos-updater read the new format only.

The only remnant left of the old format is some code in
eos-updater-avahi.c which deletes the old
/etc/avahi/services/eos-updater.service file if it still exists, to
stop Avahi from advertising the old records further.

https://phabricator.endlessm.com/T18774


pwithnall:
(The first commit here is from #124, which should be reviewed first. It’s needed here in order to avoid merge conflicts.)

PR #133, opened 2017-12-15
Fixes for leaks spotted while reviewing T16682

pwithnall:
A few leaks spotted while reviewing the next commit in https://phabricator.endlessm.com/T16682.


pwithnall:
Fixed a couple more instances (thanks for reminding me). Ready to merge.

PR #135, opened 2017-12-18
installer: Add eos-updater-flatpak-installer

pwithnall:
(**This doesn’t really need reviewing; it’s the results of my review of one of the commits on #129, which I’m pushing here to get Jenkins to test run the build.**)

This binary uses the functions in libeos-updater-flatpak-installer
to actually install flatpaks on reboot. It also has a "check" mode
which checks if the flatpak actions we claim to have applied are actually
fully-up-to-date on the system.

(Modified by Philip Withnall to improve formatting on man pages and fix
a few minor bugs found during review.)


pwithnall:
Checks passed! Merging.

PR #139, opened 2017-12-20
T16682 more parser tests

pwithnall:
Add more tests for the JSON autoinstall file parser, and get our code coverage up to 100% on it.

### DIFF
--- a/libeos-updater-util/flatpak.c
+++ b/libeos-updater-util/flatpak.c
@@ -873,6 +873,10 @@ read_flatpak_ref_actions_from_node (JsonNode      *node,
       if (is_skipped)
         {
           g_autofree gchar *node_str = json_node_to_string (element_node);
+          g_debug ("%s while parsing %s. Skipping this action and it "
+                   "will not be reapplied later. System may be in an "
+                   "inconsistent state from this point forward.",
+                   local_error->message, filename);
           g_ptr_array_add (skipped_action_entries, g_steal_pointer (&node_str));
           continue;
         }
--- a/libeos-updater-util/flatpak.c
+++ b/libeos-updater-util/flatpak.c
@@ -130,7 +130,7 @@ flatpak_remote_ref_action_type_parse (const gchar                    *action,
       g_set_error (error,
                    EOS_UPDATER_ERROR,
                    EOS_UPDATER_ERROR_UNKNOWN_ENTRY_IN_AUTOINSTALL_SPEC,
-                   "Unknown action type %s specified in autoinstall spec",
+                   "Unknown action type ‘%s’ specified in autoinstall spec",
                    action);
       return FALSE;
     }
--- a/libeos-updater-util/flatpak.c
+++ b/libeos-updater-util/flatpak.c
@@ -421,6 +421,9 @@ flatpak_remote_ref_action_from_json_node (const gchar  *source,
           return NULL;
         }
 
+      /* Not currently possible to hit these lines of code, given that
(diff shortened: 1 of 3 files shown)


libeos-updater-util/flatpak.c
```diff
@@ -873,10 +891,8 @@ read_flatpak_ref_actions_from_node (JsonNode      *node,
                                     EOS_UPDATER_ERROR,
                                     EOS_UPDATER_ERROR_UNKNOWN_ENTRY_IN_AUTOINSTALL_SPEC))
             {
-              g_warning ("%s while parsing %s. Skipping this action and it "
-                         "will not be reapplied later. System may be in an "
-                         "inconsistent state from this point forward.",
-                         local_error->message, filename);
+              g_autofree gchar *node_str = json_node_to_string (element_node);
+              g_ptr_array_add (skipped_action_entries, g_steal_pointer (&node_str));
```
smspillaz: You can also just pass the return value of `json_node_to_string` directly to this function.

libeos-updater-util/flatpak.c
```diff
@@ -854,6 +865,13 @@ read_flatpak_ref_actions_from_node (JsonNode      *node,
           return NULL;
         }
 
+      if (is_skipped)
+        {
+          g_autofree gchar *node_str = json_node_to_string (element_node);
+          g_ptr_array_add (skipped_action_entries, g_steal_pointer (&node_str));
```
smspillaz: You can also just pass the return value of `json_node_to_string` directly to this function.

libeos-updater-util/flatpak.c
```diff
@@ -713,11 +714,13 @@ action_node_should_be_filtered_out (JsonNode  *node,
   GList *iter = NULL;
 
   g_return_val_if_fail (is_filtered != NULL, FALSE);
+  g_return_val_if_fail (is_skipped != NULL, FALSE);
 
   /* No filters, so this action cannot be filtered out */
   if (filters_object_node == NULL)
     {
       *is_filtered = FALSE;
+      *is_skipped = FALSE;
```
smspillaz: Since `is_skipped` is an error case, it may be better to immediately set the outparam to `FALSE` upon calling the function and only set it to `TRUE` when the error actually happens (as opposed to having to set it in every return location).

smspillaz: Actually, maybe instead of having `is_skipped`, it may be better to just propagate the `EOS_UPDATER_ERROR_UNKNOWN_ENTRY_IN_AUTOINSTALL_SPEC` error (or better yet a new error type `EOS_UPDATER_ERROR_UNKNOWN_FILTER_IN_AUTOINSTALL_SPEC`) up to the caller and then let it figure out that it wants to skip the action. That would avoid having a "pseudo-error" case in this function.

smspillaz:
Just had a few thoughts on the `*is_skipped` stuff but other than that LGTM

libeos-updater-util/flatpak.c
pwithnall: Good suggestions.

pwithnall:
Since it was a fairly trivial refactoring to address your comments, I won’t wait for re-review (but will wait for Jenkins).

pwithnall:
Jenkins, retest this please

pwithnall:
Oh Jenkins. :-( Reported as https://phabricator.endlessm.com/T2254#523464. Merging regardless.

PR #147, opened 2018-01-14
util: Override architecture during testing

smspillaz:
This means explicitly passing the architecture to
FlatpakRef's constructor so that we can pass an overridden one.

### DIFF
--- a/libeos-updater-util/flatpak.c
+++ b/libeos-updater-util/flatpak.c
@@ -247,6 +247,13 @@ parse_flatpak_ref_from_entry (JsonObject      *entry,
   return TRUE;
 }
 
+static const gchar *
+eos_updater_get_system_architecture_string (void)
+{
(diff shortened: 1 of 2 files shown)


pwithnall:
The [build fails](https://ci.endlessm-sf.com/job/obs-eos-updater-master/216/console) with this branch, but it also fails without this branch. The failure has been introduced by OSTree now being [compiled with Avahi support](https://github.com/endlessm/ostree/commit/79f55b18d41f0320dc8fe64db2ce02aaa0e985b6) enabled, which causes anything using the `ostree` command to fail on the CI machines, since the Avahi daemon isn’t running on them.

We need a separate PR to pass `--finders=config` or `--finders=config,mount` (but not `--finders=config,mount,lan`) to `ostree`, as per [this commit which does it in ostree.git](https://github.com/endlessm/ostree/commit/bb93ada4de90dfb055a1cccf1ae433f58070f037). Note that we’d need to bump our OSTree dependency in `debian-master`. @smspillaz, does that make sense?

PR #149, opened 2018-01-15
T15923 Thread safety documentation and warning fixes

pwithnall:
https://phabricator.endlessm.com/T15923


pwithnall:
> Do we pass `EosUpdater` to worker threads? Maybe we shouldn't be doing that ...

Argh, so we do. I’ll follow up on that in the task.

PR #150, opened 2018-01-16
src: Improvements to thread safety for apply and fetch operations

pwithnall:
This does not make things fully thread safe. In particular, the poll
operation is still a complete mess, and the fetch operation still has
some non-thread-safe accesses to the EosUpdaterData struct. The apply
operation should be thread-safe now though.

All the operations (poll, fetch, apply) use a single worker thread each
to do the bulk of their work. These were previously passed pointers to
the EosUpdater and EosUpdaterData structs, which are also used in the
main thread. This was not thread safe.

Fix that by adding a per-operation struct (ApplyData or FetchData) into
which the data needed by the worker thread is copied before the main
thread forks. In the case of the fetch operation, this includes an
OstreeAsyncProgress object, which reports progress data back to the main
thread. This is thread safe, since its callback is scheduled in the main
thread’s default GMainContext.

https://phabricator.endlessm.com/T15923

### DIFF
--- a/src/eos-updater-fetch.c
+++ b/src/eos-updater-fetch.c
@@ -30,6 +30,28 @@
 
 #include <flatpak.h>
 
+/* Closure containing the data for the fetch worker thread. The
+ * worker thread must not access EosUpdater or EosUpdaterData directly,
+ * as they are not thread safe. */
+typedef struct
+{
+  gchar *update_id;  /* (owned) */
+  gchar *update_refspec;  /* (owned) */
+  EosUpdaterData *data;  /* (unowned) */
+  OstreeAsyncProgress *progress;  /* (owned) */
+} FetchData;
+
+static void
+fetch_data_free (FetchData *data)
+{
+  g_free (data->update_id);
+  g_free (data->update_refspec);
+  g_clear_object (&data->progress);
+  g_free (data);
+}
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (FetchData, fetch_data_free)
+
 static void
 content_fetch_finished (GObject *object,
                         GAsyncResult *res,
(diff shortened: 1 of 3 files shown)


src/eos-updater-fetch.c
```diff
@@ -744,9 +758,18 @@ handle_fetch (EosUpdater            *updater,
       return TRUE;
     }
 
+  fetch_data = g_new0 (FetchData, 1);
+  fetch_data->update_id = g_strdup (eos_updater_get_update_id (updater));
+  fetch_data->update_refspec = g_strdup (eos_updater_get_update_refspec (updater));
+  fetch_data->progress = ostree_async_progress_new_and_connect (update_progress, updater);
```
smspillaz: This is a very minor nit, but I've found that for closures like this, it usually makes sense to have a separate constructor func, eg `fetch_data_new` which takes all the parameters it needs to construct the closure and then takes ownership of them through `g_strdup`, `g_object_ref` etc there.

That way if you ever need to construct `FetchData` in more than one place, changing ownership semantics of its members implies that only a single function needs to be updated.

src/eos-updater-fetch.c
```diff
@@ -744,9 +758,18 @@ handle_fetch (EosUpdater            *updater,
...
+
+  /* FIXME: Passing the EosUpdaterData to the worker thread is not thread safe.
+   * See: https://phabricator.endlessm.com/T15923 */
+  fetch_data->data = data;
+
   eos_updater_clear_error (updater, EOS_UPDATER_STATE_FETCHING);
-  task = g_task_new (updater, NULL, content_fetch_finished, user_data);
-  g_task_set_task_data (task, user_data, NULL);
+  task = g_task_new (updater, NULL, content_fetch_finished, NULL);
+  g_task_set_task_data (task, g_steal_pointer (&fetch_data), (GDestroyNotify) fetch_data_free);
```
smspillaz: As an added benefit, if you have a `fetch_data_new` here, then there's no need for `g_steal_pointer` etc, just pass the result directly to `g_task_set_task_data`

src/eos-updater-fetch.c
pwithnall: I find it’s clearer to have the code inline in the first instance. If we end up needing to create a `FetchData` from a second call site, then it’s time to factor the construction out.

PR #152, opened 2018-01-20
T20812: Add testcase for systems that don't yet have collection-ids

smspillaz:
It's the test we had to have.

https://phabricator.endlessm.com/T20812

### DIFF
--- a/test-common/utils.c
+++ b/test-common/utils.c
@@ -2087,7 +2087,8 @@ gboolean
 eos_test_setup_flatpak_repo_with_preinstalled_apps (GFile        *updater_dir,
                                                     const gchar  *branch,
                                                     const gchar  *repo_name,
-                                                    const gchar  *collection_id,
+                                                    const gchar  *repo_collection_id,
+                                                    const gchar  *remote_config_collection_id,
                                                     const gchar **flatpak_names,
                                                     const gchar **preinstall_flatpak_names,
                                                     GError      **error)
@@ -2133,7 +2134,8 @@ eos_test_setup_flatpak_repo_with_preinstalled_apps (GFile        *updater_dir,
                                  "org.test.Runtime",
                                  runtime_formatted_ref_name,
                                  branch,
-                                 collection_id,
+                                 remote_config_collection_id,
+                                 repo_collection_id,
                                  error))
     return FALSE;
 
@@ -2188,11 +2190,11 @@ eos_test_setup_flatpak_repo_with_preinstalled_apps (GFile        *updater_dir,
    * so the remote collection ID needs to be set after the flatpak is
    * installed. FIXME: The tests should be using GPG in order to be
    * representative of a real system. */
-  if (collection_id != NULL)
+  if (remote_config_collection_id != NULL)
     {
       if (!set_flatpak_remote_collection_id (updater_dir,
                                              repo_name,
-                                             collection_id,
+                                             remote_config_collection_id,
                                              error))
         return FALSE;
     }
@@ -2204,7 +2206,8 @@ gboolean
 eos_test_setup_flatpak_repo (GFile        *updater_dir,
                              const gchar  *branch,
                              const gchar  *repo_name,
-                             const gchar  *collection_id,
+                             const gchar  *repo_collection_id,
+                             const gchar  *remote_config_collection_id,
                              const gchar **flatpak_names,
                              GError      **error)
 {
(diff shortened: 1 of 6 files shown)


test-common/utils.c
```diff
@@ -2169,11 +2169,14 @@ eos_test_setup_flatpak_repo_with_preinstalled_apps (GFile        *updater_dir,
    * so the remote collection ID needs to be set after the flatpak is
    * installed. FIXME: The tests should be using GPG in order to be
    * representative of a real system. */
-  if (!set_flatpak_remote_collection_id (updater_dir,
-                                         repo_name,
-                                         collection_id,
-                                         error))
-    return FALSE;
+  if (collection_id != NULL)
```
pwithnall: The `collection_id` is used in the `flatpak_populate_runtime()` call above, which claims to be setting up the repository as well as the remote config. Is it OK for it to be `NULL` there? Would it make more sense to have two function arguments: one for the collection ID as used in the repository, and one for it as configured in the remote config?

smspillaz: The `collection_id` gets passed from there to `ostree_init` - that can handle a `NULL` `collection_id` just fine, it just won't pass it to ostree when it sets up the repo:

```
gboolean
ostree_init (GFile *repo,
             RepoMode mode,
             const gchar *collection_id,
             CmdResult *cmd,
             GError **error)
{
  CmdArg args[] =
    {
      { NULL, "init" },
      { "mode", repo_mode_to_string (mode) },
      { (collection_id != NULL) ? "collection-id" : NULL, collection_id },
      { NULL, NULL }
    };

  return spawn_ostree_in_repo_args (repo,
                                    args,
                                    cmd,
                                    error);
```

I suppose it could be useful to set up a `collection_id` on the repo and not on the remote.

pwithnall: > I suppose it could be useful to set up a `collection_id` on the repo and not on the remote.

I’d say it’s necessary to test both, since we have both setups. Most of our repositories on endlessm.com have collection IDs; some don’t.

smspillaz: Okay. I've split up the configuration, though I have yet to add a test for the latter case.

tests/test-update-install-flatpaks.c
```diff
@@ -624,6 +624,153 @@ test_update_install_flatpaks_in_repo (EosUpdaterFixture *fixture,
...
+  g_autofree gchar *flatpak_user_installation = NULL;
+  g_autoptr(GFile) flatpak_user_installation_dir = NULL;
+  g_auto(GStrv) wanted_flatpaks = flatpaks_to_install_app_ids_strv (flatpaks_to_install,
+                                                                    G_N_ELEMENTS (flatpaks_to_install));
+  g_auto(GStrv) flatpaks_in_repo = NULL;
+  g_autoptr(GFile) updater_directory = NULL;
+  g_autofree gchar *updater_directory_str = NULL;
+  g_autoptr(GError) error = NULL;
+
+  g_test_bug ("T16682");
```
pwithnall: T20812

tests/test-update-install-flatpaks.c
```diff
@@ -624,6 +624,153 @@ test_update_install_flatpaks_in_repo (EosUpdaterFixture *fixture,
...
```
pwithnall: T20812

tests/test-update-install-flatpaks.c
```diff
@@ -624,6 +624,153 @@ test_update_install_flatpaks_in_repo (EosUpdaterFixture *fixture,
...
+  updater_directory = g_file_get_child (data->client->root, "updater");
+  updater_directory_str = g_file_get_path (updater_directory);
+  flatpak_user_installation = g_build_filename (updater_directory_str,
+                                                "flatpak-user",
+                                                NULL);
+  flatpak_user_installation_dir = g_file_new_for_path (flatpak_user_installation);
+  eos_test_setup_flatpak_repo (updater_directory,
+                               "test-repo",
+                               "com.endlessm.TestInstallFlatpaksCollection",
+                               NULL,
```
pwithnall: Would be good to add a `/* remote config */` comment here to clarify the `NULL`, like you’ve done in the following test.

tests/test-update-install-flatpaks.c
```diff
@@ -700,12 +847,113 @@ test_update_install_flatpaks_in_repo_error_using_remote_name (EosUpdaterFixture
...
+  g_auto(GStrv) flatpaks_in_repo = NULL;
+  g_autoptr(GFile) updater_directory = NULL;
+  g_autofree gchar *updater_directory_str = NULL;
+  g_autoptr(EosTestAutoupdater) autoupdater = NULL;
+  g_autoptr(GFile) autoupdater_root = NULL;
+  g_auto(CmdResult) reaped_updater = CMD_RESULT_CLEARED;
+  g_auto(CmdAsyncResult) updater_cmd = CMD_ASYNC_RESULT_CLEARED;
+  g_autoptr(GError) error = NULL;
+
+  g_test_bug ("T16682");
```
pwithnall: T20812

PR #163, opened 2018-01-29
T20958 Fix NULL remote names in refspecs

pwithnall:
Potential fix for https://phabricator.endlessm.com/T20958. Only compile tested. See the commit messages for details.


mwleeds:
LGTM. There's still one call site for `ostree_parse_refspec` that doesn't check for `remote == NULL` but I can submit a separate PR for that

PR #166, opened 2018-01-31
flatpak-installer: tidy up debug messages

ramcq:
Always printed "is installed" regardless of return value of function.


pwithnall:
I’m going to wait for Jenkins to approve this before merging it.

pwithnall:
> However, that first change should be something which should have been caught by compiler warnings. Could you quickly check to make sure we’re passing all the appropriate warning flags in here?

I’ve tracked this down; it’s [an upstream bug with the structured version of `g_message()`](https://bugzilla.gnome.org/show_bug.cgi?id=793074). In eos-updater, we are passing the correct flag to the compiler (`-Wformat=2`), and if we (for example) have a format problem with a raw `printf()` call, it is detected.

pwithnall:
I’ll fix this upstream and backport it.

PR #168, opened 2018-02-08
T20396 Add metered data scheduling support to eos-updater

pwithnall:
https://phabricator.endlessm.com/T20396


src/eos-updater-ctl
```diff
@@ -293,6 +297,14 @@ def command_server(server_command, quiet=False):
     return 0
 
 
+def nonnegative_int(value):
+    int_value = int(value)
+    if int_value < 0:
+        raise argparse.ArgumentTypeError(
+            '%s must be a non-negative integer' % value)
```
wjt: The error message here will read:

```
eos-updater-ctl: error: argument --timeout: -1 must be a non-negative integer
```

which I don't really think makes sense. I think `'%s is negative'` would be clearer. In the case where it's not an integer the error will be:

```
eos-updater-ctl: error: argument --timeout: invalid nonnegative_integer value: 'snake'
```

You could just `raise ValueError` in the `< 0` case and the default message would be fine:

```
eos-updater-ctl: error: argument --timeout: invalid nonnegative_integer value: '-1'
```

src/eos-updater-fetch.c
```diff
@@ -690,6 +701,181 @@ prepare_flatpaks_to_deploy (OstreeRepo    *repo,
...
+
+  /* Schedule the download. */
+  g_variant_dict_insert (&parameters_dict, "resumable", "b", FALSE);
+  parameters = g_variant_ref_sink (g_variant_dict_end (&parameters_dict));
+
+  mwsc_scheduler_schedule_async (scheduler, parameters, cancellable,
+                                 async_result_cb, &schedule_result);
+
+  while (schedule_result == NULL)
+    g_main_context_iteration (context, TRUE);
```
wjt: I guess the assumption in this loop, and the loop above, is that they will in practice return quickly (and if Mogwai is buggy and never returns, the bus will time them out after 25 seconds) so you do not need the user-provided timeout to apply here?

data/docs/eos-autoupdater.conf.5
```diff
@@ -64,9 +64,7 @@ that the wait is always exactly \fIIntervalDays=\fP long.
 .\"
 .IP "\fIUpdateOnMobile=\fP"
 .IX Item "UpdateOnMobile="
-Whether to check for and download updates when connected to the internet over a
-mobile internet connection, such as GPRS or 3G. This may be \fItrue\fP or
-\fIfalse\fP.
+Deprecated. The value of this setting is ignored.
```
wjt: Should we teach Mogwai to check this setting? What was the old default?

src/eos-updater-fetch.c
pwithnall: I had actually neglected to think about timeouts there, but your suggestion is good. The default D-Bus timeouts should work here (we never really expect these paths to fail at all, unless `mogwai-scheduled` fails to start entirely). `scheduling_timeout_seconds` can be kept as the timeout waiting for the signal, which is the more common place to wait.

data/docs/eos-autoupdater.conf.5
pwithnall: The old default was `false` (i.e., do not update on a metered connection). Mogwai currently hard-codes to not download on metered connections: there is currently no setting to alter it. However, the pending g-c-c changes ([task](https://phabricator.endlessm.com/T20818), [PR](https://github.com/endlessm/gnome-control-center/pull/116)) will allow the control centre to set a preference on the NM connection which [says whether to download when that connection is metered](https://phabricator.endlessm.com/T20818#540067).

I think what we’re missing is to get g-c-c and Mogwai to load the default value for that preference from the old configuration file. If a value is already set on the NM connection, then that should be given priority over anything from `eos-autoupdater.conf`.

Good spot.

pwithnall: Commented in https://phabricator.endlessm.com/T20818#542708.

pwithnall: And filed https://phabricator.endlessm.com/T21135 about the Mogwai changes.

pwithnall:
Fixups and some additional commits pushed, which should get the tests going (although I suspect Jenkins will fail due to the Mogwai packages not being built successfully by OBS yet).

wjt:
@eos-shell-build retest this please

src/eos-updater-fetch.c
```diff
@@ -864,6 +895,14 @@ check_scheduler (FetchData     *fetch_data,
                  EOS_UPDATER_ERROR, EOS_UPDATER_ERROR_METERED_CONNECTION,
                  "Error fetching update: %s",
                  "Timed out waiting for download to be scheduled");
+  else if (invalidated_error != NULL)
+    {
+      g_autofree gchar *message = g_strdup_printf ("Download scheduler disappeared unexpectedly: %s",
+                                                   invalidated_error->message);
+      g_set_error (error,
+                   EOS_UPDATER_ERROR, EOS_UPDATER_ERROR_METERED_CONNECTION,
+                   "Error fetching update: %s", message);
```
wjt: Why not build this error in a single call? (No particular need to change it.)

configure.ac
```diff
@@ -108,7 +108,7 @@ AX_PKG_CHECK_MODULES([EOS_UPDATER_UTIL_TESTS],[flatpak >= $FLATPAK_REQUIRED_VERS
 AX_PKG_CHECK_MODULES([EOS_UPDATER_FLATPAK_INSTALLER_TESTS],[flatpak >= $FLATPAK_REQUIRED_VERSION glib-2.0 >= $GLIB_REQUIRED_VERSION gio-2.0 gobject-2.0 json-glib-1.0 >= $JSON_GLIB_REQUIRED_VERSION ostree-1 >= $OSTREE_REQUIRED_VERSION])
 AX_PKG_CHECK_MODULES([EOS_UPDATER_FLATPAK_INSTALLER],[flatpak >= $FLATPAK_REQUIRED_VERSION glib-2.0 >= $GLIB_REQUIRED_VERSION gio-2.0 gobject-2.0 json-glib-1.0 >= $JSON_GLIB_REQUIRED_VERSION ostree-1 >= $OSTREE_REQUIRED_VERSION])
 
-EOS_UPDATER_MODULES="glib-2.0 >= $GLIB_REQUIRED_VERSION gio-2.0 gobject-2.0 ostree-1 >= $OSTREE_REQUIRED_VERSION flatpak >= $FLATPAK_REQUIRED_VERSION avahi-client >= $AVAHI_REQUIRED_VERSION avahi-glib >= $AVAHI_REQUIRED_VERSION"
+EOS_UPDATER_MODULES="glib-2.0 >= $GLIB_REQUIRED_VERSION gio-2.0 gobject-2.0 ostree-1 >= $OSTREE_REQUIRED_VERSION flatpak >= $FLATPAK_REQUIRED_VERSION avahi-client >= $AVAHI_REQUIRED_VERSION avahi-glib >= $AVAHI_REQUIRED_VERSION mogwai-schedule-client-0"
```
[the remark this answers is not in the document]
pwithnall: But the pkg-config file is `mogwai-schedule-client-0.pc`: https://github.com/endlessm/mogwai/blob/master/debian/libmogwai-schedule-client-0-dev.install#L2

src/eos-updater-fetch.c
pwithnall: In order to keep the `Error fetching update: %s` string separate, and therefore reusable (it’s used in other error messages in the file).

debian/control
```diff
@@ -24,6 +24,7 @@ Build-Depends:
  libglib2.0-dev (>= 2.54.0),
  libgsystem-dev,
  libjson-glib-dev (>= 1.2.6-1endless2),
+ libmogwai-schedule-client-0,
```
wjt: So I think this should be `-dev` since the `.pc` file is in `libmogwai-schedule-client-0-dev`.

pwithnall: Argh, of course.

pwithnall: Fixup pushed.

debian/control
```diff
@@ -24,6 +24,7 @@ Build-Depends:
  libglib2.0-dev (>= 2.54.0),
  libgsystem-dev,
  libjson-glib-dev (>= 1.2.6-1endless2),
+ libmogwai-schedule-client-0-dev,
```
wjt:
```
/usr/bin/ld: cannot find -lmogwai-schedule-client-0
collect2: error: ld returned 1 exit status
Makefile:753: recipe for target 'eos-updater' failed
```

Reviewing the [build log](https://ci.endlessm-sf.com/job/obs-eos-updater-pr/79/consoleFull) it looks like the -dev package may not depend on the shared library package? I think it is conventional to have this dependency.

pwithnall: That seems reasonable. https://github.com/endlessm/mogwai/pull/22 opened.

pwithnall: I suspect I was expecting the `${shlibs:Depends}` dependency for `eos-updater` itself to pull `libmogwai-schedule-client-0` in; but it will actually only do that when you install the `eos-updater` package — not at build time, right?

wjt: Yes, I think that's right. I was surprised to rediscover this recently.

wjt:
retest this please

pwithnall:
And now it’s testing again because I quoted the command. 😩 

I’ll merge based on your test result. Thanks for your persistent reviews. 😄

pwithnall:
Squashed and merged manually.

PR #170, opened 2018-02-14
data: remove non-existent @chown syscall group

wjt:
I spotted the following in the journal:

    Feb 14 19:33:05 endless systemd[1]:
    [/lib/systemd/system/eos-updater-flatpak-installer.service:42] Don't
    know system call group, ignoring: @chown

The `@` syntax is used to refer to groups of system calls. The list of
known groups is documented in `systemd.exec(5)` and does not include
`@chown`. The `chown` syscall is part of the `@privileged` syscall group,
which per a comment above is deliberately allowed.

The current behaviour is that this service is allowed to invoke `chown`.
I think that's fine given it is also deliberately allowed to invoke much
more serious syscalls.


smspillaz:
I believe that @pwithnall wanted to keep this line despite the warning

wjt:
Ah, `@chown` is introduced in systemd v235. Fine.

PR #173, opened 2018-02-26
eufi: pass branch to flatpak_installation_update()

wjt:
The documentation for the 'branch' parameter says:

> name of the branch of the app or runtime to update (default: master).

'master' is never the correct branch for any Flatpaks we want to install or update after an ostree update. We need to pass it here, just like we do to other functions. Without this, if a Flatpak is specified to be installed, and it's already installed, the operation will fail:

    libeos-updater-flatpak-installer-Message: Attempting to install eos-sdk:runtime/org.freedesktop.Platform.Icontheme.EndlessOS/x86_64/1.0
    libeos-updater-flatpak-installer-Message: eos-sdk:runtime/org.freedesktop.Platform.Icontheme.EndlessOS/x86_64/1.0 already installed, updating
    libeos-updater-flatpak-installer-Message: Failed to update eos-sdk:runtime/org.freedesktop.Platform.Icontheme.EndlessOS/x86_64/1.0
    eos-updater-flatpak-installer: Couldn’t apply some flatpak update actions for this boot: org.freedesktop.Platform.Icontheme.EndlessOS branch master is not installed

This is not caught by the test suite because it uses the "master" branch throughout. As a result, there's also a patch here that attempts to use the branch "stable" throughout the test suite. However, the tests are flaky for me locally (with or without this change). Let's see what Jenkins thinks.

https://phabricator.endlessm.com/T20809


pwithnall:
Jenkins likes this; please merge!

PR #174, opened 2018-03-01
eos-updater-ctl: right-align property names

wjt:
Before:

    ======= Properties =======
    State: Ready
     CurrentID: '4b4e0fb003bae61d18152376a597675063c2dd1e3561ffe4eec8cd637f533f62'
     DownloadSize: int64 0
     DownloadedBytes: int64 0
     ErrorCode: uint32 0
     ErrorMessage: ''
     ErrorName: ''
     FullDownloadSize: int64 0
     FullUnpackedSize: int64 0
     OriginalRefspec: ''
     UnpackedSize: int64 0
     UpdateID: ''
     UpdateLabel: ''
     UpdateMessage: ''
     UpdateRefspec: ''

After:

    ======= Properties =======
               State: Ready
           CurrentID: '4b4e0fb003bae61d18152376a597675063c2dd1e3561ffe4eec8cd637f533f62'
        DownloadSize: int64 0
     DownloadedBytes: int64 0
           ErrorCode: uint32 0
        ErrorMessage: ''
           ErrorName: ''
    FullDownloadSize: int64 0
    FullUnpackedSize: int64 0
     OriginalRefspec: ''
        UnpackedSize: int64 0
            UpdateID: ''
         UpdateLabel: ''
       UpdateMessage: ''
       UpdateRefspec: ''

I think this is easier to skim, and in particular makes it a bit easier to eyeball the DownloadedBytes / DownloadSize ratio.

PR #175, opened 2018-03-06
eos-updater-ctl: fix syntax

wjt:
I broke this in #174. Sorry.

Any preferred way to check this at build time? One way would be to add a test case which just runs s `python3 -m py_compile eos-updater-ctl`. Another way would be to build-depend on a linter and run that. A third way would be to rename to `eos-updater-ctl.in` and do:

    eos-updater-ctl: eos-updater-ctl.in
    	$(AM_V_GEN) python3 -m py_compile $< && cp -a $< $@


pwithnall:
(Pulling in a linter means we will forever have build failures due to irrelevant lint failures. Renaming to `.in` just seems more ugly than option 1.)

PR #176, opened 2018-03-06
Ensure eos-updater-ctl compiles, by running it

wjt:
This is implemented as a build step, not a test, to increase the likelihood of it being run. I looked at adding it as an installed test in tests/ but installed tests are not run at build time.

I also took the liberty of fixing the (very few) warnings raised by flake8, my conservative Python linter of choice. While it does have some (IMO reasonable) opinions about line length and indentation, it can also detect many cases of (for example) referencing variables before they are defined. I'd suggest making `python3-flake8` a build-dependency and running it against `eos-updater-ctl` rather than invoking `eos-updater-ctl --help`, but I have not since this has already been deemed unpopular!

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -154,4 +154,8 @@ EXTRA_DIST += \
 	$(eos_updater_resource_deps) \
 	$(NULL)
 
+check-local: eos-updater-ctl.flake8
+%.flake8: %
+	$(AM_V_GEN) python3 -m flake8 $<
+
 -include $(top_srcdir)/git.mk
--- a/src/eos-updater-ctl
+++ b/src/eos-updater-ctl
@@ -60,7 +60,7 @@ EOS_UPDATER_CONFS = [
 current_proxy = None
 
 
-def signal_emitted(proxy, signal, parameters):
+def signal_emitted(proxy, sender, signal, parameters):
     if signal != "StateChanged":
         return
     state = UPDATER_STATES[parameters[0]]
@@ -78,6 +78,10 @@ def dump_daemon_properties(proxy):
     print("")
 
 
+def g_properties_changed_cb(proxy, changed, invalidated):
+    dump_daemon_properties(proxy)
+
+
 def name_appeared_cb(connection, name, name_owner):
     global current_proxy
 
@@ -92,12 +96,8 @@ def name_appeared_cb(connection, name, name_owner):
                                    'com.endlessm.Updater',
                                    None)  # cancellable
 
-    proxy.connect('g-properties-changed',
-                  lambda proxy, changed, invalidated, user_data: \
-                      dump_daemon_properties(proxy), None)
-    proxy.connect('g-signal',
-                  lambda proxy, sender, signal, parameters, user_data: \
-                      signal_emitted(proxy, signal, parameters), None)
+    proxy.connect('g-properties-changed', g_properties_changed_cb)
+    proxy.connect('g-signal', signal_emitted, None)
 
     dump_daemon_properties(proxy)
     current_proxy = proxy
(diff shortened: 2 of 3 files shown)


src/eos-updater-ctl
```diff
@@ -92,12 +96,8 @@ def name_appeared_cb(connection, name, name_owner):
...
                                    None)  # cancellable
 
-    proxy.connect('g-properties-changed',
-                  lambda proxy, changed, invalidated, user_data: \
-                      dump_daemon_properties(proxy), None)
-    proxy.connect('g-signal',
-                  lambda proxy, sender, signal, parameters, user_data: \
-                      signal_emitted(proxy, signal, parameters), None)
+    proxy.connect('g-properties-changed', g_properties_changed_cb)
+    proxy.connect('g-signal', signal_emitted, None)
```
pwithnall: OK, but shouldn’t the function names be prefixed with `__` to make them private?

src/Makefile.am
```diff
@@ -154,4 +154,10 @@ EXTRA_DIST += \
 	$(eos_updater_resource_deps) \
 	$(NULL)
 
+# Ensure that eos-updater-ctl compiles and runs at least as far as --help
+CLEANFILES += eos-updater-ctl.help-stamp
+all-local: eos-updater-ctl.help-stamp
+eos-updater-ctl.help-stamp: eos-updater-ctl
+	$(AM_V_GEN) $< --help >/dev/null && touch $@
```
pwithnall: I was only set on using `eos-updater-ctl --help` as a test case if it was an installed-test, since that would test it in-place on a system, and hence catch issues with dependencies not being installed (or whatever). Calling `--help` as a compile-time step seems quite odd.

In descending order of preference, I like the idea of:
 * A Python test harness which is run by `make check` and is an installed-test, which (for the moment) smoketests by calling `eos-updater-ctl --help` as a subprocess — but in future could be extended to test `eos-updater-ctl`’s behaviour in different ways.
 * A simpler `make check` test akin to this one, which runs `python3-flake8` on `eos-updater-ctl` at compile time. Since you’ve helpfully fixed the `python3-flake8` warnings, adding it as a dependency and using it on future commits won’t add much of a maintenance burden.
 * This.

I should have communicated this preference list earlier, sorry.

src/eos-updater-ctl
wjt: This script can't be imported as a module because its name contains hyphens, so by definition all top-level functions are private…

src/Makefile.am
wjt: I thought I had implemented option 2 but in fact it is running flake8 during `make all`. I'll do another round to move it to the `check` stage.

src/eos-updater-ctl
pwithnall: TIL

wjt: Well, "can't" is a strong word.

```python
In [1]: import eos-updater-ctl
  File "<ipython-input-1-89bee1140f50>", line 1
    import eos-updater-ctl
              ^
SyntaxError: invalid syntax


In [2]: import imp

In [3]: eos_updater_ctl = imp.load_source('m', 'eos-updater-ctl')

In [4]: dir(eos_updater_ctl)
Out[4]: 
['EOS_UPDATER_CONFS',
 'EOS_UPDATE_SERVER_CONFS',
 'GLib',
 'Gio',
 'METHODS',
 'UPDATER_STATES',
 '__builtins__',
 '__doc__',
 '__file__',
 '__loader__',
 '__name__',
 '__package__',
 '__spec__',
 'argparse',
 'command_apply',
 'command_dbus',
 'command_fetch',
 'command_monitor',
 'command_poll',
 'command_server',
 'command_update',
 'current_proxy',
 'dump_daemon_properties',
 'g_properties_changed_cb',
 'main',
 'name_appeared_cb',
 'name_vanished_cb',
 'nonnegative_int',
 'os',
 'signal_emitted',
 'sys']
```

I guess I mean they are all conceptually private. I don't really like the `__` prefix at the best of times but it really seems like overkill in a standalone script.

PR #177, opened 2018-03-09
Minor fixes to eos-updater-ctl

pwithnall:
This fixes a warning which @GeorgesStavracas was seeing when using `eos-updater-ctl monitor`.


pwithnall:
CI passed. Merging.

PR #178, opened 2018-03-16
T21623

joaquimrocha:
https://phabricator.endlessm.com/T21623

### DIFF
--- a/src/eos-updater-data.h
+++ b/src/eos-updater-data.h
@@ -48,14 +48,21 @@ struct EosUpdaterData
    * This needs to be passed from poll() to fetch().
    * May be NULL if using the fallback code in poll(). */
   OstreeRepoFinderResult **results;
+
+  /* The object to pass to the tasks performed by the updater, in order to be
+   * able to cancel them. Upon cancellation (which is done by the Cancel()
+   * method), the object is renewed (unreffed + replaced by a new instance). */
+  GCancellable *cancellable;
 };
 
-#define EOS_UPDATER_DATA_CLEARED { NULL, NULL }
+#define EOS_UPDATER_DATA_CLEARED { NULL, NULL, NULL }
 
 void eos_updater_data_init (EosUpdaterData *data,
                             OstreeRepo *repo);
 
 void eos_updater_data_clear (EosUpdaterData *data);
+
+void eos_updater_data_reset_cancellable (EosUpdaterData *data);
 
 G_DEFINE_AUTO_CLEANUP_CLEAR_FUNC (EosUpdaterData, eos_updater_data_clear)
 
--- a/src/eos-updater-poll.c
+++ b/src/eos-updater-poll.c
@@ -587,8 +587,9 @@ handle_poll (EosUpdater            *updater,
 
   /* FIXME: Passing the #OstreeRepo to the worker thread here is not thread safe.
    * See: https://phabricator.endlessm.com/T15923 */
+  eos_updater_data_reset_cancellable (data);
   eos_updater_clear_error (updater, EOS_UPDATER_STATE_POLLING);
-  task = g_task_new (updater, NULL, metadata_fetch_finished, data);
+  task = g_task_new (updater, data->cancellable, metadata_fetch_finished, data);
   g_task_set_task_data (task, g_object_ref (data->repo), g_object_unref);
   g_task_run_in_thread (task, metadata_fetch);
 
@@ -735,8 +736,9 @@ handle_poll_volume (EosUpdater            *updater,
   /* FIXME: The #OstreeRepo instance here is not thread safe. */
   poll_volume_data = poll_volume_data_new (data->repo, path);
 
+  eos_updater_data_reset_cancellable (data);
   eos_updater_clear_error (updater, EOS_UPDATER_STATE_POLLING);
-  task = g_task_new (updater, NULL, metadata_fetch_finished, data);
+  task = g_task_new (updater, data->cancellable, metadata_fetch_finished, data);
   g_task_set_task_data (task, g_steal_pointer (&poll_volume_data),
                         (GDestroyNotify) poll_volume_data_free);
   g_task_run_in_thread (task, poll_volume);
--- a/src/eos-updater-poll.c
+++ b/src/eos-updater-poll.c
@@ -469,6 +469,12 @@ metadata_fetch_internal (OstreeRepo     *repo,
 
       if (local_error != NULL)
         {
+          if (g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+            {
+              g_propagate_error (error, g_steal_pointer (&local_error));
+              return FALSE;
+            }
+
           use_new_code = FALSE;
 
           g_warning ("Error polling for updates using libostree P2P code; falling back to old code: %s",
@@ -510,7 +516,8 @@ metadata_fetch_internal (OstreeRepo     *repo,
                                task_context,
                                cancellable,
                                fetchers,
-                               config.download_order);
+                               config.download_order,
+                               &local_error);
         }
       else
         {
--- a/src/eos-updater.c
+++ b/src/eos-updater.c
@@ -71,6 +71,43 @@ local_data_init (LocalData *local_data,
   local_data->loop = g_main_loop_ref (loop);
 }
 
+static gboolean
+handle_cancel (EosUpdater            *updater,
+               GDBusMethodInvocation *call,
+               gpointer               user_data)
+{
+  EosUpdaterData *data = user_data;
+  EosUpdaterState state = eos_updater_get_state (updater);
+
+  switch (state)
+    {
+      case EOS_UPDATER_STATE_POLLING:
+      case EOS_UPDATER_STATE_FETCHING:
+      case EOS_UPDATER_STATE_APPLYING_UPDATE:
+        break;
+      case EOS_UPDATER_STATE_NONE:
+      case EOS_UPDATER_STATE_READY:
+      case EOS_UPDATER_STATE_UPDATE_AVAILABLE:
+      case EOS_UPDATER_STATE_UPDATE_READY:
+      case EOS_UPDATER_STATE_ERROR:
+      case EOS_UPDATER_STATE_UPDATE_APPLIED:
+      default:
+        g_dbus_method_invocation_return_error (call,
+                                               EOS_UPDATER_ERROR,
+                                               EOS_UPDATER_ERROR_WRONG_STATE,
+                                               "Can't call Cancel() while in "
+                                               "state %s (nothing to be cancelled)",
+                                               eos_updater_state_to_string (state));
+        return TRUE;
+    }
+
+  g_cancellable_cancel (data->cancellable);
+  eos_updater_data_reset_cancellable (data);
+
+  eos_updater_complete_cancel (updater, call);
+  return TRUE;
+}
+
 static void
 on_bus_acquired (GDBusConnection *connection,
                  const gchar     *name,
@@ -83,6 +120,11 @@ on_bus_acquired (GDBusConnection *connection,
   g_autofree gchar *sum = NULL;
 
   g_message ("Acquired a message bus connection");
+
+  /* Associate GIO's cancellation error with the EosUpdater domain's, since this
+   * is an error that can happen commonly */
+  g_dbus_error_register_error (G_IO_ERROR, G_IO_ERROR_CANCELLED,
+                               "com.endlessm.Updater.Error.Cancelled");
 
   /* Create a new org.freedesktop.DBus.ObjectManager rooted at /com/endlessm */
   local_data->manager = g_dbus_object_manager_server_new ("/com/endlessm");
@@ -129,6 +171,7 @@ on_bus_acquired (GDBusConnection *connection,
       g_signal_connect (updater, "handle-poll-volume",
                         G_CALLBACK (handle_on_live_boot), local_data->data);
       g_signal_connect (updater, "handle-apply", G_CALLBACK (handle_on_live_boot), local_data->data);
+      g_signal_connect (updater, "handle-cancel",  G_CALLBACK (handle_on_live_boot), local_data->data);
 
       eos_updater_set_error (updater, error);
       g_clear_error (&error);
@@ -142,6 +185,7 @@ on_bus_acquired (GDBusConnection *connection,
       g_signal_connect (updater, "handle-poll-volume",
                         G_CALLBACK (handle_poll_volume), local_data->data);
       g_signal_connect (updater, "handle-apply", G_CALLBACK (handle_apply), local_data->data);
+      g_signal_connect (updater, "handle-cancel",  G_CALLBACK (handle_cancel), local_data->data);
     }
 
   /* Export the object (@manager takes its own reference to @object) */
--- a/test-common/utils.c
+++ b/test-common/utils.c
@@ -109,6 +109,27 @@ eos_updater_fixture_teardown (EosUpdaterFixture *fixture,
                               gconstpointer user_data)
 {
   g_autoptr (GError) error = NULL;
+  g_auto(CmdResult) cmd = CMD_RESULT_CLEARED;
+  g_autofree gchar *gpg_home_path = g_file_get_path (fixture->gpg_home);
+  const gchar *argv[] =
+    {
+      "gpg-connect-agent",
+      "--homedir",
+      gpg_home_path,
+      "killagent",
+      "/bye",
+      NULL
+    };
+
+  /* kill the gpg-agent in order because if too many get spawned, it will
+   * result in connections getting refused... */
+  if (!test_spawn ((const gchar * const *) argv, NULL, &cmd, &error) ||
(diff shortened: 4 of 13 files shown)


src/eos-updater-data.h
```diff
@@ -48,15 +48,22 @@ struct EosUpdaterData
    * This needs to be passed from poll() to fetch().
    * May be NULL if using the fallback code in poll(). */
   OstreeRepoFinderResult **results;
+
+  /* The object to pass to the tasks performed by the updater, in order to be
+   * able to cancel them. Upon cancellation (which is donde by the Cancel()
```
pwithnall: Nitpick: s/donde/done/

src/eos-updater.c
```diff
@@ -144,6 +163,8 @@ on_bus_acquired (GDBusConnection *connection,
       g_signal_connect (updater, "handle-apply", G_CALLBACK (handle_apply), local_data->data);
     }
 
+  g_signal_connect (updater, "handle-cancel",  G_CALLBACK (handle_cancel), local_data->data);
```
pwithnall: I think this should be conditional like the other method callbacks, so that on a live boot it will return `EOS_UPDATER_ERROR_LIVE_BOOT` rather than succeeding (and doing nothing).

src/eos-updater-poll.c
```diff
@@ -469,6 +469,9 @@ metadata_fetch_internal (OstreeRepo     *repo,
 
       if (local_error != NULL)
         {
+          if (g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+            goto out;
```
pwithnall: Nitpick: Rather than an `out` label here, I think I’d prefer to put the `g_propagate_error(); return FALSE` inline here. I prefer to avoid labels if possible, to make the control flow simpler.

src/eos-updater.c
joaquimrocha: I did it like this because nothing special will happen (it'll just cancel the GCancellable...). But I agree that for consistency it's better to have this function as the other methods. Will change.

pwithnall: Yeah, I can see why you did it; but I think consistency is more important here. Thanks.

pwithnall:
Looks like Jenkins caught a problem with the build system changes:

```
Making all in tests
make[2]: Entering directory '/src/tests'
Makefile:1157: ../src/.deps/test_update_direct-eos-updater-generated.Po: No such file or directory
make[2]: *** No rule to make target '../src/.deps/test_update_direct-eos-updater-generated.Po'.  Stop.
make[2]: Leaving directory '/src/tests'
```

tests/Makefile.am
```diff
@@ -115,6 +116,14 @@ test_update_refspec_checkpoint_LDFLAGS = $(test_ldflags)
 test_update_refspec_checkpoint_LDADD = $(test_ldadd)
 test_update_refspec_checkpoint_SOURCES = test-update-refspec-checkpoint.c
 
+test_update_direct_CPPFLAGS = $(test_cppflags)
+test_update_direct_CFLAGS = $(test_cflags)
+test_update_direct_LDFLAGS = $(test_ldflags)
+test_update_direct_LDADD = $(test_ldadd)
+test_update_direct_SOURCES = test-update-direct.c \
+				$(top_srcdir)/src/eos-updater-generated.c \
+				$(top_srcdir)/src/eos-updater-generated.h
```
pwithnall: Should this be s/`top_srcdir`/`top_builddir`?

pwithnall: Also, please break lists up like this:
```
test_update_direct_SOURCES = \
	test-update-direct.c \
	$(top_builddir)/src/eos-updater-generated.c \
	$(top_builddir)/src/eos-updater-generated.h \
	$(NULL)
```

That makes the indentation a bit more consistent. The trailing `$(NULL)` reduces diff noise if more files are added to the list in future.

tests/test-update-direct.c
```diff
@@ -0,0 +1,255 @@
...
+      case EOS_UPDATER_N_STATES: /* just supress the warning about covering all enum values... */
+      default:
+        /* let it go until the next state change occurs */
+        break;
+    }
+}
+
+static void
+test_cancel_update (EosUpdaterFixture *fixture,
+                    gconstpointer user_data)
```
pwithnall: For each unit test (so, just the one in this file), please add a comment directly above the test function which briefly explains what the test checks. (I realise other tests in eos-updater probably don’t do this, but it’s something I’m trying to introduce).

For example:
```
/* Test that calling Cancel() on each state in a complete update correctly causes a transition to the Error state. */
static void
test_cancel_update (…)
```

tests/test-update-direct.c
```diff
@@ -0,0 +1,255 @@
...
+
+  updater = eos_updater_proxy_new_for_bus_sync (G_BUS_TYPE_SESSION,
+                                                G_DBUS_PROXY_FLAGS_NONE,
+                                                "com.endlessm.Updater",
+                                                "/com/endlessm/Updater",
+                                                NULL,
+                                                &error);
+  g_assert_no_error (error);
+
+  g_signal_connect (updater, "notify::state",
```
pwithnall: This signal handler should be disconnected at the end of the test.

(I realise signal handlers are automatically disconnected when the source object is finalised, but I don’t want any new code to rely on that, since it’s fragile and depends on every other user of the object dropping their refs before you.)

tests/test-update-direct.c
```diff
@@ -0,0 +1,255 @@
...
+                    G_CALLBACK (updater_state_changed_cb),
+                    loop);
+
+  /* start the state changes */
+  updater_state_changed_cb (updater, NULL, loop);
+
+  g_main_loop_run (loop);
+
+  eos_test_client_has_commit (client,
```
pwithnall: This only tests that the client has the given commit, not that it’s the currently deployed commit. I think you should also add something along the lines of `eos_test_client_has_commit (…, 2, …); g_assert_no_error (…); g_assert_false (has_commit);` to check that the update didn’t succeed at all.

pwithnall: (I might’ve misunderstood what’s going on here, though, so correct me if I’m wrong.)

tests/test-update-direct.c
```diff
@@ -0,0 +1,255 @@
...
+
+  eos_test_client_run_updater (client,
+                               &main_source,
+                               1,
+                               NULL,
+                               NULL,
+                               &error);
+  g_assert_no_error (error);
+
+  updater = eos_updater_proxy_new_for_bus_sync (G_BUS_TYPE_SESSION,
```
pwithnall: Since this section of code differs from other tests, it could do with a comment.

pwithnall: Also, if you use `eos_updater_proxy_new()`, it takes a `GDBusConnection` as an argument, so you could explicitly pass in the `GDBusConnection` from `fixture->dbus` and hence make it more obvious that the proxy here is using the fake session bus.

tests/test-update-direct.c
```diff
@@ -0,0 +1,255 @@
...
+	EOS_UPDATER_STATE_UPDATE_AVAILABLE,
+	EOS_UPDATER_STATE_FETCHING,
+	EOS_UPDATER_STATE_UPDATE_READY,
+	EOS_UPDATER_STATE_APPLYING_UPDATE,
+	EOS_UPDATER_STATE_UPDATE_APPLIED,
+        EOS_UPDATER_N_STATES,
+} EosUpdaterState;
+
+gboolean cancelled_states[EOS_UPDATER_N_STATES] = { FALSE };
+const gchar *states_str[] = {"None",
```
pwithnall: You can use `eos_updater_state_to_string()` rather than needing to define your own `states_str`.

tests/test-update-direct.c
```diff
@@ -0,0 +1,255 @@
...
+	EOS_UPDATER_STATE_READY,
+	EOS_UPDATER_STATE_ERROR,
+	EOS_UPDATER_STATE_POLLING,
+	EOS_UPDATER_STATE_UPDATE_AVAILABLE,
+	EOS_UPDATER_STATE_FETCHING,
+	EOS_UPDATER_STATE_UPDATE_READY,
+	EOS_UPDATER_STATE_APPLYING_UPDATE,
+	EOS_UPDATER_STATE_UPDATE_APPLIED,
+        EOS_UPDATER_N_STATES,
+} EosUpdaterState;
```
pwithnall: This is already defined in `libeos-updater-util/types.h`.

tests/test-update-direct.c
```diff
@@ -0,0 +1,255 @@
...
+                             "Polling",
+                             "UpdateAvailable",
+                             "Fetching",
+                             "UpdateReady",
+                             "ApplyingUpdate",
+                             "UpdateApplied"
+};
+
+guint cancelled_error_count = 0;
+guint cancel_calls_count = 0;
```
pwithnall: Nitpick: I’d prefer to keep the test state all within locals in `test_cancel_update()` rather than as static globals. Means the test is a bit better encapsulated.

tests/test-update-direct.c
```diff
@@ -0,0 +1,255 @@
...
+      case EOS_UPDATER_STATE_UPDATE_READY:
+        eos_updater_call_apply (updater, NULL, NULL, NULL);
+        break;
+      case EOS_UPDATER_STATE_UPDATE_APPLIED:
+        g_main_loop_quit (loop);
+        break;
+      case EOS_UPDATER_STATE_POLLING:
+      case EOS_UPDATER_STATE_FETCHING:
+      case EOS_UPDATER_STATE_APPLYING_UPDATE:
+      case EOS_UPDATER_N_STATES: /* just supress the warning about covering all enum values... */
```
pwithnall: Nitpick: s/supress/suppress/

tests/test-update-direct.c
```diff
@@ -0,0 +1,255 @@
...
+      case EOS_UPDATER_STATE_READY:
+      case EOS_UPDATER_STATE_UPDATE_AVAILABLE:
+      case EOS_UPDATER_STATE_UPDATE_READY:
+      case EOS_UPDATER_STATE_UPDATE_APPLIED:
+        should_succeed = FALSE;
+        break;
+      case EOS_UPDATER_STATE_POLLING:
+      case EOS_UPDATER_STATE_FETCHING:
+      case EOS_UPDATER_STATE_APPLYING_UPDATE:
+      case EOS_UPDATER_N_STATES: /* just supress the warning about covering all enum values... */
```
pwithnall: Nitpick: s/supress/suppress/

tests/test-update-direct.c
```diff
@@ -0,0 +1,255 @@
...
+
+guint cancelled_error_count = 0;
+guint cancel_calls_count = 0;
+
+static gboolean
+cancel_update (EosUpdater *updater)
+{
+  gboolean should_succeed;
+  EosUpdaterState state = eos_updater_get_state (updater);
+  g_autoptr(GError) error = NULL;
```
pwithnall: Nitpick: s/`error`/`local_error`/ — I’m trying to keep the distinction between `GError **error` parameters and `GError *local_error` error pointers for use within the function. Keeping the naming consistent means refactoring (for example, to add a `GError **error` parameter in future) is easy.

tests/test-update-direct.c
```diff
@@ -0,0 +1,255 @@
...
+
+  if (should_succeed)
+    {
+      g_assert_no_error (error);
+      ++cancel_calls_count;
+      g_debug ("Cancelled state %s", states_str[state]);
+    }
+  else
+    {
+      g_assert (error != NULL);
```
pwithnall: Use `g_assert_nonnull (error)`. `g_assert()` will be compiled out if compiling with `G_DISABLE_ASSERT`.

tests/test-update-direct.c
```diff
@@ -0,0 +1,255 @@
...
+      g_debug ("Error cancelling %s: %s", states_str[state], error->message);
```
pwithnall: If the assertion fails, this `g_debug()` won’t be reached. How about making it a `g_test_message()` and moving it just above the `g_assert()`?

tests/test-update-direct.c
joaquimrocha: I took most of this messy code from `test-update-from-main` which I believe should be testing for the deployed commit.
Looking into `utils.c`, the `eos_test_client_has_commit` starts by calling:
```
  if (!get_deploy_ids (sysroot, osname, &ids, error))
    return FALSE;
```
Before looking for the commit filename for the commit number. So I'd say it's looking for the right deployed commit here (but I've spent more time than I wanted looking at this code and still only partially understand what's going on such is the mess... so please bear with me :slightly_smiling_face: ).

tests/test-update-direct.c
joaquimrocha: I'd have to instance a new connection (`g_dbus_connection_new_for_address_sync`) from the address in `fixture->dbus`. So is the extra code worth it?
I can add a comment saying that it'll take the connection set up by the test...

tests/test-update-direct.c
joaquimrocha: Completely missed that! Thanks!

tests/test-update-direct.c
pwithnall: Either would work, as long as you’re sure the right bus is actually being used!

tests/test-update-direct.c
joaquimrocha: I moved it under the assert on purpose. If the error is `NULL`, then the debug message shouldn't be shown as it'd not be true.

tests/test-update-direct.c
joaquimrocha: Alright...

tests/test-update-direct.c
pwithnall: Hmm, catch 22. Stick it in an `if` then, I guess.

joaquimrocha: Why is it catch 22? If the error is NULL, then the assert will get it, in which case it makes no sense to print the debug msg `"Error cancelling"` because there's been no error...

joaquimrocha:
New fixup commit with the requested changes.

tests/test-update-direct.c
pwithnall: Sorry, brain fart. Ignore my comments here.

tests/test-update-direct.c
pwithnall: This hasn’t been done, though I won’t block merging on it because it’s a nitpick.

tests/test-update-direct.c
```diff
@@ -195,50 +183,53 @@ test_cancel_update (EosUpdaterFixture *fixture,
...
                                &main_source,
                                1,
                                NULL,
                                NULL,
-                               &error);
-  g_assert_no_error (error);
+                               &local_error);
+  g_assert_no_error (local_error);
 
+  /* the proxy will use the DBus connection set up by the test */
```
pwithnall: Nitpick: s/DBus/D-Bus/

pwithnall:
Hang on, Jenkins failed again :(
```
# eos-updater-tests:ERROR:../../../tests/test-update-direct.c:75:cancel_update: assertion failed (error == NULL): GDBus.Error:com.endlessm.Updater.Error.WrongState: Can't call Cancel() while in state Error (nothing to be cancelled) (g-io-error-quark, 36)
ERROR: test-update-direct - too few tests run (expected 1, got 0)
ERROR: test-update-direct - exited with status 134 (terminated by signal 6?)
```

joaquimrocha:
Jenkins test this please.

joaquimrocha:
Jenkins test this please.

test-common/utils.c
```diff
@@ -109,6 +109,24 @@ eos_updater_fixture_teardown (EosUpdaterFixture *fixture,
...
+      "/bye",
+      NULL
+    };
+
+  /* kill the gpg-agent in order because if too many get spawned, it will
+   * result in connections getting refused... */
+  if (!test_spawn ((const gchar * const *) argv, NULL, &cmd, &error) ||
+      !cmd_result_ensure_ok (&cmd, &error))
+    g_warning ("Failed to kill gpg-agent %s: %s", gpg_home_path,
+               error->message);
```
pwithnall: Need to `g_clear_error (&error)` after this.

tests/Makefile.am
```diff
@@ -115,6 +116,30 @@ test_update_refspec_checkpoint_LDFLAGS = $(test_ldflags)
 test_update_refspec_checkpoint_LDADD = $(test_ldadd)
 test_update_refspec_checkpoint_SOURCES = test-update-refspec-checkpoint.c
 
+eos-updater-generated.c: eos-updater-generated.h
+	cp ../src/eos-updater-generated.c .
+
+eos-updater-generated.h:
+	cp ../src/eos-updater-generated.h .
```
joaquimrocha: @dbnicholson In the end just adding the `eos-updater-generated.*` files to the `nodist_test_update_direct_SOURCES` didn't solve the problem for distclean. So I added these changes instead. I am sure this is not the most elegant way to do it, so if you know the right way, please let me know.

pwithnall: I’ve just pushed a fixup which rearranges this to link against `libeos-updater-dbus.la` instead, which builds for me, though you should probably test it.

PR #179, opened 2018-03-20
Provide the update version property

joaquimrocha:
We now have the update version added to the commit metadata, so this
patch retrieves it and also adds the matching property to the dbus
interface.

https://phabricator.endlessm.com/T21610

### DIFF
--- a/src/eos-updater-poll-common.c
+++ b/src/eos-updater-poll-common.c
@@ -153,6 +153,7 @@ eos_update_info_finalize_impl (EosUpdateInfo *info)
   g_free (info->checksum);
   g_free (info->new_refspec);
   g_free (info->old_refspec);
+  g_free (info->version);
   g_strfreev (info->urls);
   g_clear_pointer (&info->results, ostree_repo_finder_result_freev);
 }
@@ -169,6 +170,7 @@ eos_update_info_new (const gchar *checksum,
                      GVariant *commit,
                      const gchar *new_refspec,
                      const gchar *old_refspec,
+                     const gchar *version,
                      const gchar * const *urls,
                      OstreeRepoFinderResult **results)
 {
@@ -184,6 +186,7 @@ eos_update_info_new (const gchar *checksum,
   info->commit = g_variant_ref (commit);
   info->new_refspec = g_strdup (new_refspec);
   info->old_refspec = g_strdup (old_refspec);
+  info->version = g_strdup (version);
   info->urls = g_strdupv ((gchar **) urls);
   info->results = g_steal_pointer (&results);
 
@@ -455,6 +458,7 @@ fetch_latest_commit (OstreeRepo *repo,
                      const gchar *url_override,
                      gchar **out_checksum,
                      gchar **out_new_refspec,
+                     gchar **out_version,
                      GError **error)
 {
   g_autofree gchar *checksum = NULL;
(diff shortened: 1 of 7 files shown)


ramcq:
This seems reasonable if setting NULL on a GDBus "s" property keeps it at "". Don't know enough about the internals to be sure of that, but I think that's what happens?

src/eos-updater-poll-common.c
```diff
@@ -771,10 +786,11 @@ eos_update_info_to_string (EosUpdateInfo *update)
 
   results = g_string_free (g_steal_pointer (&results_string), FALSE);
 
-  return g_strdup_printf ("%s, %s, %s, %s\n   %s%s",
+  return g_strdup_printf ("%s, %s, %s, %s, %s\n   %s%s",
                           update->checksum,
                           update->new_refspec,
                           update->old_refspec,
+                          update->version,
```
pwithnall: What if this is `NULL`?

joaquimrocha: Good point. I can change it to `(NULL)`...

src/eos-updater-poll-common.c
```diff
@@ -574,6 +583,12 @@ parse_latest_commit (OstreeRepo           *repo,
   else
     *out_redirect_followed = FALSE;
 
+  version = g_variant_lookup_value (metadata, "version", G_VARIANT_TYPE_STRING);
```
dbnicholson: Might be nice to use the `OSTREE_COMMIT_META_KEY_VERSION` macro even though `version` is a well known key.

src/eos-updater-poll-common.c
```diff
@@ -761,6 +761,7 @@ eos_update_info_to_string (EosUpdateInfo *update)
   g_autofree gchar *timestamp_str = NULL;
   g_autoptr(GString) results_string = NULL;
   g_autofree gchar *results = NULL;
+  const gchar *version = (update->version != NULL) ? update->version : "(NULL)";
```
pwithnall: `(no version information)` would probably be better than `(NULL)`.

tests/test-update-direct.c
```diff
@@ -246,6 +246,110 @@ test_cancel_update (EosUpdaterFixture *fixture,
...
+  GMainLoop *loop = (GMainLoop *) data;
+
+  if (state == EOS_UPDATER_STATE_POLLING)
+    return;
+
+  g_main_loop_quit (loop);
+}
+
+static void
+test_update_version (EosUpdaterFixture *fixture,
```
pwithnall: Could you add a comment above the test function explaining what the test does and what it’s trying to test, please?

tests/test-update-direct.c
```diff
@@ -246,6 +246,110 @@ test_cancel_update (EosUpdaterFixture *fixture,
...
+  g_autoptr(EosTestServer) server = NULL;
+  g_autofree gchar *keyid = get_keyid (fixture->gpg_home);
+  g_autoptr(GError) error = NULL;
+  g_autoptr(EosTestSubserver) subserver = NULL;
+  g_autoptr(GFile) client_root = NULL;
+  g_autoptr(EosTestClient) client = NULL;
+  g_autoptr(EosUpdater) updater = NULL;
+  g_autoptr(GMainLoop) loop = g_main_loop_new (NULL, FALSE);
+  DownloadSource main_source = DOWNLOAD_MAIN;
+  const gchar *version = user_data != NULL ? (const gchar *) user_data : "";
```
pwithnall: Nitpick: Brackets around the conditional makes the operator precedence clearer (just a code style issue to make it fit in with the rest of the code):
```
const gchar *version = (user_data != NULL) ? (const gchar *) user_data : "";
```

tests/test-update-direct.c
joaquimrocha: Right! Sorry I forgot this. I had a bunch of code already from last week, so I forgot to update that part.

PR #180, opened 2018-03-23
Update the copyright of the configure.ac

joaquimrocha:
This is just a trivial PR to check whether the tests are passing on master or not.


pwithnall:
The tests did pass. Let’s reduce unnecessary churn on the files and drop this branch though, since I think you’re done with testing Jenkins.

PR #181, opened 2018-03-26
systemd: Don't use SystemCallFilter with the flatpak installer

smspillaz:
Part of the design of the flatpak installer is that it was assumed
to never fail except in very exceptional cases, such disk space
running out. SystemCallFilter is a noble concept for reducing
attack surface area, but ultimately, we need the flatpak installer
to work correctly across updates, which means that changes in
the system calls that ostree or flatpak itself might use should
not break the flatpak installer due to security policy.

https://phabricator.endlessm.com/T21751

### DIFF
--- a/eos-updater-flatpak-installer/eos-updater-flatpak-installer.service.in
+++ b/eos-updater-flatpak-installer/eos-updater-flatpak-installer.service.in
@@ -42,7 +42,19 @@ SystemCallArchitectures=native
 SystemCallErrorNumber=EPERM
 # @network-io is required for logging to the journal to work
 # @privileged is required for certain ostree operations
-SystemCallFilter=~@chown @clock @cpu-emulation @debug @keyring @mount @module @obsolete @raw-io @resources
+#
+# Currently commented out. Changes in flatpak, ostree or systemd
+# itself can easily break the flatpak installer by preventing it
+# from doing what it needs to do. Every time this happens, it is
+# a pain to debug, since we need to do a process of elimination
+# to determine what system call was causing it to trip up. We
+# need the flatpak installer to work correctly in order to ensure
+# the integrity of updates and it should not fail because a change
+# in security policy prevented it from working. Once flatpak, ostree
+# and systemd stabilize a little more, we can reconsider enabling
+# this option again.
+#
+# SystemCallFilter=~@chown @clock @cpu-emulation @debug @keyring @mount @module @obsolete @raw-io @resources
 
 [Install]
 WantedBy=multi-user.target


pwithnall:
But `eos-updater-flatpak-installer` runs from the new deployed OSTree, as does systemd, OSTree and flatpak; so we should always be able to ensure the system call filter in this `.service` file is appropriate for the set of system calls made by the running `eos-updater-flatpak-installer` instance. I think we should just fix the `SystemCallFilter=` value to not blacklist whatever system call is being made.

The `SystemCallFilter=` value we currently have is a blacklist, so it implicitly allows new or unknown system calls, and only disallows the groups which are listed. Either some of our code has started using one of the blacklisted system calls, or systemd has added a system call that we’re using to one of these blacklisted groups. Or the system call group which was ignored with systemd 232 (`@chown`, [T20944](https://phabricator.endlessm.com/T20944)) is now being enforced by systemd 237 and is breaking things.

Here’s a rough list of what’s changed in the system call filter groups between systemd 232 and 237:
 * statx(2), fremovexattr, inotify_init, oldfstat, oldlstat, oldstat, utime added to `@file-system`
 * _llseek added to `@base-io`
 * wait*, usage counters, process/thread identity, rt_sigqueueinfo, rt_tgsigqueueinfo added to `@process`
 * nice, ioprio_set added to `@resources`
 * prlimit64, ugetrlimit added to `@default`
 * Add `@aio`, `@sync` and `@chown` groups

smspillaz:
> But eos-updater-flatpak-installer runs from the new deployed OSTree, as does systemd, OSTree and flatpak; so we should always be able to ensure the system call filter in this .service file is appropriate for the set of system calls made by the running eos-updater-flatpak-installer instance. I think we should just fix the SystemCallFilter= value to not blacklist whatever system call is being made.

I appreciate that this can be fixed by just updating the `SystemCallFilter=` value to something that systemd, flatpak and ostree all like. My point is more about risk management and tradeoffs. This feature was designed to ideally never fail. It doesn't require internet access and should require very minimal available disk space, if any, because the relevant ostree objects have already been pulled into the repo on reboot. Because of the fact that we're relying on it not to fail across updates, we're using it for features like replacing LibreOffice and the Discovery Feed with the flatpak version.

Consider the worst case scenario. Lets say we rebase ostree, flatpak and/or systemd. I don't think its feasible for us to say that we'll audit all of those three components for new system calls and/or system call filter groups and consider the impact on downstream dependencies. Some new system call or system call filter gets introduced and which will trip up the flatpak installer. As happened in this case, nobody notices for a little while that this is the case since it isn't often that the flatpak installer gets run. Meanwhile, we ship an update removing LibreOffice from the OSTree and shipping it as a flatpak. The feature is broken and we only find out either at QA time (holding up a release) or worse, we ship the release and a bunch of people end up without LibreOffice, then we find out that its broken.

`SystemCallFilter=` is certainly a great development that allows us to mitigate some the damage in case someone gets code execution on the flatpak installer. But consider that the flatpak installer is already pretty well sandboxed. We don't allow executing writable memory, it can't touch the network and it can't touch /home. Granted, it can touch the system, but the ostree is mounted read-only anyway so the worst that it can do is touch /var and /etc. Given that we already have this protection, my point is that going forward, I'd rather that the tradeoff go in favour of keeping this feature working across updates at the cost of slightly less security as opposed to potentially having it break every time we rebase systemd, flatpak or ostree.

pwithnall:
I appreciate your point, and you have a valid viewpoint on the tradeoff between ease of maintenance and security. However, this is exactly the kind of thing we should be doing (automated) integration testing for. A single test run of upgrading the OSTree and `eos-updater-flatpak-installer` being run would have caught this, if I understand the situation correctly.

Rather than removing functionality, I would much prefer it if we set up automated tests for that functionality instead. Given how much of the system `eos-updater` interacts with, there are many things which can go wrong with it (or with `eos-updater-flatpak-installer`) — we’re just as likely to be hit by a change in libflatpak as we are to be hit by a change in systemd’s sandboxing. Automated integration testing on our OS images is a solution which would catch all such problems. Anything else is a stopgap.

smspillaz:
> Rather than removing functionality, I would much prefer it if we set up automated tests for that functionality instead. Given how much of the system eos-updater interacts with, there are many things which can go wrong with it (or with eos-updater-flatpak-installer) 

That would certainly be nice, but I am mindful of the fact that this is going to be quite a lot of work to scope out. In order to test this properly, you need to run the relevant code with the given systemd sandboxing options.

Two options here I guess. I figure out what system call is causing the problem and fix it now and we block further systemd/flatpak/ostree rebases on getting these integration tests done or we remove the `SystemCallFilter=` and block re-adding it on getting these integration tests done.

The other thing to keep in mind is that *even if* we have integration tests to catch this sort of thing, there is still the point of the fact that it will potentially break every time systemd and/or flatpak/ostree are updated and we take on the maintenance burden going through the trial-and-error process to ix it every time. This is not as big a problem as shipping a broken update to users

For now I'll figure out what system call is causing the problem and then we'll make a decision as to what to do from there.

pwithnall:
I’d start your debugging by looking at the `@chown` group.

cosimoc:
@smspillaz @pwithnall FWIW I am leaning towards a somewhat pragmatic approach here:
- I want to see this fixed ASAP in the short term. I think the best way to achieve that is to identify the syscall group that's causing the failure and just commenting that one out. Only if that proves to be much more complicated than initially thought, or inconclusive, I'd remove the whole `SystemCallFilter=` directive.
- In the medium-term, I agree that it would be worth investing in automated integration testing on our OS images. That is something that we should not embark on without the buy in of the tools and QA teams though. @pwithnall Since you seem to have the best idea of what that would consist of, could you make sure that stories in Phabricator are open about it? Please copy me, Dan, Brent and Steve as well.

smspillaz:
The good news is that it was `@chown` that was causing the problem: https://github.com/endlessm/eos-updater/pull/182

I'll close this one for now. However, if this keeps breaking on systemd/flatpak/ostree updates (the likelihood of that now is probably low, since I don't think that flatpak or ostree will end up using system calls in those groups), I think it'd be prudent to re-evaluate whether the call filter is worth it. It'd be nice to see automated testing for sure, but my point still stands about there being non-negligible costs in time spent debugging every time this breaks.

pwithnall:
> In the medium-term, I agree that it would be worth investing in automated integration testing on our OS images. That is something that we should not embark on without the buy in of the tools and QA teams though. @pwithnall Since you seem to have the best idea of what that would consist of, could you make sure that stories in Phabricator are open about it? Please copy me, Dan, Brent and Steve as well.

High-level tasks (I’ve CCed you to these):
 * https://phabricator.endlessm.com/T17238
 * https://phabricator.endlessm.com/T10644
 * https://phabricator.endlessm.com/T18881

Tasks more specific to eos-updater:
 * https://phabricator.endlessm.com/T16034
 * https://phabricator.endlessm.com/T15340
 * https://phabricator.endlessm.com/T15981

High-level tasks which are tangential to integration testing:
 * https://phabricator.endlessm.com/T2265
 * https://phabricator.endlessm.com/T4563

PR #183, opened 2018-03-30
libeuu: Fix using EuuConfigFile with just the GResource

pwithnall:
If loading an EuuConfigFile with a GResource, but none of the listed
files existing on the file system, it would hit an assertion failure.
Fix the special case handling of the GResource so that this works.

Adapted from a patch in https://phabricator.endlessm.com/T21619.

### DIFF
--- a/libeos-updater-util/config.c
+++ b/libeos-updater-util/config.c
@@ -370,7 +370,7 @@ euu_config_file_get_file_for_key (EuuConfigFile  *self,
     }
 
   /* Not found? */
-  if (i >= self->n_paths)
+  if (i > self->n_paths)
     {
       key_file = NULL;
       path = NULL;
--- a/libeos-updater-util/tests/config.c
+++ b/libeos-updater-util/tests/config.c
@@ -240,6 +240,28 @@ test_config_file_nonexistent (Fixture       *fixture,
   g_assert_cmpuint (loaded_file, ==, 1);
 }
 
+/* Test that if none of the files exist, but the GResource does, we successfully
+ * use that.. */
+static void
+test_config_file_resource_only (Fixture       *fixture,
+                                gconstpointer  user_data G_GNUC_UNUSED)
+{
+  g_autoptr(GError) error = NULL;
+  g_autoptr(EuuConfigFile) config = NULL;
+  const gchar * const paths[] =
+    {
+      fixture->key_file_nonexistent_path,
+      NULL
+    };
+  guint loaded_file;
+
+  config = euu_config_file_new (paths, fixture->default_resource, fixture->default_path);
+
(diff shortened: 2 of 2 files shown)


libeos-updater-util/config.c
```diff
@@ -370,7 +370,7 @@ euu_config_file_get_file_for_key (EuuConfigFile  *self,
     }
 
   /* Not found? */
-  if (i >= self->n_paths)
+  if (i > self->n_paths)
```
smspillaz: This makes sense in light of line 360 above

joaquimrocha: Yes. I see what it does too. However, the loop in line 360 does look confusing and unsurprisingly leads to errors like what's being fixed here.
I will approve the changes too but I'd rather just use a loop where `i < n_paths` and abstract the loop's body logic in a different function in order to call it after the loop with `default_path` as an argument. I think this would make things clear, require no comments for understanding what's going on, and thus be less error prone.

Either way, it's @pwithnall 's choice.

pwithnall: I don’t think abstracting the loop body out would particularly help with understandability. I think the comment is enough, and I don’t want to spend any more time on this at the moment. Thanks for the reviews!

pwithnall:
I really should write a test for this though.

pwithnall:
Test added. Ready for re-review.

Once this is approved, I need to cherry-pick the test to the copy of this code in eos-paygd. (The code needs factoring out into a proper separate library at some point — that’s https://phabricator.endlessm.com/T21621.)

libeos-updater-util/tests/config.c
```diff
@@ -240,6 +240,28 @@ test_config_file_nonexistent (Fixture       *fixture,
...
+  const gchar * const paths[] =
+    {
+      fixture->key_file_nonexistent_path,
+      NULL
+    };
+  guint loaded_file;
+
+  config = euu_config_file_new (paths, fixture->default_resource, fixture->default_path);
+
+  loaded_file = euu_config_file_get_uint (config, "Test", "File", 0, G_MAXUINT, &error);
```
smspillaz: If `paths` is moved above `config` these can probably all be declared and initialized on the same line

pwithnall: Meh, not worth it.

pwithnall:
Thanks for the review. Merging, and I’ll cherry-pick the new test to [libgsystemservice](https://gitlab.gnome.org/pwithnall/libgsystemservice) which is the new upstream for this helper code. (Eventually I will [factor it out of eos-updater](https://phabricator.endlessm.com/T22280).)

pwithnall:
Cherry picked: https://gitlab.gnome.org/pwithnall/libgsystemservice/merge_requests/3

PR #188, opened 2018-04-09
Add autoinstall flatpaks to update USBs

mwleeds:
When an OS update has flatpak(s) set to be updated or installed along
with it (via eos-updater-flatpak-installer), the flatpaks are required
for the update for succeed. So this commit adds the flatpaks to a USB
being created with eos-updater-prepare-volume so that USB updates work
even when the update in question has flatpaks to be autoinstalled. This
is accomplished by adding a "print" mode to
eos-updater-flatpak-installer and parsing the output from that in
eos-updater-prepare-volume. This way the logic that parses the
configuration files can stay in one place. Since we can't make any
assumptions about the OS version on the client receiving the USB update,
every flatpak configured to be automatically updated or installed on
OS update must be included.

https://phabricator.endlessm.com/T21756

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -487,7 +487,7 @@ EXTRA_DIST += m4/introspection.m4
 if HAVE_INTROSPECTION
 # GObject Introspection for libeos-updater-util
 libeos-updater-util/EosUpdaterUtil-@EUU_API_VERSION@.gir: libeos-updater-util/libeos-updater-util-@EUU_API_VERSION@.la
-libeos_updater_util_EosUpdaterUtil_@EUU_API_VERSION@_gir_INCLUDES = GObject-2.0 Gio-2.0 Flatpak-1.0
+libeos_updater_util_EosUpdaterUtil_@EUU_API_VERSION@_gir_INCLUDES = GObject-2.0 Gio-2.0 Flatpak-1.0 OSTree-1.0
 libeos_updater_util_EosUpdaterUtil_@EUU_API_VERSION@_gir_CFLAGS = \
 	$(libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_CPPFLAGS) \
 	$(libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_CFLAGS)
--- a/Makefile.am
+++ b/Makefile.am
@@ -487,7 +487,7 @@ EXTRA_DIST += m4/introspection.m4
 if HAVE_INTROSPECTION
 # GObject Introspection for libeos-updater-util
 libeos-updater-util/EosUpdaterUtil-@EUU_API_VERSION@.gir: libeos-updater-util/libeos-updater-util-@EUU_API_VERSION@.la
-libeos_updater_util_EosUpdaterUtil_@EUU_API_VERSION@_gir_INCLUDES = GObject-2.0 Gio-2.0
+libeos_updater_util_EosUpdaterUtil_@EUU_API_VERSION@_gir_INCLUDES = GObject-2.0 Gio-2.0 Flatpak-1.0
 libeos_updater_util_EosUpdaterUtil_@EUU_API_VERSION@_gir_CFLAGS = \
 	$(libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_CPPFLAGS) \
 	$(libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_CFLAGS)
--- a/Makefile.am
+++ b/Makefile.am
@@ -508,8 +508,8 @@ libeos_updater_util_EosUpdaterUtil_@EUU_API_VERSION@_gir_SCANNERFLAGS = \
 INTROSPECTION_GIRS += libeos-updater-util/EosUpdaterUtil-@EUU_API_VERSION@.gir
 
 # GObject Introspection for libeos-updater-flatpak-installer
-libeos-updater-flatpak-installer/EosUpdaterFlatpakInstaller-@EUFI_API_VERSION@.gir: libeos-updater-flatpak-installer/libeos-updater-flatpak-installer-@EUFI_API_VERSION@.la
-libeos_updater_flatpak_installer_EosUpdaterFlatpakInstaller_@EUFI_API_VERSION@_gir_INCLUDES = GObject-2.0 Gio-2.0 Flatpak-1.0
+libeos-updater-flatpak-installer/EosUpdaterFlatpakInstaller-@EUFI_API_VERSION@.gir: libeos-updater-flatpak-installer/libeos-updater-flatpak-installer-@EUFI_API_VERSION@.la libeos-updater-util/EosUpdaterUtil-@EUU_API_VERSION@.gir
+libeos_updater_flatpak_installer_EosUpdaterFlatpakInstaller_@EUFI_API_VERSION@_gir_INCLUDES = GObject-2.0 Gio-2.0 Flatpak-1.0 EosUpdaterUtil-0
 libeos_updater_flatpak_installer_EosUpdaterFlatpakInstaller_@EUFI_API_VERSION@_gir_CFLAGS = \
 	$(STD_CFLAGS) \
 	$(EOS_UPDATER_FLATPAK_INSTALLER_CFLAGS) \
@@ -520,6 +520,7 @@ libeos_updater_flatpak_installer_EosUpdaterFlatpakInstaller_@EUFI_API_VERSION@_g
 libeos_updater_flatpak_installer_EosUpdaterFlatpakInstaller_@EUFI_API_VERSION@_gir_NAMESPACE = EosUpdaterFlatpakInstaller
 libeos_updater_flatpak_installer_EosUpdaterFlatpakInstaller_@EUFI_API_VERSION@_gir_SCANNERFLAGS = \
 	$(WARN_SCANNERFLAGS) \
+	--add-include-path=$(top_builddir)/libeos-updater-util \
 	--nsversion=@EUFI_API_VERSION@ \
 	--c-include="libeos-updater-flatpak-installer/installer.h" \
 	--identifier-prefix=Eufi \
@@ -527,6 +528,7 @@ libeos_updater_flatpak_installer_EosUpdaterFlatpakInstaller_@EUFI_API_VERSION@_g
 	-I$(top_srcdir) \
 	$(NULL)
 
+INTROSPECTION_COMPILER_ARGS += --includedir=$(top_builddir)/libeos-updater-util
 INTROSPECTION_GIRS += libeos-updater-flatpak-installer/EosUpdaterFlatpakInstaller-@EUFI_API_VERSION@.gir
 
 girdir = $(datadir)/gir-1.0
--- a/Makefile.am
+++ b/Makefile.am
@@ -492,7 +492,7 @@ libeos_updater_util_EosUpdaterUtil_@EUU_API_VERSION@_gir_CFLAGS = \
 	$(libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_CPPFLAGS) \
 	$(libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_CFLAGS)
 libeos_updater_util_EosUpdaterUtil_@EUU_API_VERSION@_gir_FILES = \
-	$(libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_SOURCES)
+	$(filter-out %-private.h,$(libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_SOURCES))
 libeos_updater_util_EosUpdaterUtil_@EUU_API_VERSION@_gir_LIBS = libeos-updater-util/libeos-updater-util-@EUU_API_VERSION@.la
 libeos_updater_util_EosUpdaterUtil_@EUU_API_VERSION@_gir_NAMESPACE = EosUpdaterUtil
 libeos_updater_util_EosUpdaterUtil_@EUU_API_VERSION@_gir_SCANNERFLAGS = \
--- a/eos-updater-flatpak-installer/main.c
+++ b/eos-updater-flatpak-installer/main.c
@@ -125,7 +125,6 @@ main (int    argc,
   g_autoptr(GOptionContext) context = NULL;
   g_autoptr(FlatpakInstallation) installation = NULL;
   const gchar *resolved_mode = NULL;
-  const gchar *pending_flatpak_deployments_state_path = euu_pending_flatpak_deployments_state_path ();
   EosUpdaterInstallerMode parsed_mode;
 
   g_autofree gchar *mode = NULL;
@@ -208,7 +207,6 @@ main (int    argc,
             return EXIT_OK;
 
           if (!eufi_check_ref_actions_applied (installation,
-                                               pending_flatpak_deployments_state_path,
                                                squashed_ref_actions_to_check,
                                                &error))
             return fail (EXIT_CHECK_FAILED,
--- a/eos-updater-prepare-volume/eos-updater-prepare-volume
+++ b/eos-updater-prepare-volume/eos-updater-prepare-volume
@@ -90,7 +90,8 @@ class VolumePreparer:
         """Get the collection–ref tuple for the booted OS, or None."""
         deployment = self.sysroot.get_booted_deployment()
         if not deployment:
-            if 'EOS_UPDATER_TEST_UPDATER_DEPLOYMENT_FALLBACK' not in os.environ:
+            if 'EOS_UPDATER_TEST_UPDATER_DEPLOYMENT_FALLBACK' not in \
+               os.environ:
                 return None
 
             deployments = self.sysroot.get_deployments()
@@ -200,7 +201,8 @@ class VolumePreparer:
                                     (collection_id, ref)
                                     for (collection_id, ref) in invalid_refs]))
             return self.__fail(self.EXIT_INVALID_ARGUMENTS,
-                               'Invalid flatpak collection–refs: %s' % refs_list)
+                               'Invalid flatpak collection–refs: %s' %
+                               refs_list)
 
         # FIXME: Do we also want to pull in related refs, like locales?
         # Currently, they can be listed explicitly on the command line.
@@ -246,7 +248,8 @@ def main():
                         help='do not print anything; check exit status '
                              'for success')
     parser.add_argument('flatpak_refs', metavar='COLLECTION-ID REF', nargs='*',
-                        help='collection IDs and refs of flatpaks to put on the USB drive')
+                        help='collection IDs and refs of flatpaks to put on '
+                             'the USB drive')
 
     args = parser.parse_args()
 
--- a/eos-updater-prepare-volume/eos-updater-prepare-volume
+++ b/eos-updater-prepare-volume/eos-updater-prepare-volume
@@ -27,7 +27,9 @@ import gi
 gi.require_version('Flatpak', '1.0')
 gi.require_version('GLib', '2.0')
 gi.require_version('OSTree', '1.0')
+gi.require_version('EosUpdaterUtil', '0')
 from gi.repository import Flatpak, GLib, OSTree  # noqa
+from gi.repository import EosUpdaterUtil  # noqa
 
 
 class VolumePreparer:
@@ -174,6 +176,27 @@ class VolumePreparer:
 
         return runtimes
 
+    def _get_autoinstall_flatpaks(self):
+        """
+        Read all the autoinstall lists (see eos-updater-flatpak-installer(8))
+        and return a set of the flatpaks listed as needing an install or
+        upgrade. Load the set irrespective of the state of the autoinstall
+        counter on this system, since whichever system the USB drive is used on
+        might have a different counter value.
+        """
+        autoinstalls = set()
+
+        applied_actions = EosUpdaterUtil.flatpak_ref_actions_from_paths(None)
+
+        for filename, actions in applied_actions.items():
+            for action in actions:
+                if action.type != \
+                   EosUpdaterUtil.FlatpakRemoteRefActionType.UNINSTALL:
+                    autoinstalls.add((action.ref.collection_id,
+                                      action.ref.ref.format_ref()))
+
+        return autoinstalls
+
     def prepare_volume(self):
         # We need to be root in order to read all the files in the OSTree repo
         # (unless we’re running the unit tests). */
@@ -192,7 +215,7 @@ class VolumePreparer:
         # Work out which collection–refs we want on the USB stick. Each one is
         # a collection ID followed by a ref name.
         refs_iter = iter(self.flatpak_refs)
-        flatpak_collection_refs = list(zip(refs_iter, refs_iter))
+        flatpak_collection_refs = set(zip(refs_iter, refs_iter))
         invalid_refs = [collection_ref
                         for collection_ref in flatpak_collection_refs
                         if not self._validate_flatpak_ref(collection_ref)]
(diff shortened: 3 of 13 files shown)


mwleeds:
I'm not super happy with the way this is written, especially since it would break if a prefix other than `/usr` is used, but I'm submitting the PR anyway as a starting point for discussion.

eos-updater-flatpak-installer/main.c
```diff
@@ -249,6 +249,37 @@ main (int    argc,
...
+          if (new_flatpak_ref_actions_to_apply == NULL)
+            return fail (EXIT_FAILED,
+                         "Could not get information on which flatpak ref actions to apply: %s",
+                         error->message);
+
+          formatted_flatpak_ref_actions_to_apply =
+            euu_format_all_flatpak_ref_actions ("All flatpak ref actions that are not yet applied",
+                                                new_flatpak_ref_actions_to_apply);
+          g_print ("%s\n", formatted_flatpak_ref_actions_to_apply);
+
```
smspillaz: Same as above, can you split this out into another commit?

smspillaz: It strikes me that this probably repeats a lot of the existing debug messages in the tool. I wonder if it would make more sense to add a `--dry-run`option.

eos-updater-prepare-volume/eos-updater-prepare-volume
```diff
@@ -173,6 +174,17 @@ class VolumePreparer:
...
+    def _add_autoinstall_flatpaks(self, flatpak_collection_refs):
+        all_actions = self.__run(['/usr/libexec/eos-updater-flatpak-installer', '--mode=print'])
+        for line in all_actions.split('\n'):
+            pattern = '^ *- (install|update) \(collection-id: (?P<cid>.*?)\|.*?\):(?P<ref>.*?)$'
+            match = re.search(pattern, line)
+            if match is not None:
+                ref_tuple = (match.group('cid'), match.group('ref'))
+                if self.debug:
+                    print('Adding a Flatpak autoinstall app: ' + str(ref_tuple))
+                flatpak_collection_refs.append(ref_tuple)
```
smspillaz: Oh, I see how this is being used now.

Instead of parsing the output of `--print` (which was adapted from the existing debug messages), perhaps it might be better to have `eos-updater-flatpak-installer` dump JSON and use that instead?

eos-updater-prepare-volume/eos-updater-prepare-volume
```diff
@@ -173,6 +174,17 @@ class VolumePreparer:
 
         return runtimes
 
+    def _add_autoinstall_flatpaks(self, flatpak_collection_refs):
+        all_actions = self.__run(['/usr/libexec/eos-updater-flatpak-installer', '--mode=print'])
```
smspillaz: `libexec` is probably not correct here - `$(libexecdir)` can change depending on whether or not a debian package is being built (for instance, all the libexec stuff in debian packages gets installed into /usr/lib/arch-triple

You probably want to either preprocess this file or have a `config.py.in` where you can substitute the correct `$(libexecdir)` from autotools.

libeos-updater-util/flatpak.c
```diff
@@ -1578,6 +1578,9 @@ format_remote_ref_action_type (EuuFlatpakRemoteRefActionType action_type)
   return enum_value->value_nick;
 }
 
+/* NOTE: The output produced by this function is parsed by
+ * eos-updater-prepare-volume, so be mindful of that if you make any changes!
+ */
```
smspillaz: It might be better to dump JSON in this case :)

eos-updater-prepare-volume/eos-updater-prepare-volume
smspillaz: I was also thinking that perhaps `--print` might not belong in `eos-updater-flatpak-installer`. The good news is that `eufi_determine_flatpak_ref_actions_to_check` and friends are already in a separate library, so you could create a new binary and link directly to that. Not sure if we want yet another binary though. @pwithnall any thoughts on that?

eos-updater-prepare-volume/eos-updater-prepare-volume
mwleeds: That's a good thought. I thought about preprocessing but I didn't think about doing it in a separate Python file. That will be neater.

libeos-updater-util/flatpak.c
mwleeds: On the one hand that's added complexity and one more thing to maintain, but I suppose it's worth it to have something machine readable and a bit more robust.

eos-updater-prepare-volume/eos-updater-prepare-volume
smspillaz: FWIW, I agree with @pwithnall , though I don't know if libeos-updater-flatpak-installer is introspected yet. Shouldn't be too bad to introspect it, though it does use GPtrArray quite heavily and not sure if the element types would need to be boxed.

Makefile.am
```diff
@@ -295,6 +295,47 @@ libeos_updater_flatpak_installer_libeos_updater_flatpak_installer_@EUFI_API_VERS
...
 libexec_PROGRAMS += eos-updater-flatpak-installer/eos-updater-flatpak-installer
 
+###############################################################################
+# Introspection
+-include $(INTROSPECTION_MAKEFILE)
+INTROSPECTION_GIRS =
+INTROSPECTION_SCANNER_ARGS =
+INTROSPECTION_COMPILER_ARGS =
+AM_DISTCHECK_CONFIGURE_FLAGS += --enable-introspection
+EXTRA_DIST += m4/introspection.m4
```
pwithnall: You need to add `m4/introspection.m4` to git.

Makefile.am
```diff
@@ -295,6 +295,47 @@ libeos_updater_flatpak_installer_libeos_updater_flatpak_installer_@EUFI_API_VERS
...
+
+if HAVE_INTROSPECTION
+libeos-updater-flatpak-installer/Libeufi-@EUFI_API_VERSION@.gir: libeos-updater-flatpak-installer/libeos-updater-flatpak-installer-@EUFI_API_VERSION@.la
```
pwithnall: By convention, `.gir` files don’t have a `Lib` prefix — so this should probably be `EosUpdaterFlatpakInstaller-@EUFI_API_VERSION@.gir`.

pwithnall: For introspection to work, the library needs to be linkable against, which means you need to change `noinst_LTLIBRARIES += libeos-updater-flatpak-installer/libeos-updater-flatpak-installer-@EUFI_API_VERSION@.la` above to `lib_LTLIBRARIES += libeos-updater-flatpak-installer/libeos-updater-flatpak-installer-@EUFI_API_VERSION@.la`.

Makefile.am
```diff
@@ -295,6 +295,47 @@ libeos_updater_flatpak_installer_libeos_updater_flatpak_installer_@EUFI_API_VERS
...
+libeos_updater_flatpak_installer_Libeufi_@EUFI_API_VERSION@_gir_INCLUDES = GObject-2.0 Gio-2.0
+libeos_updater_flatpak_installer_Libeufi_@EUFI_API_VERSION@_gir_CFLAGS = \
+	$(libeos_updater_flatpak_installer_libeos_updater_flatpak_installer_@EUFI_API_VERSION@_la_CFLAGS) \
```
pwithnall: I think this should be:
```
$(STD_CFLAGS) \
$(EOS_UPDATER_FLATPAK_INSTALLER_CFLAGS) \
$(NULL)
```
rather than using the full list of `CFLAGS` from the library, otherwise you’ll end up with warnings and code coverage enabled for the GIR scanner, which will probably break things. GIR generates a C file which it compiles and runs to extract the runtime `GType` information from a library. Since it’s generated code, compiling it with warnings enabled is generally a bad idea.

Makefile.am
```diff
@@ -295,6 +295,47 @@ libeos_updater_flatpak_installer_libeos_updater_flatpak_installer_@EUFI_API_VERS
...
+	$(NULL)
+libeos_updater_flatpak_installer_Libeufi_@EUFI_API_VERSION@_gir_FILES =
+	libeos-updater-flatpak-installer/determine-flatpak-actions.c \
+	libeos-updater-flatpak-installer/installer.h \
+	libeos-updater-flatpak-installer/perform-flatpak-actions.c \
```
pwithnall: This list could just be `$(libeos_updater_flatpak_installer_libeos_updater_flatpak_installer_@EUFI_API_VERSION@_la_SOURCES)`.

Makefile.am
```diff
@@ -295,6 +295,47 @@ libeos_updater_flatpak_installer_libeos_updater_flatpak_installer_@EUFI_API_VERS
...
+	$(NULL)
+libeos_updater_flatpak_installer_Libeufi_@EUFI_API_VERSION@_gir_LIBS = libeos-updater-flatpak-installer/libeos-updater-flatpak-installer-@EUFI_API_VERSION@.la
+libeos_updater_flatpak_installer_Libeufi_@EUFI_API_VERSION@_gir_NAMESPACE = Libeufi
```
pwithnall: `EosUpdaterFlatpakInstaller`

Makefile.am
```diff
@@ -295,6 +295,47 @@ libeos_updater_flatpak_installer_libeos_updater_flatpak_installer_@EUFI_API_VERS
...
+libeos_updater_flatpak_installer_Libeufi_@EUFI_API_VERSION@_gir_SCANNERFLAGS = \
+	--nsversion=@EUFI_API_VERSION@ \
+	--warn-all \
```
pwithnall: Drop `--warn-all` if you’re using `$(WARN_SCANNERFLAGS)`.

Makefile.am
```diff
@@ -295,6 +295,47 @@ libeos_updater_flatpak_installer_libeos_updater_flatpak_installer_@EUFI_API_VERS
...
+	--c-include="libeos-updater-flatpak-installer/installer.h" \
```
pwithnall: Also add `--identifier-prefix=Eufi --symbol-prefix=eufi`.

Makefile.am
pwithnall: > By convention, …

Ah, it seems you already did that in the follow up commit. :+1:

mwleeds:
This is still WIP, but I put a status update on the ticket.

mwleeds:
This is ready for re-review now.

mwleeds:
There is a critical warning at runtime:
`** (process:6018): CRITICAL **: Unsupported type array`

But I'm not sure it's anything to lose sleep over. The array is clearly usable by Python.

pwithnall:
> There is a critical warning at runtime:
> `** (process:6018): CRITICAL **: Unsupported type array`
> 
> But I'm not sure it's anything to lose sleep over. The array is clearly usable by Python.

It’s definitely something I’d want fixed before merging. Run under `gdb` with `G_DEBUG=fatal-warnings` in the environment, and you should get a backtrace from where the critical’s being emitted, which should give you some clues about what’s causing it. Poke me in Slack if you need a second opinion.

wrt the Jenkins failure:

```
Couldn't find include 'EosUpdaterUtil-0.gir' (search path: '['gir-1.0', '/usr/share/gir-1.0', '/usr/share/gir-1.0', '/usr/share/gir-1.0']')
libtool: link: ( cd "test-common/.libs" && rm -f "libeos-updater-test-common.la" && ln -s "../libeos-updater-test-common.la" "libeos-updater-test-common.la" )
/usr/share/gobject-introspection-1.0/Makefile.introspection:159: recipe for target 'libeos-updater-flatpak-installer/EosUpdaterFlatpakInstaller-0.gir' failed
```

That can be fixed by adding `--add-include-path=$(top_builddir)/libeos-updater-util` to the relevant `SCANNERFLAGS`.

eos-updater-prepare-volume/eos-updater-prepare-volume
```diff
@@ -173,6 +176,13 @@ class VolumePreparer:
 
         return runtimes
 
+    def _add_autoinstall_flatpaks(self, flatpak_collection_refs):
+        applied_actions = EosUpdaterFlatpakInstaller.determine_flatpak_ref_actions_to_check(None)
```
pwithnall: Not entirely sure this is the right function to call. `determine_flatpak_ref_actions_to_check()` will load the autoinstall lists, parse them, and then *filter out* any entries whose serial number is lower than the latest applied serial number for this machine. I think you probably want to take part of the implementation for `determine_flatpak_ref_actions_to_check()` and call that instead. i.e. Call `flatpak_ref_actions_and_progresses()` and not `euu_filter_for_existing_flatpak_ref_actions()`. Since `flatpak_ref_actions_and_progresses()` is not public, it’s probably best to just copy part of its implementation (the bit relating to `out_actions`) instead. i.e. Call `euu_flatpak_ref_actions_from_paths()`.

Also, the above paragraph aside, did you consider calling `euu_flatten_flatpak_ref_actions_table()` on the `applied_actions` and reject the idea on the basis that the computer the updates are being applied to might have a different set of autoinstall list overrides, and hence it would be better to copy apps onto the USB stick even if their autoinstall list entries were overridden (squashed) on this machine?

eos-updater-prepare-volume/eos-updater-prepare-volume
```diff
@@ -173,6 +176,13 @@ class VolumePreparer:
...
+        for filename in applied_actions:
+            for action in applied_actions[filename]:
```
pwithnall: I think you can omit a hash operation on the dict key here:
```
for filename, actions in applied_actions.items():
    for action in actions:
        …
```

eos-updater-prepare-volume/eos-updater-prepare-volume
```diff
@@ -202,6 +212,9 @@ class VolumePreparer:
             return self.__fail(self.EXIT_INVALID_ARGUMENTS,
                                'Invalid flatpak collection–refs: %s' % refs_list)
 
+        # Add the flatpaks that will be installed by eos-updater-flatpak-installer
+        self._add_autoinstall_flatpaks(flatpak_collection_refs)
```
pwithnall: Need to handle exceptions from this call, so that the error message can be formatted nicely and a suitable exit status (`self.EXIT_FAILED`) returned.

eos-updater-prepare-volume/eos-updater-prepare-volume
pwithnall: But this code will change once you’ve called `euu_flatten_flatpak_ref_actions_table()` on the `applied_actions` anyway.

libeos-updater-util/flatpak.c
```diff
@@ -121,6 +121,18 @@ euu_flatpak_remote_ref_action_unref (EuuFlatpakRemoteRefAction *action)
   g_slice_free (EuuFlatpakRemoteRefAction, action);
 }
 
+char *
```
pwithnall: Nitpick: The eos-updater coding style uses `gchar` rather than `char`.

pwithnall: Please add gtk-doc comments for these new functions, too, since this is now essentially a public library, so we should be working towards full documentation coverage.

libeos-updater-util/flatpak.c
```diff
@@ -121,6 +121,18 @@ euu_flatpak_remote_ref_action_unref (EuuFlatpakRemoteRefAction *action)
...
+euu_flatpak_remote_ref_action_get_collection_id (EuuFlatpakRemoteRefAction *action)
+{
+    return g_strdup (action->ref->collection_id);
+}
+
+char *
+euu_flatpak_remote_ref_action_get_ref (EuuFlatpakRemoteRefAction *action)
```
pwithnall: I’d call this `euu_flatpak_remote_ref_action_format_ref()`, since it’s returning the string formatted version of the ref, not the `FlatpakRef*` object.

Same comments from above about `gchar` and documentation apply here. Please make sure to mention that the formatted ref is generated by `flatpak_ref_format_ref()` and hence can be parsed by `flatpak_ref_parse()`.

pwithnall:
Also, there are quite a few remaining warnings from `g-ir-scanner` in [the log](https://ci.endlessm-sf.com/job/obs-eos-updater-pr/222/console). Can you please file a follow-up Phabricator task about fixing them, if you haven’t already? Thanks.

mwleeds:
> Also, there are quite a few remaining warnings from g-ir-scanner in the log. Can you please file a follow-up Phabricator task about fixing them, if you haven’t already? Thanks.

https://phabricator.endlessm.com/T22636

test-common/utils.c
```diff
@@ -2932,12 +2932,17 @@ eos_test_client_prepare_volume (EosTestClient *client,
                                                                                "eos-updater-prepare-volume",
                                                                                "eos-updater-prepare-volume",
                                                                                NULL);
+  g_autofree gchar *libeos_updater_util_path = g_test_build_filename (G_TEST_DIST,
```
pwithnall: Shouldn’t this be s/`G_TEST_DIST`/`G_TEST_BUILT`/, since the typelib is generated at build time, rather than shipped in git?

pwithnall:
I’m going to look into the [runtime warning](https://github.com/endlessm/eos-updater/pull/188#issuecomment-389413696) and will do some work on this branch to fix up the comments while I’m at it, so please don’t do any more work on this yourself for the moment!

libeos-updater-util/flatpak.c
pwithnall: Also, the indentation is wrong inside these functions.

pwithnall:
OK, I’ve got some fixups [here](https://github.com/mwleeds/eos-updater/pull/1) which could be merged into this PR. I’d then like @smspillaz to take a look, [specifically at this bit](https://github.com/endlessm/eos-updater/pull/188#discussion_r188616028) (if he’s around and able).

If this PR gets accepted, before it’s merged, there are a few things to do:
 * Fix the [typo in a commit message](https://github.com/endlessm/eos-updater/pull/188#pullrequestreview-120624815)
 * Squash together [these two commits](https://github.com/endlessm/eos-updater/pull/188#pullrequestreview-120636020)
 * Ensure “Remove unused function parameter” is ordered before “Fix some g-ir-scanner warnings”
 * Rebase and squash in all the fixups (which will be non-trivial; it should work if they’re squashed in one-by-one, but squashing them in all at once causes conflicts because git doesn’t do it in order)

---

I looked at the critical warning message. It’s easily reproducible by running `python3` and then entering:
```
from gi.repository import EosUpdaterFlatpakInstaller, EosUpdaterUtil
EosUpdaterFlatpakInstaller.determine_flatpak_ref_actions_to_check(['/opt/endless/source/eos-application-tools/data'])
```

It gives the following backtrace:
```
** (process:31455): CRITICAL **: Unsupported type array

Program received signal SIGTRAP, Trace/breakpoint trap.
_g_log_abort (breakpoint=1) at /opt/endless/source/glib/glib/gmessages.c:554
554	}
(gdb) bt
#0  _g_log_abort (breakpoint=1) at /opt/endless/source/glib/glib/gmessages.c:554
#1  0x00007fffeebe9b3a in g_logv (log_domain=0x0, log_level=G_LOG_LEVEL_CRITICAL, 
    format=0x7fffef37fec8 "Unsupported type %s", args=0x7fffffffcbb8)
    at /opt/endless/source/glib/glib/gmessages.c:1362
#2  0x00007fffeebe9c2b in g_log (log_domain=0x0, log_level=G_LOG_LEVEL_CRITICAL, 
    format=0x7fffef37fec8 "Unsupported type %s") at /opt/endless/source/glib/glib/gmessages.c:1403
#3  0x00007fffef37ae47 in _pygi_marshal_to_py_ghash ()
   from /usr/lib64/python3.6/site-packages/gi/_gi.cpython-36m-x86_64-linux-gnu.so
#4  0x00007fffef371e4e in pygi_invoke_c_callable ()
   from /usr/lib64/python3.6/site-packages/gi/_gi.cpython-36m-x86_64-linux-gnu.so
#5  0x00007fffef3739b8 in pygi_function_cache_invoke ()
   from /usr/lib64/python3.6/site-packages/gi/_gi.cpython-36m-x86_64-linux-gnu.so
#6  0x00007ffff79980f4 in _PyObject_FastCallDict () from /lib64/libpython3.6m.so.1.0
#7  0x00007ffff7a06d8d in call_function () from /lib64/libpython3.6m.so.1.0
#8  0x00007ffff7a30e9a in _PyEval_EvalFrameDefault () from /lib64/libpython3.6m.so.1.0
#9  0x00007ffff7a0fb38 in PyEval_EvalCodeEx () from /lib64/libpython3.6m.so.1.0
#10 0x00007ffff7a108bb in PyEval_EvalCode () from /lib64/libpython3.6m.so.1.0
#11 0x00007ffff7a9c692 in run_mod () from /lib64/libpython3.6m.so.1.0
#12 0x00007ffff7967101 in PyRun_InteractiveOneObjectEx () from /lib64/libpython3.6m.so.1.0
#13 0x00007ffff7969bf6 in PyRun_InteractiveLoopFlags () from /lib64/libpython3.6m.so.1.0
#14 0x00007ffff7a9efc5 in PyRun_AnyFileExFlags () from /lib64/libpython3.6m.so.1.0
#15 0x00007ffff796ac61 in Py_Main.cold.2848 () from /lib64/libpython3.6m.so.1.0
#16 0x0000555555554e22 in main ()
```

Looking at that, and at the pygobject code, it seems that pygobject just doesn’t like having a `GPtrArray` as the value type of a `GHashTable`. As Matthew says, it does function perfectly well. I can’t see why the implementation of `GI_TYPE_TAG_ARRAY` should differ from the implementation of `GI_TYPE_TAG_INTERFACE` (GObject instances); no refcounting or checks are done in the latter case, so we shouldn’t need to do any for arrays. I think this is a bug in pygobject, and it should just not be warning about arrays.

⇒ I’ve filed https://gitlab.gnome.org/GNOME/pygobject/merge_requests/73 about it

libeos-updater-flatpak-installer/perform-flatpak-actions.c
```diff
@@ -440,6 +456,20 @@ check_if_flatpak_is_installed (FlatpakInstallation        *installation,
   return TRUE;
 }
 
+/**
+ * eufi_check_ref_actions_applied:
+ * @installation: a #FlatpakInstallation
+ * @pending_flatpak_deployments_state_path: unused
```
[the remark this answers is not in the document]
smspillaz: It could be a bug. If that parameter was there for anything, it would be there to limit the checked actions to include the ones up to and including the serials specified in that path. Perpahps better to just remove it though

eos-updater-prepare-volume/eos-updater-prepare-volume
smspillaz: In light of what @pwithnall here says about the overrides being potentially different on the target, the best bet here is to read in all the actions that might be necessary on the ostree and squash them from there.

I'm not sure how much can be done to handle the overrides situation - the source cant know what ref actions will be needed on the target since /var and /etc are not really managed by ostree

libeos-updater-util/flatpak.c
```diff
@@ -1063,6 +1067,13 @@ euu_flatpak_ref_actions_from_data (const gchar   *data,
   return g_steal_pointer (&actions);
 }
 
+/**
+ * euu_flatpak_remote_ref_actions_file_new:
+ * @remote_ref_actions: (element-type EuuFlatpakRemoteRefAction): A #GPtrArray
+ * @priority: the priority of the file
```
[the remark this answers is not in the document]
smspillaz: Higher numbers take priority, See `test_update_deploy_flatpaks_on_reboot_override_ostree`. So the docs should be adjusted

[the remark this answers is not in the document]
pwithnall: The documentation (and implementation) for `euu_flatpak_ref_actions_append_from_directory()` have lower numeric values as more important. This is the priority which comes from the directory’s position in the `directories_to_search` in `euu_flatpak_ref_actions_from_paths()`; not the priority which comes from the file’s lexicographical position within its directory.

mwleeds:
I'll leave this for @pwithnall to merge after looking at my changes to "libeuu: Ignore non-JSON files when loading autoinstall files from a dir"

PR #189, opened 2018-04-13
T22054: Pull and install dependency ref actions

smspillaz:
This adds a new function `euu_add_dependency_ref_actions_for_installation`. Given a flattened autoinstall list and a `FlatpakInstallation`, it will figure out what dependencies need to be added (taken from `flatpak_installation_list_remote_related_refs_sync`) and add appropriate actions for those to the autoinstall list.

This PR also contains a few incidental changes, required to make this work, such as signing all exported builds.

https://phabricator.endlessm.com/T22054


test-common/utils.c
```diff
@@ -2104,125 +2104,307 @@ format_flatpak_ref_name_with_branch_override_arch (const gchar *name,
...
+                                          formatted_ref_name,
+                                          error))
+                    return FALSE;
+                }
+            }
+            break;
+          case FLATPAK_INSTALL_INFO_TYPE_APP:
+            break;
+          default:
+            g_assert_not_reached();
```
pwithnall: Nitpick: Missing space before `(`.

test-common/utils.c
```diff
@@ -2104,125 +2104,307 @@ format_flatpak_ref_name_with_branch_override_arch (const gchar *name,
...
+                                                           const gchar **preinstall_flatpak_names,
+                                                           GError      **error)
+{
+  g_autoptr(GHashTable) repo_infos = g_hash_table_new_full (g_str_hash,
+                                                            g_str_equal,
+                                                            g_free,
+                                                            (GDestroyNotify) flatpak_repo_info_free);
+  g_autoptr(GPtrArray) flatpak_install_infos = g_ptr_array_new_full (g_strv_length ((GStrv) flatpak_names) + 1,
+                                                                     (GDestroyNotify) flatpak_install_info_free);
+  const gchar **flatpak_names_iter = flatpak_names;
```
pwithnall: Nitpick: Move this assignment down into the `for` loop, since it isn’t needed earlier, and it’s more readable to keep all the loop logic together

test-common/utils.h
```diff
@@ -285,6 +290,44 @@ gboolean eos_test_setup_flatpak_repo_with_preinstalled_apps_simple (GFile
                                                                     const gchar **preinstalled_flatpak_names,
                                                                     GError      **error);
 
+typedef enum _FlatpakInstallInfoType {
+  FLATPAK_INSTALL_INFO_TYPE_RUNTIME,
+  FLATPAK_INSTALL_INFO_TYPE_APP,
+  FLATPAK_INSTALL_INFO_TYPE_EXTENSION
```
pwithnall: Nitpick: Add a trailing comma to reduce future diff noise

test-common/utils.h
```diff
@@ -285,6 +290,44 @@ gboolean eos_test_setup_flatpak_repo_with_preinstalled_apps_simple (GFile
...
+} FlatpakInstallInfoType;
+
+typedef struct _FlatpakInstallInfo {
```
pwithnall: Nitpick: You don’t need the `_FlatpakInstallInfo` name. Just do
```
typedef struct
{
  …
} FlatpakInstallInfo;
```

test-common/utils.h
```diff
@@ -285,6 +290,44 @@ gboolean eos_test_setup_flatpak_repo_with_preinstalled_apps_simple (GFile
...
+
+void flatpak_install_info_free (FlatpakInstallInfo *info);
+
+typedef struct _FlatpakRepoInfo {
+  gchar *name;
+  gchar *collection_id;
+  gchar *remote_collection_id;
+} FlatpakRepoInfo;
+
+
```
pwithnall: Nitpick: Spurious double empty line

src/eos-updater-fetch.c
```diff
@@ -589,12 +589,11 @@ perform_action_preparation (FlatpakInstallation        *installation,
 }
 
 static gboolean
-pull_flatpaks (GPtrArray     *pending_flatpak_ref_actions,
-               GCancellable  *cancellable,
-               GError       **error)
+pull_flatpaks (FlatpakInstallation  *installation,
+               GPtrArray            *pending_flatpak_ref_actions,
+               GCancellable         *cancellable,
+               GError              **error)
```
pwithnall: Rearranging this function to take `FlatpakInstallation` as a parameter could be a separate commit.

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,651 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
                                            progresses);
 }
 
+static guint
+flatpak_remote_hash (gconstpointer data)
+{
+  FlatpakRemote *remote = FLATPAK_REMOTE (data);
+  /* The casting dance here is because want to g_direct_hash the priority
```
pwithnall: Nitpick: s/because want/because we want/

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,651 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+static gboolean
+list_all_remote_refs_in_flatpak_installation (FlatpakInstallation  *installation,
+                                              GPtrArray           **out_remotes_priority_order,
+                                              GHashTable          **out_refs_for_remotes,
+                                              GCancellable         *cancellable,
+                                              GError              **error)
+{
+  g_autoptr(GPtrArray) remotes = NULL;
+  g_autoptr(GHashTable) refs_for_remotes = NULL;
+  guint i = 0;
```
pwithnall: Nitpick: `gsize`, because it’s used as an array index

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,651 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+
+  g_return_val_if_fail (out_remotes_priority_order != NULL, FALSE);
+  g_return_val_if_fail (out_refs_for_remotes != NULL, FALSE);
```
pwithnall: Nitpick: Could also add preconditions for `cancellable`, `installation` and `error`:
```
g_return_val_if_fail (FLATPAK_IS_INSTALLATION (installation), FALSE);
g_return_val_if_fail (cancellable == NULL || G_IS_CANCELLABLE (cancellable), FALSE);
g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
```

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,651 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+
+static gboolean
+list_all_remote_refs_in_flatpak_installation (FlatpakInstallation  *installation,
+                                              GPtrArray           **out_remotes_priority_order,
+                                              GHashTable          **out_refs_for_remotes,
+                                              GCancellable         *cancellable,
+                                              GError              **error)
+{
+  g_autoptr(GPtrArray) remotes = NULL;
+  g_autoptr(GHashTable) refs_for_remotes = NULL;
```
pwithnall: Note: I think I’m going to relax the requirement for declarations to be at the top of the function in eos-updater code. Having written a few projects with inline declarations now, I can say I’m a convert. I don’t expect you to change any of this code to use inline declarations though — just an FYI for future patches!

smspillaz:
> Review of “test: Should probably install the runtime from repo_name, not "test-r…”: Should this not be merged in to other commits?

Actually this was something that I missed earlier when this feature was first written. But the change  libeos-updater-flatpak-installer/tests/installer.c was caught in there which shouldn't be there. Will fix.

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,651 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+  g_clear_object (&related_refs_for_remote->remote);
+  g_clear_pointer (&related_refs_for_remote->related_refs, (GDestroyNotify) g_hash_table_unref);
+
+  g_free (related_refs_for_remote);
+}
+
+/* The runtime ref is a special case that we cannot get from
+ * flatpak_installation_fetch_remote_metadata_sync. We need to manually
+ * fetch the metadata for each remote (this will usually come from the cache)
+ * and parse it for the runtime ref. */
```
pwithnall: Any chance of upstreaming this to libflatpak?

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,651 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+{
+  g_autoptr(FlatpakRef) runtime_ref = NULL;
+  g_autoptr(FlatpakRemoteRef) runtime_remote_ref = NULL;
+  g_autoptr(GBytes) bytes = NULL;
+  g_autoptr(GKeyFile) metadata = NULL;
+  g_autofree gchar *runtime_ref_str = NULL;
+  g_autofree gchar *runtime_ref_str_complete = NULL;
+  g_autoptr(GError) local_error = NULL;
+
+  g_return_val_if_fail (out_remote_runtime_ref != NULL, FALSE);
```
pwithnall: Might as well add the other preconditions too.

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,651 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+
+  if (bytes == NULL)
+    {
+      /* G_IO_ERROR_NOT_FOUND here means that either the remote did not have
+       * a summary cache or that the source ref was not found in the
+       * summary cache. In either case, we just treat that as a non-fatal
+       * error and continue. */
+      if (g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
+        {
+          *out_remote_runtime_ref = NULL;
```
pwithnall: Maybe add a `g_debug()` here before returning, so it’s easier to work out the control flow from debug logs.

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,651 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+{
+  g_autoptr(GPtrArray) remote_related_refs =
+    flatpak_installation_list_remote_related_refs_sync (installation,
+                                                        remote_name,
+                                                        formatted_ref,
+                                                        cancellable,
+                                                        error);
+  g_autoptr(FlatpakRemoteRef) remote_runtime_ref = NULL;
+  FlatpakRef *remote_runtime_ref_as_ref = NULL;
+  g_autoptr(GError) local_error = NULL;
```
pwithnall: `local_error` is unused.

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,651 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+
+  installed_related_refs =
+    flatpak_installation_list_installed_related_refs_sync (installation,
+                                                           remote_name,
+                                                           formatted_ref,
+                                                           cancellable,
+                                                           error);
+
+  /* Only error that can occurr here is FLATPAK_ERROR_NOT_INSTALLED
```
pwithnall: s/occurr/occur/

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,651 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+  if (installed_ref == NULL)
+    {
+      if (!g_error_matches (local_error, FLATPAK_ERROR, FLATPAK_ERROR_NOT_INSTALLED))
+        {
+          g_propagate_error (error, g_steal_pointer (&local_error));
+          return NULL;
+        }
+    }
+
+  g_ptr_array_add (installed_related_refs, g_steal_pointer (&installed_ref));
```
pwithnall: Should this line be skipped if `installed_ref == NULL`?

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,651 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+                                  EuuFlatpakRemoteRefActionType *out_action_type)
+{
+  g_return_val_if_fail (out_action_type != NULL, FALSE);
+
+  switch (source_ref_action->type)
+    {
+      case EUU_FLATPAK_REMOTE_REF_ACTION_INSTALL:
+        if (flatpak_related_ref_should_download (related_ref))
+          {
+            if (g_hash_table_contains (installed_related_refs, related_ref))
```
pwithnall: You could factor part of this out above:
```
gboolean related_ref_is_installed = g_hash_table_contains (…);
```
and then use `related_ref_is_installed` in the `if` conditions.

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,651 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+}
+
+/**
+ * euu_add_dependency_ref_actions_for_installation:
+ * @installation: A #FlatpakInstallation
+ * @ref_actions: (element-type EuuFlatpakRemoteRefAction): A #GPtrArray of
+ *               #EuuFlatpakRemoteRefAction containing pending ref actions.
+ * @error: A #GError
+ *
+ * Walk throuhg the list of remote ref actions in @ref_actions and
```
pwithnall: through

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,651 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+ * yield a new list with runtime and extension dependencies added. In the
+ * install or update case, runtime dependencies will be added before
+ * the original ref action and extension dependencies will be added after
+ * the original ref action. In the uninstall case extension dependencies
+ * will be added before the original ref action and runtime dependencies
+ * after the original ref action.
+ *
+ * Returns: (transfer full) (element-type EuuFlatpakRemoteRefAction: A #GPtrArray
```
pwithnall: Missing `)` on the `(element-type)` annotation

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,651 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+                                                               ref_action->ref->ref,
+                                                               remotes,
+                                                               remote_related_refs,
+                                                               installed_related_refs,
+                                                               cancellable,
+                                                               error))
+        return FALSE;
+
+      /* If the source ref action is to uninstall then it should be
+       * go before its dependencies. */
```
pwithnall: ‘should be go’

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,651 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+      if (ref_action->type == EUU_FLATPAK_REMOTE_REF_ACTION_UNINSTALL)
+        g_ptr_array_add (dependency_ref_actions,
+                         euu_flatpak_remote_ref_action_ref (ref_action));
+
+      /* Go through each of the related refs and add it to the dependency
+       * ref actions depending on what we are doing with the dependency
+       * and the state of the dependnecy */
```
pwithnall: ‘dependency’

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,651 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+                                   euu_flatpak_remote_ref_action_new (action_type,
+                                                                      location_ref,
+                                                                      ref_action->source,
+                                                                      -1));
+                }
+            }
+        }
+
+      /* If the source ref action is to install or update then it should be
+        * go before its dependencies. */
```
pwithnall: ‘should be go’

pwithnall: s/before/after/

pwithnall: Don’t you need to split it so that (if installing, for example), runtimes get installed first, then the app, then the app’s extension runtimes? Or do we not care about installing platform runtimes like `org.freedesktop.Platform`?

tests/test-update-install-flatpaks.c
```diff
@@ -1388,6 +1388,106 @@ test_update_install_flatpaks_conflicting_location_error (EosUpdaterFixture *fixt
...
+  g_autoptr(GHashTable) flatpak_repo_infos = g_hash_table_new_full (g_str_hash,
+                                                                    g_str_equal,
+                                                                    g_free,
+                                                                    (GDestroyNotify) flatpak_repo_info_free);
+  g_auto(GStrv) flatpaks_in_repo = NULL;
+  g_autoptr(GFile) updater_directory = NULL;
+  g_autofree gchar *updater_directory_str = NULL;
+  g_autoptr(GError) error = NULL;
+
+  g_test_bug ("T16682");
```
pwithnall: s/T16682/T22054/?

tests/test-update-install-flatpaks.c
```diff
@@ -2873,6 +2873,127 @@ test_update_deploy_flatpaks_on_reboot (EosUpdaterFixture *fixture,
...
+  g_autofree gchar *deployment_csum = NULL;
+  g_autofree gchar *refspec = concat_refspec (default_remote_name, default_ref);
+  g_autoptr(GFile) deployment_repo_dir = NULL;
+  g_autoptr(GFile) updater_directory = NULL;
+  g_autofree gchar *updater_directory_str = NULL;
+  g_autofree gchar *keyid = get_keyid (fixture->gpg_home);
+  g_autoptr(GFile) gpg_key_file = get_gpg_key_file_for_keyid (fixture->gpg_home, keyid);
+  g_autoptr(GError) error = NULL;
+
+  g_test_bug ("T16682");
```
pwithnall: s/T16682/T22054/?

libeos-updater-util/flatpak.c
smspillaz: As far as I can tell the flatpak API does not care if extension runtimes are installed completely separately before or after an app. I figured it'd be easier to handle both in the same way (eg, install all the related refs before the app).

test-common/utils.c
```diff
@@ -2042,10 +2084,37 @@ flatpak_install_info_new (FlatpakInstallInfoType  type,
...
+flatpak_install_info_new (FlatpakInstallInfoType  type,
+                          const gchar            *name,
+                          const gchar            *branch,
+                          const gchar            *runtime_name,
+                          const gchar            *runtime_branch,
+                          const gchar            *repo_name,
+                          gboolean                preinstall)
+{
+  g_autoptr(GPtrArray) extension_infos =
+    g_ptr_array_new_with_free_func ((GDestroyNotify) flatpak_extension_point_info_free);
```
pwithnall: Instead of passing an empty `GPtrArray` around, you could just accept `NULL` as equivalent to an empty array in the extension info functions above.

tests/test-update-install-flatpaks.c
```diff
@@ -3906,6 +3906,338 @@ test_update_uninstall_flatpaks_on_reboot (EosUpdaterFixture *fixture,
...
+                                  default_remote_name,
+                                  &error);
+  g_assert_no_error (error);
+
+  /* Extension should not be installed here */
+  deployed_flatpaks = eos_test_get_installed_flatpaks (updater_directory, &error);
+  g_assert_no_error (error);
+
+  g_assert_false (g_strv_contains ((const gchar * const *) deployed_flatpaks,
+                                   "org.test.Test.Extension"));
```
pwithnall: You should probably also verify in the middle step that the app and its extension are installed.

pwithnall: You should also verify here that the app is uninstalled.

tests/test-update-install-flatpaks.c
```diff
@@ -3906,6 +3906,338 @@ test_update_uninstall_flatpaks_on_reboot (EosUpdaterFixture *fixture,
...
+                                  default_remote_name,
+                                  &error);
+  g_assert_no_error (error);
+
+  /* Extension should not be installed here */
+  deployed_flatpaks = eos_test_get_installed_flatpaks (updater_directory, &error);
+  g_assert_no_error (error);
+
+  g_assert_true (g_strv_contains ((const gchar * const *) deployed_flatpaks,
+                                  "org.test.Test.Extension"));
```
pwithnall: You should also verify in the middle step that both the app and extension are installed; and you should verify here that the app is uninstalled.

libeos-updater-util/flatpak.c
pwithnall: Probably worth putting that in a comment.

libeos-updater-util/flatpak.c
smspillaz: Yeah I tried to do that in https://github.com/flatpak/flatpak/pull/1578 but Alex didn't like the idea of putting it with the remote refs. He mentioned that there is a longer term goal to export such an API flatpak/flatpak#1234

pwithnall: ⇒ Add that as a `FIXME` comment, with links.

libeos-updater-util/flatpak.c
smspillaz: It should be, yes

tests/test-update-install-flatpaks.c
smspillaz: Hmm, my preference is that tests have a single assertion set at the end, but I think in this case I'm willing to break that rule (we should make sure that the app and extension got preinstalled by the framework)

smspillaz:
> Review: “fetch: Invariant is that there must be a collection-id OR a remote” — I don’t think this is right. IIRC, we require collection IDs to be set in the JSON, and don’t require remote names to be set. The remote names are then filled in from the collection IDs, which means that eventually both will be set in the EuuFlatpakLocationRef.

Hmm okay, the comment in the code seems to indicate otherwise. I seem to recall that this check was causing some tests to fail but it might be those ones for which we have a collection ID with no associated remote.

tests/test-update-install-flatpaks.c
pwithnall: OOI, why do you hold that preference? That would imply that if you want to test a multi-stage process, you’d need several tests, each one going one stage further than the previous and then doing its assertions.

tests/test-update-install-flatpaks.c
smspillaz: Ah I think this test got pushed down to the end in the diff noise

tests/test-update-install-flatpaks.c
smspillaz: See also: http://xunitpatterns.com/Assertion%20Roulette.html

pwithnall: That testing pattern has some merit, but I’m not convinced it’s an improvement here.

Firstly, when `g_assert()` fails, you get the line number of the failure, so locating the failed assertion is generally trivial (unless it’s in a loop — then you should probably use `g_test_message()` at the start of each loop iteration to print the iterator).

Secondly, in this situation you’re trying to test that at the end both the app and its extension are uninstalled. If you take the pattern above literally, that means doing the test once to assert that the app is uninstalled, and doing it again to assert that the extension is uninstalled, which will double the amount of code and double the runtime.

Thirdly, adding assertions part-way through is essentially one half of asserting that the installation state of the app has changed. It’s protecting against the failure mode where (somehow) the app doesn’t get installed, and hence the final assertion that the app is uninstalled is trivially true (but for the wrong reasons). It’s kind of a take on “who tests the tests?”.

That pattern does have merit in moderation, though, and I’ve seen many tests which essentially test the entire functionality of an object in a single (very long) test case. That’s something to avoid like the plague; not because it’s hard to work out which assertion failed, but rather because it’s hard to work out what state the object under test was in at that point in the test, and whether all states of the object have actually been covered by that monster test case.

pwithnall: (Sorry for the essay.)

smspillaz:
> Hmm okay, the comment in the code seems to indicate otherwise. I seem to recall that this check was causing some tests to fail but it might be those ones for which we have a collection ID with no associated remote.

I did a little bit more investigation and found that the problem here is that a dependency ref action could be pulled in from a remote that does not have a collection-id set on its configuration. When that happens, we don't know what collection-id to use (we can't use the collection-id from the source ref as the dependency could come from a different remote).

For now I've made it so that dependency ref actions don't get pulled from remotes that do not have a collection-id set.

tests/test-update-install-flatpaks.c
smspillaz: Oh weird, my last comment here must not have been saved here. Anyway, too many assertions in a test indicates that a test is testing too much or that it probably could use a nicer fixture (which has its own internal assertions validating that the fixture works correctly). See http://xunitpatterns.com/Assertion%20Roulette.html

smspillaz: Anyway, I added assertions for now.

tests/test-update-install-flatpaks.c
smspillaz: Added these for now

smspillaz:
> Wrt “flatpak: Ignore related refs that don't have a collection-id set” — this (and other commits) could do with some rationale for their changes in their commit messages.

Makes sense. I'll add one tonight

tests/test-update-install-flatpaks.c
smspillaz: I basically agree with you, though I had the following thoughts:

> Secondly, in this situation you’re trying to test that at the end both the app and its extension are uninstalled. If you take the pattern above literally, that means doing the test once to assert that the app is uninstalled, and doing it again to assert that the extension is uninstalled, which will double the amount of code and double the runtime.

If one were to take xunit-test-patterns to the extreme the correct solution to this would be a matcher that checks that a subset of flatpaks are in the list. Unfortunately matchers are highly polymorphic and that doesn't mix with C :( (I tried doing this in the Gjs unit tests, it wasn't pretty).

> Thirdly, adding assertions part-way through is essentially one half of asserting that the installation state of the app has changed. It’s protecting against the failure mode where (somehow) the app doesn’t get installed, and hence the final assertion that the app is uninstalled is trivially true (but for the wrong reasons). It’s kind of a take on “who tests the tests?”.

Yeah, usually you'd have assertions in the fixture itself. Google Test has a distinction between fatal and non-fatal assertions: https://github.com/google/googletest/blob/master/googletest/docs/Primer.md#basic-assertions - in a past life I used the fatal assertions to test the test fixtures and non-fatal assertions to test the actual code.

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,702 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+
+static gboolean
+populate_related_refs_in_all_remotes (FlatpakInstallation  *installation,
+                                      FlatpakRef           *source_ref,
+                                      GPtrArray            *related_refs_for_remotes,
+                                      GCancellable         *cancellable,
+                                      GError              **error)
+{
+  g_autofree gchar *formatted_ref = flatpak_ref_format_ref (source_ref);
+  guint i = 0;
```
pwithnall: Nitpick: Now that #192 has been merged, you could actually move this into the `for` loop:
```
for (gsize i = 0; i < related_refs_for_remotes->len; i++)
```

How exciting!!

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,702 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+      EuuFlatpakRelatedRefsForRemote *related_refs_for_remote =
+        g_ptr_array_index (related_refs_for_remotes, i);
+      g_autoptr(GPtrArray) related_refs =
+        list_related_refs_for_remote (installation,
+                                      flatpak_remote_get_name (related_refs_for_remote->remote),
+                                      source_ref,
+                                      formatted_ref,
+                                      cancellable,
+                                      error);
+      guint j = 0;
```
pwithnall: Same for `j` as for `i`.

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,702 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+populate_installed_related_refs (FlatpakInstallation  *installation,
+                                 FlatpakRef           *ref,
+                                 GPtrArray            *remotes,
+                                 GHashTable           *installed_related_refs_set,
+                                 GCancellable         *cancellable,
+                                 GError              **error)
+{
+  g_autoptr(FlatpakInstalledRef) installed_ref = NULL;
+  g_autofree gchar *formatted_ref = flatpak_ref_format_ref (ref);
+  guint i = 0;
```
pwithnall: Nitpick: Same here with `i`.

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,702 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+                                             error))
+    return FALSE;
+
+  return TRUE;
+}
+
+static void
+initially_populate_remote_and_installed_related_refs (GPtrArray  **out_remote_related_refs,
+                                                      GHashTable **out_installed_related_refs,
+                                                      GPtrArray   *remotes)
```
pwithnall: Nitpick: Typically, the ‘out’ arguments are listed after the ‘in’ ones. So I think `remotes` should be first here.

test-common/utils.h
pwithnall: (Note: This looks like it still applies to a few of the other new structures in the branch.)

libeos-updater-util/flatpak.c
pwithnall: (For future reference, this was done as #192.)

tests/test-update-install-flatpaks.c
pwithnall: I’m not seeing them, but I could be being confused by diff helpfully adding the new tests as modifications of the old ones. Did you definitely push the change which added them?

pwithnall:
> For now I've made it so that dependency ref actions don't get pulled from remotes that do not have a collection-id set.

That makes sense. This might be something we want unit tests for.

libeos-updater-flatpak-installer/perform-flatpak-actions.c
```diff
@@ -379,9 +379,12 @@ eufi_apply_flatpak_ref_actions (FlatpakInstallation       *installation,
           return FALSE;
         }
 
-      g_hash_table_replace (new_progresses,
-                            (gpointer) source,
-                            GINT_TO_POINTER (pending_action->serial));
+      /* If the serial was -1, it means that we should not update the
+       * counter because we were a dependency. */
```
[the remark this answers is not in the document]
smspillaz: Likewise. I'll add the flag.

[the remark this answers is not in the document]
smspillaz: Well, I suppose the flag could just be `pending_action->source == NULL`.

... I can't imagine us writing to a `NULL` file?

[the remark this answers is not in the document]
pwithnall: That definitely deserves to be documented somewhere though.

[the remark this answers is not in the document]
smspillaz: Actually, now that I think of it, a flag makes more sense. Specifying the source allows us to at least show where the dependency came from in the debug output.

libeos-updater-flatpak-installer/perform-flatpak-actions.c
```diff
@@ -449,7 +452,7 @@ eufi_check_ref_actions_applied (FlatpakInstallation  *installation,
   for (i = 0; i < actions->len; ++i)
     {
       EuuFlatpakRemoteRefAction *pending_action = g_ptr_array_index (actions, i);
-      const gchar *name = pending_action->source;
+      const gchar *name = pending_action->source ? pending_action->source : "another source as a dependency";
```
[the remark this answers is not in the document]
smspillaz: Decided to make this always non-null

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,702 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+  runtime_ref = flatpak_ref_parse (runtime_ref_str_complete, error);
+
+  if (runtime_ref == NULL)
+    return FALSE;
+
+  runtime_remote_ref = FLATPAK_REMOTE_REF (g_object_new (FLATPAK_TYPE_REMOTE_REF,
+                                                         "kind", flatpak_ref_get_kind (runtime_ref),
+                                                         "name", flatpak_ref_get_name (runtime_ref),
+                                                         "arch", flatpak_ref_get_arch (runtime_ref),
+                                                         "branch", flatpak_ref_get_branch (runtime_ref),
```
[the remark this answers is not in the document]
smspillaz: Done

smspillaz:
Okay, if you pull the repository `git log` will definitely display them in the right order for squashing. Not sure how to fix the github fail here...

smspillaz:
This now depends on https://github.com/endlessm/eos-updater/pull/189

libeos-updater-flatpak-installer/perform-flatpak-actions.c
```diff
@@ -379,9 +477,12 @@ eufi_apply_flatpak_ref_actions (FlatpakInstallation       *installation,
...
 
-      g_hash_table_replace (new_progresses,
-                            (gpointer) source,
-                            GINT_TO_POINTER (pending_action->serial));
+      /* Don't update the counter for dependencies, they have a non-monotonic
+       * serial number. */
+      if (!is_dependency)
+        g_hash_table_replace (new_progresses,
+                              (gpointer) source,
+                              GINT_TO_POINTER (pending_action->serial));
```
smspillaz: This is no longer necessary. Might better have an assert here actually.

libeos-updater-flatpak-installer/tests/installer.c
```diff
@@ -44,28 +185,59 @@ flatpak_deployments_fixture_setup (FlatpakDeploymentsFixture  *fixture,
...
     NULL
   };
+  const ExtensionInfo extensions[] = {
+    {
+      "org.test.PreinstalledTestHavingExtension",
+      "org.test.PreinstalledTestHavingExtension.Extension",
+      FLATPAK_EXTENSION_POINT_AUTODELETE
+    },
+    { NULL, NULL }
+  };
```
smspillaz: I know, this is getting dangerously close to the existing `FlatpakInstallInfo` and friends structs. I wanted to have these ones since they are easier to populate declaratively.

libeos-updater-util/flatpak.c
```diff
@@ -140,6 +142,20 @@ flatpak_remote_ref_action_type_parse (const gchar                    *action,
...
+{
+  GEnumClass *enum_class = g_type_class_ref (EUU_TYPE_FLATPAK_REMOTE_REF_ACTION_TYPE);
+  GEnumValue *enum_value = g_enum_get_value (enum_class, action_type);
+
+  g_type_class_unref (enum_class);
+
+  g_assert (enum_value != NULL);
+
+  return enum_value->value_nick;
+}
```
smspillaz: Don't need to move this anymore

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1457,702 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+    return g_steal_pointer (&remote_related_refs);
+
+  remote_runtime_ref_as_ref = FLATPAK_REF (remote_runtime_ref);
+  g_ptr_array_add (remote_related_refs, g_object_new (FLATPAK_TYPE_RELATED_REF,
+                                                      "kind", flatpak_ref_get_kind (remote_runtime_ref_as_ref),
+                                                      "name", flatpak_ref_get_name (remote_runtime_ref_as_ref),
+                                                      "arch", flatpak_ref_get_arch (remote_runtime_ref_as_ref),
+                                                      "branch", flatpak_ref_get_branch (remote_runtime_ref_as_ref),
+                                                      "should-delete", FALSE,
+                                                      "should-download", TRUE,
```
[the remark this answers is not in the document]
smspillaz: Done below

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1508,857 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+      FlatpakRemote *remote = g_ptr_array_index (remotes, i);
+      g_autoptr(GPtrArray) installed_related_refs =
+        list_installed_related_refs_for_remote (installation,
+                                                flatpak_remote_get_name (remote),
+                                                source_ref,
+                                                candidate_runtime_ref,
+                                                formatted_ref,
+                                                cancellable,
+                                                error);
+      guint j = 0;
```
smspillaz: This can probably be inlined

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1508,857 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+                                                 GPtrArray            *ref_actions,
+                                                 GCancellable         *cancellable,
+                                                 GError              **error)
+{
+  g_autoptr(GPtrArray) dependency_ref_actions =
+    g_ptr_array_new_with_free_func ((GDestroyNotify) euu_flatpak_remote_ref_action_unref);
+  g_autoptr(GPtrArray) remotes = NULL;
+  g_autoptr(GHashTable) refs_for_remotes = NULL;
+  g_autoptr(GHashTable) pending_install_related_refs = NULL;
+  guint i = 0;
```
smspillaz: Lets put this inline

src/eos-updater-fetch.c
smspillaz: Maybe ... that first commit would not use the FlatpakInstallation parameter however...

smspillaz:
This no longer depends on the linked flatpak PR.

libeos-updater-flatpak-installer/perform-flatpak-actions.c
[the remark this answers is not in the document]
ramcq: If this is always non-null, is this conditional needed?

libeos-updater-flatpak-installer/perform-flatpak-actions.c
```diff
@@ -224,7 +222,106 @@ try_uninstall_application (FlatpakInstallation  *installation,
...
+                                         error))
+    return FALSE;
+
+  g_hash_table_iter_init (&iter, uninstall_dependencies_set);
+  while (g_hash_table_iter_next (&iter, &key, &value))
+    {
+      FlatpakRef *uninstall_ref = key;
+      g_autofree gchar *formatted_related_ref = flatpak_ref_format_ref (uninstall_ref);
+
+      g_debug ("Uninstalling autodelete dependency %s of %s",
```
ramcq: So, hmm... what happens if an extension is marked as autodelete but is shared between more than one Flatpak...?

smspillaz: I'm not sure. Will need to check what flatpak itself does

tests/test-update-install-flatpaks.c
```diff
@@ -2434,90 +4157,88 @@ test_update_flatpak_pull_fail_system_not_deployed (EosUpdaterFixture *fixture,
...
 
-
-/* Insert a list of flatpaks to automatically install on the commit
- * and ensure that they are not installed before reboot */
+/* Insert a list of flatpaks to automatically install on the commit,
+ * including a flatpak that has an extension marked autodownload. Pull the update
+ * and kill the network connection by deleting the remote repositories. Simulate
+ * a reboot by running eos-updater-flatpak-installer. Both the incoming package
+ * and the extension marked autodownload should be installed. (In this case,
+ * related refs resolution should happen from the cached ostree remote
```
ramcq: really? or local metadata after installing the app?

tests/test-update-install-flatpaks.c
```diff
@@ -2282,87 +3868,178 @@ test_update_deploy_fail_flatpaks_not_deployed (EosUpdaterFixture *fixture,
...
+  /* Assert that the counter was not touched */
+  g_assert_false (g_file_test (updater_state_file_path, G_FILE_TEST_EXISTS));
+}
+
+/* Insert a list of flatpaks to automatically install on the commit,
+ * including a flatpak that has a dependency on a runtime that is not
+ * yet installed. Pull the update, kill the network connection by
+ * deleting the remote ostree repo and then simulate a reboot by running
+ * eos-updater-flatpak-installer installer. Both the incoming package
+ * and the dependency should be installed (in this case, dependency
```
ramcq: Is this comment still correct? The dependencies should already be installed and there is no reliance on a cache.

libeos-updater-flatpak-installer/perform-flatpak-actions.c
ramcq: If this is broken, let's split it out to a separate ticket and possibly even wait for Flatpak to have one real solution in the transaction stuff. We're not uninstalling anything with autodelete extensions for the moment, and it's not something we have to worry about getting right before the reboot - it's only run in the "updated" updater after reboot.

ramcq:
Looks reasonable in overall size/shape/logic, I've not done an in-depth review but picked up a few possibly outdated comments - not read the commit messages but might be worth a scan there too for outdated wording about the algorithm/approach.

libeos-updater-flatpak-installer/perform-flatpak-actions.c
smspillaz: So flatpak just unconditionally adds the related ref to the uninstall list: https://github.com/flatpak/flatpak/blob/70bb192204e733190d0b7d78773d5cbde385a5cb/common/flatpak-dir.c#L10777

Incidentally, `app` also uninstalls any extension ending in `Debug` and `Locale` too - hurrah for `lib` and `app` divergence :/ .

To be honest, I think his behavior is fine. From what I gather the "autodelete" property is something that exists on the extension point definition and not the extension itself. So if two apps could plausibly use the same extension, it wouldn't be marked as autodelete in one app.

smspillaz: I do see your point about splitting it out into a separate ticket. That's easily done, if we're happy to go down that road.

tests/test-update-install-flatpaks.c
smspillaz: I believe this comment has now bee updated.

libeos-updater-util/flatpak.c
```diff
@@ -1457,6 +1494,854 @@ euu_filter_for_existing_flatpak_ref_actions (GHashTable *ref_actions  /* (elemen
...
+euu_flatpak_related_refs_for_remote_free (EuuFlatpakRelatedRefsForRemote *related_refs_for_remote)
+{
+  g_clear_object (&related_refs_for_remote->remote);
+  g_clear_pointer (&related_refs_for_remote->related_refs, (GDestroyNotify) g_hash_table_unref);
+
+  g_free (related_refs_for_remote);
+}
+
+/* TODO: This is currently a linear scan. Would be nice
+ *       if it were a hashtable to get O(logn) performance */
```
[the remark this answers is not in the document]
smspillaz: Ah yes, I wrote this comment in a hurry and for some reason assumed that hashtables were implemented with a maxheap, which, of course they aren't -__-"

I'll fix the comment.

libeos-updater-util/tests/flatpak.c
```diff
@@ -108,8 +110,8 @@ static void
 test_compress_install_update_as_install (void)
 {
   FlatpakToInstallEntry entries[] = {
-    { EUU_FLATPAK_REMOTE_REF_ACTION_INSTALL, FLATPAK_REF_KIND_APP, "org.test.Test", "stable", 1 },
-    { EUU_FLATPAK_REMOTE_REF_ACTION_UPDATE, FLATPAK_REF_KIND_APP, "org.test.Test", "stable", 2 }
+    { EUU_FLATPAK_REMOTE_REF_ACTION_INSTALL, FLATPAK_REF_KIND_APP, "org.test.Test", "stable", 1, 0 },
```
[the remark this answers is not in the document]
smspillaz: Yeah I think 0 for brevity is better here. Otherwise all the lines wrap on my half-width editors.

src/eos-updater-fetch.c
```diff
@@ -426,11 +426,28 @@ transition_pending_ref_action_collection_ids_to_remote_names (FlatpakInstallatio
...
+static inline FlatpakInstallFlags
+install_flags_for_action_flags (EuuFlatpakRemoteRefActionFlags action_flags)
+{
+  return (action_flags & EUU_FLATPAK_REMOTE_REF_ACTION_FLAG_IS_DEPENDENCY) == 0 ?
+      FLATPAK_INSTALL_FLAGS_NO_DEPLOY :
+      FLATPAK_INSTALL_FLAGS_NONE;
+}
+
+static inline FlatpakInstallFlags
+update_flags_for_action_flags (EuuFlatpakRemoteRefActionFlags action_flags)
```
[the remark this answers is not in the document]
smspillaz: I'd rather keep them conceptually separate - they return different flags for different flatpak functions.

libeos-updater-util/flatpak.c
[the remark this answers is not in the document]
pwithnall: The comment’s still a `TODO` though — do you intend to fix it before this branch is merged? Otherwise it should be a `FIXME`.

libeos-updater-util/flatpak.h
```diff
@@ -45,6 +45,11 @@ typedef struct {
   const gchar *collection_id;
 } EuuFlatpakLocationRef;
 
+typedef enum {
+  EUU_FLATPAK_REMOTE_REF_ACTION_FLAG_NONE = 0,
+  EUU_FLATPAK_REMOTE_REF_ACTION_FLAG_IS_DEPENDENCY = 1 << 0
+} EuuFlatpakRemoteRefActionFlags;
```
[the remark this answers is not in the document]
pwithnall: The gtk-doc comment documents `_NONE` twice.

src/eos-updater-fetch.c
```diff
@@ -457,83 +474,89 @@ perform_install_preparation (FlatpakInstallation    *installation,
...
+  else
     {
-      g_autofree gchar *formatted_ref = flatpak_ref_format_ref (ref->ref);
-      g_message ("Error occurred whilst pulling flatpak %s:%s: %s",
-                 remote,
-                 formatted_ref,
-                 local_error->message);
-      g_propagate_error (error, g_steal_pointer (&local_error));
-      return FALSE;
+      if ((action_flags & EUU_FLATPAK_REMOTE_REF_ACTION_FLAG_IS_DEPENDENCY) == 0)
```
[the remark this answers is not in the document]
pwithnall: I think the sense of the check has now been flipped (in commit 8fe4d4172eb82ff0913a4b632ff2385ff2bede03):
 * The old `(action_flags & EUU_FLATPAK_REMOTE_REF_ACTION_FLAG_IS_DEPENDENCY) == 0` is equivalent to
 * `!(action_flags & EUU_FLATPAK_REMOTE_REF_ACTION_FLAG_IS_DEPENDENCY)`, which is what `install_flags_for_action_flags()` checks to return
 * `FLATPAK_INSTALL_FLAGS_NO_DEPLOY`, which is *not* equivalent to the new
 * `!(install_flags & FLATPAK_INSTALL_FLAGS_NO_DEPLOY)`

pwithnall:
I have a few comments on some of the fixups, above. They need fixing before this can be merged.

tests/test-update-install-flatpaks.c
```diff
@@ -363,6 +364,26 @@ autoinstall_flatpaks_files_override (GFile                   *updater_directory,
...
+
+  for (; *ostree_refs_stdout_lines_iter; ++ostree_refs_stdout_lines_iter)
+    {
+      if (g_strcmp0 (*ostree_refs_stdout_lines_iter, "") == 0)
+        continue;
+
+      parsed_out_ostree_refs[i++] = g_strdup (*ostree_refs_stdout_lines_iter);
+    }
+
+  return g_steal_pointer (&parsed_out_ostree_refs);
```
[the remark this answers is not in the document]
pwithnall: Nitpick: `drop_empty_lines()` could be made quite a bit more efficient by not copying to a second array, and just rearranging the result of `g_strsplit()` to drop the empty lines (especially since they’re probably only at the end), but it’s test code and nobody cares about performance, so don’t bother addressing this. Just FYI.

smspillaz:
@pwithnall Just a heads up that there's one last commit here (test-common: Don't preinstall runtimes more than once) which was necessary to fix a test failure on Jenkins.

smspillaz:
Lets do this

PR #191, opened 2018-04-16
T21756 preparation

mwleeds:
These commits had to broken off from the other T21756 commits because if I add EosUpdaterUtil-0 to `libeos_updater_flatpak_installer_EosUpdaterFlatpakInstaller_@EUFI_API_VERSION@_gir_INCLUDES` in the same commit that adds `EosUpdaterUtil-0.gir` as a target the build fails because the gir isn't installed at build time. So after this is merged, https://github.com/endlessm/eos-updater/pull/188 will be able to have that dependency.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -313,7 +313,7 @@ eos_updater_flatpak_installer_eos_updater_flatpak_installer_LDADD = \
 	$(CODE_COVERAGE_LIBS) \
 	$(EOS_UPDATER_FLATPAK_INSTALLER_LIBS) \
 	$(top_builddir)/libeos-updater-util/libeos-updater-util-@EUU_API_VERSION@.la \
-	$(top_builddir)/libeos-updater-flatpak-installer/libeos-updater-flatpak-installer-@EUFI_API_VERSION@.la
+	$(top_builddir)/libeos-updater-flatpak-installer/libeos-updater-flatpak-installer-@EUFI_API_VERSION@.la \
 	$(NULL)
 eos_updater_flatpak_installer_eos_updater_flatpak_installer_SOURCES = \
 	eos-updater-flatpak-installer/main.c \
--- a/Makefile.am
+++ b/Makefile.am
@@ -283,7 +283,7 @@ libeos_updater_flatpak_installer_libeos_updater_flatpak_installer_@EUFI_API_VERS
 libeos_updater_flatpak_installer_libeos_updater_flatpak_installer_@EUFI_API_VERSION@_la_LIBADD = \
 	$(CODE_COVERAGE_LIBS) \
 	$(EOS_UPDATER_FLATPAK_INSTALLER_LIBS) \
-	$(top_builddir)/libeos-updater-util/libeos-updater-util-@EUFI_API_VERSION@.la \
+	$(top_builddir)/libeos-updater-util/libeos-updater-util-@EUU_API_VERSION@.la \
 	$(NULL)
 libeos_updater_flatpak_installer_libeos_updater_flatpak_installer_@EUFI_API_VERSION@_la_SOURCES = \
 	libeos-updater-flatpak-installer/determine-flatpak-actions.c \
--- a/Makefile.am
+++ b/Makefile.am
@@ -23,6 +23,7 @@ sysconfexampledir = $(pkgdatadir)
 # Pre-defines to allow appending later
 libexec_PROGRAMS =
 noinst_LTLIBRARIES =
+lib_LTLIBRARIES =
 systemdsystemunit_DATA =
 dist_sysconfexample_DATA =
 dist_bin_SCRIPTS =
@@ -38,7 +39,7 @@ BUILT_SOURCES =
 EXTRA_DIST += README.md
 
 # Ensure systemd units get installed under $(prefix) for distcheck
-AM_DISTCHECK_CONFIGURE_FLAGS = --with-systemdsystemunitdir='$${libdir}/systemd/system' --enable-introspection
+AM_DISTCHECK_CONFIGURE_FLAGS = --with-systemdsystemunitdir='$${libdir}/systemd/system'
 
 pkgconfigdir = $(libdir)/pkgconfig
 pkgconfig_DATA =
@@ -58,7 +59,7 @@ AM_CPPFLAGS = \
 	$(NULL)
 
 # Utility library
-noinst_LTLIBRARIES += libeos-updater-util/libeos-updater-util-@EUU_API_VERSION@.la
+lib_LTLIBRARIES += libeos-updater-util/libeos-updater-util-@EUU_API_VERSION@.la
 
 enum_header_files = \
 	$(top_srcdir)/libeos-updater-util/flatpak.h \
@@ -264,7 +265,7 @@ dist_bin_SCRIPTS += eos-updater-prepare-volume/eos-updater-prepare-volume
 dist_man8_MANS += eos-updater-prepare-volume/docs/eos-updater-prepare-volume.8
 
 # Flatpak installer utility library
-noinst_LTLIBRARIES += libeos-updater-flatpak-installer/libeos-updater-flatpak-installer-@EUFI_API_VERSION@.la
+lib_LTLIBRARIES += libeos-updater-flatpak-installer/libeos-updater-flatpak-installer-@EUFI_API_VERSION@.la
 
 libeos_updater_flatpak_installer_libeos_updater_flatpak_installer_@EUFI_API_VERSION@_la_CPPFLAGS = \
 	$(AM_CPPFLAGS) \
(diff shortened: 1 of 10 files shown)


eos-updater-flatpak-installer/docs/eos-updater-flatpak-installer.8
```diff
@@ -10,6 +10,8 @@ eos\-updater\-flatpak\-installer — Endless OS Updater Flatpak Installer
 .IX Header "SYNOPSIS"
 .\"
 \fBeos\-updater\-flatpak\-installer [\-m \fPmode\fB] [\-p]
+.PP
+\fBeos\-updater\-flatpak\-installer \--dry-run
```
pwithnall: You need to escape each hyphen, so this should end with `\-\-dry\-run`. Isn’t troff/groff fun?

eos-updater-flatpak-installer/docs/eos-updater-flatpak-installer.8
```diff
@@ -23,6 +25,11 @@ the old OS deployment is still running.
 .SH OPTIONS
 .IX Header "OPTIONS"
 .\"
+.IP "\fB\-\-dry-run\fP"
```
pwithnall: Same here.

eos-updater-flatpak-installer/docs/eos-updater-flatpak-installer.8
```diff
@@ -23,6 +25,11 @@ the old OS deployment is still running.
...
+Print to stdout the actions that have been applied and the actions left to
```
pwithnall: No need to mention ‘stdout’. How about “Print the actions that have been applied…”?

eos-updater-flatpak-installer/docs/eos-updater-flatpak-installer.8
```diff
@@ -40,7 +47,7 @@ the updated actions to
 When set to \fBcheck\fP, the tool will check to see if all actions are applied (e.g., that
 applications that should have been installed are installed and every app that
 should have been uninstalled is not installed).
-.\"
+\".
```
pwithnall: Nitpick: This should be in a separate commit.

eos-updater-flatpak-installer/main.c
```diff
@@ -147,6 +149,38 @@ main (int    argc,
...
+                     "Could not get information on which flatpak ref actions to apply: %s",
+                     error->message);
+
+      formatted_flatpak_ref_actions_to_apply =
+        euu_format_all_flatpak_ref_actions ("All flatpak ref actions that are not yet applied",
+                                            new_flatpak_ref_actions_to_apply);
+      g_print ("%s\n", formatted_flatpak_ref_actions_to_apply);
+
+      return EXIT_OK;
+    }
```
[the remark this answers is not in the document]
pwithnall: Yeah, duplicating the code paths to implement `--dry-run` isn’t the most maintainable.

Makefile.am
```diff
@@ -283,7 +283,7 @@ libeos_updater_flatpak_installer_libeos_updater_flatpak_installer_@EUFI_API_VERS
 libeos_updater_flatpak_installer_libeos_updater_flatpak_installer_@EUFI_API_VERSION@_la_LIBADD = \
 	$(CODE_COVERAGE_LIBS) \
 	$(EOS_UPDATER_FLATPAK_INSTALLER_LIBS) \
-	$(top_builddir)/libeos-updater-util/libeos-updater-util-@EUFI_API_VERSION@.la \
+	$(top_builddir)/libeos-updater-util/libeos-updater-util-@EUU_API_VERSION@.la \
```
pwithnall: Good catch. :see_no_evil:

Makefile.am
```diff
@@ -58,7 +59,7 @@ AM_CPPFLAGS = \
 	$(NULL)
 
 # Utility library
-noinst_LTLIBRARIES += libeos-updater-util/libeos-updater-util-@EUU_API_VERSION@.la
+lib_LTLIBRARIES += libeos-updater-util/libeos-updater-util-@EUU_API_VERSION@.la
```
pwithnall: These libraries will also need to be listed in `debian/*.install`, and you will need to add a new subpackage containing the GIR files. There’s some convention for naming the GIR subpackage which I can’t remember, but I’m sure you can cargo-cult it easily enough.

Makefile.am
```diff
@@ -295,6 +296,67 @@ libeos_updater_flatpak_installer_libeos_updater_flatpak_installer_@EUFI_API_VERS
...
+	$(libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_SOURCES)
+libeos_updater_util_EosUpdaterUtil_@EUU_API_VERSION@_gir_LIBS = libeos-updater-util/libeos-updater-util-@EUU_API_VERSION@.la
+libeos_updater_util_EosUpdaterUtil_@EUU_API_VERSION@_gir_NAMESPACE = EosUpdaterUtil
+libeos_updater_util_EosUpdaterUtil_@EUU_API_VERSION@_gir_SCANNERFLAGS = \
+	$(WARN_SCANNERFLAGS) \
+	--nsversion=@EUU_API_VERSION@ \
+	--c-include="libeos-updater-util/util.h" \
+	--identifier-prefix=EosUpdater \
+	--identifier-prefix=Euu \
+	--symbol-prefix=euu \
```
pwithnall: Do you also need `--symbol-prefix=eos_updater`?

Makefile.am
```diff
@@ -295,6 +296,67 @@ libeos_updater_flatpak_installer_libeos_updater_flatpak_installer_@EUFI_API_VERS
...
+	-I$(top_srcdir) \
```
pwithnall: I’d drop the `-I` here and add `$(libeos_updater_util_libeos_updater_util_@EUU_API_VERSION@_la_CPPFLAGS)` to `libeos_updater_util_EosUpdaterUtil_@EUU_API_VERSION@_gir_CFLAGS`. Ensures they’re built with the same set of preprocessor flags.

Makefile.am
```diff
@@ -295,6 +296,67 @@ libeos_updater_flatpak_installer_libeos_updater_flatpak_installer_@EUFI_API_VERS
...
+	$(NULL)
+
+INTROSPECTION_GIRS += libeos-updater-util/EosUpdaterUtil-@EUU_API_VERSION@.gir
+
+# GObject Introspection for libeos-updater-flatpak-installer
+libeos-updater-flatpak-installer/EosUpdaterFlatpakInstaller-@EUFI_API_VERSION@.gir: libeos-updater-flatpak-installer/libeos-updater-flatpak-installer-@EUFI_API_VERSION@.la
```
pwithnall: Same comments for this GIR file as above.

libeos-updater-flatpak-installer/determine-flatpak-actions.c
```diff
@@ -79,7 +79,18 @@ flatpak_ref_actions_and_progresses (GStrv        directories_to_search,
...
+ * @directories_to_search: (nullable): directories to search for JSON files
+ * @error: return location for a #GError, or %NULL
+ *
+ * Search each directory in @directories_to_search for JSON files (as defined
+ * by eos-updater-flatpak-autoinstall.d(5)) and find actions that should
+ * already have been applied.
+ *
+ * Returns: (element-type filename EuuFlatpakRemoteRefAction) (transfer full):
+ * a mapping from file names to actions that should have been applied
+ */
```
pwithnall: Nitpick: I’d factor all the gtk-doc comment changes out into separate commits, because they’re valuable even if the introspection build changes weren’t to be committed.

libeos-updater-flatpak-installer/determine-flatpak-actions.c
```diff
@@ -97,7 +108,18 @@ eufi_determine_flatpak_ref_actions_to_check (GStrv    directories_to_search,
...
+/**
+ * eufi_determine_flatpak_ref_actions_to_apply:
+ * @directories_to_search: (nullable): directories to search for JSON files
+ * @error: return location for a #GError, or %NULL
+ *
+ * Search each directory in @directories_to_search for JSON files (as defined
+ * by eos-updater-flatpak-autoinstall.d(5)) and find actions that should
+ * be applied.
+ *
+ * Returns: (element-type filename EuuFlatpakRemoteRefAction) (transfer full):
```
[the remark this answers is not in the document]
pwithnall: Nope, `(element-type filename EuuFlatpakRemoteRefAction)` is correct here. [See the documentation](https://wiki.gnome.org/Projects/GObjectIntrospection/Annotations#Type_signature). The `(element-type)` annotation for hash tables has to contain the types of both the key and the value.

Whether `EuuFlatpakRemoteRefAction` is bindable is another question. GIR should be happy with it being listed in the GIR XML as `introspectable="0"`, and the rest of the API will be usable as long as you don’t want to use `eufi_determine_flatpak_ref_actions_to_apply()` from Python.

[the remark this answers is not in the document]
smspillaz: Ah of course. I forgot that this was a hashtable amd has a different syntax. Ignore my comment :)

eos-updater-flatpak-installer/docs/eos-updater-flatpak-installer.8
mwleeds: If you just run it on the command line the output is pretty messy because there's other stuff on stderr

pwithnall: OK.

eos-updater-flatpak-installer/docs/eos-updater-flatpak-installer.8
mwleeds: I don't understand this change or know why it's in there, so I guess I'll take it out.

mwleeds:
I pushed a fixed branch. I would have done fixups but with this small amount of changes I don't think it matters much.

debian/control
```diff
@@ -106,3 +106,18 @@ Description: Updater for Endless OS - utility tests
...
  other update packaging tools.
+
+Package: gir1.2-eos-updater-0
+Section: introspection
+Architecture: any
+Multi-arch: same
+Depends:
+ ${gir:Depends},
+ ${misc:Depends},
+ ${shlibs:Depends},
```
pwithnall: Not sure the `${shlibs:Depends}` is relevant here, since there aren’t any shared libraries in `gir1.2-eos-updater-0.install`.

PR #193, opened 2018-04-25
Cancel fetching when Mogwai asks it to not download

joaquimrocha:
Fetching the updates was honoring the 'download-now' property from
Mogwai's Scheduled Entry for starting, but once it started, it'd not get
canceled even if Mogwai reported download-now=false for the entry.

This patch implements that cancellation by using a local cancellable
object for the fetching operation instead of the one that's passed to
the call, since the latter should only be cancelled by the Cancel()
DBus method (in order to avoid future issues).

https://phabricator.endlessm.com/T22237

### DIFF
--- a/src/eos-updater-fetch.c
+++ b/src/eos-updater-fetch.c
@@ -64,6 +64,35 @@ fetch_data_free (FetchData *data)
 
 G_DEFINE_AUTOPTR_CLEANUP_FUNC (FetchData, fetch_data_free)
 
+typedef struct {
+  MwscScheduleEntry *scheduled_entry; /* (owned) */
+  GCancellable *general_cancellable; /* (owned) */
+  GCancellable *scheduled_entry_cancellable; /* (owned) */
+  gulong download_now_handler_id;
+  gulong invalidated_handler_id;
+  gulong cancellables_bind_handler_id;
+} ScheduledEntryCancellableHelper;
+
+static void
+schedule_entry_cancellable_helper_free (ScheduledEntryCancellableHelper *helper)
+{
+  if (helper->download_now_handler_id > 0)
+    g_signal_handler_disconnect (helper->scheduled_entry, helper->download_now_handler_id);
+
+  if (helper->invalidated_handler_id > 0)
+    g_signal_handler_disconnect (helper->scheduled_entry, helper->invalidated_handler_id);
+
+  if (helper->cancellables_bind_handler_id > 0)
+    g_cancellable_disconnect (helper->general_cancellable, helper->cancellables_bind_handler_id);
+
+  g_clear_object (&helper->scheduled_entry);
+  g_clear_object (&helper->general_cancellable);
+  g_clear_object (&helper->scheduled_entry_cancellable);
+  g_free (helper);
+}
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (ScheduledEntryCancellableHelper, schedule_entry_cancellable_helper_free)
+
 static void
 content_fetch_finished (GObject *object,
                         GAsyncResult *res,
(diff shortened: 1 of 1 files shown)


src/eos-updater-fetch.c
```diff
@@ -64,6 +64,35 @@ fetch_data_free (FetchData *data)
 
 G_DEFINE_AUTOPTR_CLEANUP_FUNC (FetchData, fetch_data_free)
 
+typedef struct {
+  MwscScheduleEntry *scheduled_entry;
+  GCancellable *general_cancellable;
+  GCancellable *scheduled_entry_cancellable;
```
pwithnall: Please add a `  /* (owned) */` comment after each of these pointers, to document that they hold a strong reference. It’s not an official gobject-introspection annotation, but I hope to get round to proposing it as such someday. It’s useful documentation in any case.

src/eos-updater-fetch.c
```diff
@@ -820,6 +849,34 @@ unschedule_download (FetchData     *fetch_data,
...
+                                 GParamSpec *pspec,
+                                 gpointer    user_data)
+{
+  MwscScheduleEntry *entry = MWSC_SCHEDULE_ENTRY (obj);
+  GCancellable *cancellable = user_data;
+
+  if (!mwsc_schedule_entry_get_download_now (entry))
+    {
+      g_message ("Fetch cancelled because Mogwai's scheduled entry %s can no "
+                 "longer download.", mwsc_schedule_entry_get_id (entry));
```
pwithnall: This function needs a FIXME comment, since Mogwai can actually toggle the status of `download-now` several times during the lifetime of a schedule entry. For example, if an OS update was scheduled and downloading, and then a higher-priority update was scheduled by flatpak, Mogwai would pause the OS update (by setting its `download-now` to false), do the flatpak update, and then could resume the OS update (setting `download-now` to true again) once the flatpak update’s schedule entry was removed.

Supporting resuming updates in eos-updater is a fairly big change, though, and not one I’d expect for this PR. Cancelling the update and re-polling next time the updater is triggered should be OK for now; since OSTree downloads files individually, this should allow progress to be saved between update attempts, if none of them are allowed to complete.

Anyway, in the meantime, a comment here like the following would be good:
```
/* FIXME: Eventually add support for resuming OS update downloads when allowed
 * to by the scheduler. When that happens `get_download_now()` will return
 * %TRUE again; while waiting for that to happen, eos-updater should stay in the
 * `fetching` state. Currently, we go to the `error` state instead. */
```

src/eos-updater-fetch.c
```diff
@@ -964,7 +1071,7 @@ content_fetch (FetchData     *fetch_data,
       if (data->results == NULL)
         g_message ("Fetch: using old code due to lack of repo finder results");
 
-      if (content_fetch_old (fetch_data, context, cancellable, &local_error))
+      if (content_fetch_old (fetch_data, context, fetch_cancellable, &local_error))
```
pwithnall: I suspect you also need to use `fetch_cancellable` in the `prepare_flatpaks_to_deploy()` call below.

joaquimrocha: True! For some reason I thought the `prepare_flatpaks_to_deploy` function was not downloading stuff but I've verified that's not the case.

src/eos-updater-fetch.c
pwithnall: And the other ones which are owned, not just `scheduled_entry_cancellable`.

joaquimrocha: Ah, okay, I thought you meant only the one that's fully owned by this struct (`scheduled_entry_cancellable`). But I see what you mean now.

PR #194, opened 2018-04-26
T22279: Also uninstall autodelete extensions

smspillaz:
This is a follow-up to #189 

If an extension is marked autodelete, we should remove it when the source flatpak gets uninstalled.

### DIFF
--- a/libeos-updater-flatpak-installer/perform-flatpak-actions.c
+++ b/libeos-updater-flatpak-installer/perform-flatpak-actions.c
@@ -192,9 +192,9 @@ try_install_application (FlatpakInstallation       *installation,
 }
 
 static gboolean
-try_uninstall_application (FlatpakInstallation  *installation,
-                           FlatpakRef           *ref,
-                           GError              **error)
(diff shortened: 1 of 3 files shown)


smspillaz:
Currently stacked on #189

smspillaz:
Will need to be rebased once #189 is merged.

pwithnall:
> Will need to be rebased once #189 is merged

It’s merged; this can be rebased.

smspillaz:
(Do we want this anymore? If not, we should close this)

pwithnall:
> (Do we want this anymore? If not, we should close this)

I think the conclusion so far from https://phabricator.endlessm.com/T22279 is that we might want this, but it should be discussed upstream first, since the behaviour of `flatpak uninstall` might need to change as well. That discussion hasn’t happened yet.

I’ve rebased this on `master` as of today (trivial rebase), just so it’s obvious what code hasn’t been merged here. Let’s close this PR but keep the branch in place until the discussion has happened on the task (which I don’t think has a particularly high priority at the moment).

PR #195, opened 2018-04-29
T22295: Check booted branch first before following update refspec

smspillaz:
Previously we would always unconditionally follow the upgrade refspec
if we were booted into a commit that had the update refspec as part
of its metadata. However, such an approach means that the client must
always be able to use the current updater on the checkpoint commit
to upgrade to the most recent commit. This leaves very little room
for error in case the updater on the current commit is not able
to process the new upgrade (for instance, incorrectly configured
flatpak remotes).

Considering that, we might want to "fix" the updater again before
we try upgrading on the branch that really requires it. That can
be done by adding new commits to the currently booted branch. However,
if we do that we need the updater to pay attention to the currently
booted branch and not the "upgrade" branch in the booted commit's
metadata.

This branch does exactly that. It wraps get_refspec_to_upgrade_on
with another function that first pulls on the currently booted branch
to see if there is a new commit and if so uses that, then, if there
is not a new commit, pulls the upgrade-refspec branch for new commits
and attempts to apply them.

Finally, this branch also includes updates to the tests to allow tests
to build up their own non-linear commit graph.

https://phabricator.endlessm.com/T22295

### DIFF
--- a/src/eos-updater-poll.c
+++ b/src/eos-updater-poll.c
@@ -257,6 +257,197 @@ async_result_cb (GObject      *source_object,
   *result_out = g_object_ref (result);
 }
 
+typedef struct {
+  gchar *refspec;
+  gchar *remote;
+  gchar *ref;
+  OstreeCollectionRef *collection_ref;
+  gchar *new_refspec;
+  gchar *checksum;
+  gchar *version;
+  GVariant *commit;
+} UpdateRefInfo;
+
+static void
+update_ref_info_init (UpdateRefInfo *update_ref_info)
+{
+  update_ref_info->refspec = NULL;
+  update_ref_info->remote = NULL;
+  update_ref_info->ref = NULL;
+  update_ref_info->collection_ref = NULL;
+  update_ref_info->new_refspec = NULL;
+  update_ref_info->checksum = NULL;
+  update_ref_info->version = NULL;
+  update_ref_info->commit = NULL;
+}
+
+static void
+update_ref_info_clear (UpdateRefInfo *update_ref_info)
+{
+  g_clear_pointer (&update_ref_info->refspec, g_free);
+  g_clear_pointer (&update_ref_info->remote, g_free);
+  g_clear_pointer (&update_ref_info->ref, g_free);
+  g_clear_pointer (&update_ref_info->collection_ref, ostree_collection_ref_free);
+  g_clear_pointer (&update_ref_info->new_refspec, g_free);
+  g_clear_pointer (&update_ref_info->checksum, g_free);
+  g_clear_pointer (&update_ref_info->version, g_free);
+  g_clear_pointer (&update_ref_info->commit, g_variant_unref);
+}
+
+G_DEFINE_AUTO_CLEANUP_CLEAR_FUNC (UpdateRefInfo, update_ref_info_clear)
+
+static gboolean
+check_for_update_using_booted_branch (OstreeRepo           *repo,
+                                      gboolean             *out_is_update,
+                                      UpdateRefInfo        *out_update_ref_info,
+                                      GCancellable         *cancellable,
+                                      GError              **error)
+{
+  g_autofree gchar *booted_refspec = NULL;
+  g_autofree gchar *remote = NULL;
+  g_autofree gchar *ref = NULL;
+  g_autoptr(OstreeCollectionRef) collection_ref = NULL;
+  g_autofree gchar *new_refspec = NULL;
+  g_autofree gchar *version = NULL;
+  gboolean is_update = FALSE;
+  g_autofree gchar *checksum = NULL;
+  g_autoptr(GVariant) commit = NULL;
+  g_autoptr(OstreeSysroot) sysroot = ostree_sysroot_new_default ();
+  g_autoptr(OstreeDeployment) booted_deployment = NULL;
+
+  g_return_val_if_fail (out_is_update != NULL, FALSE);
+  g_return_val_if_fail (out_update_ref_info != NULL, FALSE);
+  g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+
+  if (!ostree_sysroot_load (sysroot, NULL, error))
+    return FALSE;
+
+  booted_deployment = eos_updater_get_booted_deployment_from_loaded_sysroot (sysroot,
+                                                                             error);
+
+  if (booted_deployment == NULL)
+    return FALSE;
+
+  if (!get_booted_refspec (booted_deployment,
+                           &booted_refspec,
+                           &remote,
+                           &ref,
+                           &collection_ref, error))
+    return FALSE;
+
+  if (!fetch_latest_commit (repo,
+                            cancellable,
+                            booted_refspec,
+                            NULL,
+                            &checksum,
+                            &new_refspec,
+                            &version,
+                            error))
+    return FALSE;
+
+  if (!is_checksum_an_update (repo, checksum, &commit, error))
+    return FALSE;
+
+  is_update = (commit != NULL);
+  *out_is_update = is_update;
+
+  if (is_update)
+    {
+      out_update_ref_info->refspec = g_steal_pointer (&booted_refspec);
+      out_update_ref_info->remote = g_steal_pointer (&remote);
+      out_update_ref_info->ref = g_steal_pointer (&ref);
+      out_update_ref_info->collection_ref = g_steal_pointer (&collection_ref);
+      out_update_ref_info->new_refspec = g_steal_pointer (&new_refspec);
+      out_update_ref_info->checksum = g_steal_pointer (&checksum);
+      out_update_ref_info->version = g_steal_pointer (&version);
+      out_update_ref_info->commit = g_steal_pointer (&commit);
+    }
+  else
+    {
+      update_ref_info_clear (out_update_ref_info);
+    }
+
+  return TRUE;
+}
+
+static gboolean
+check_for_update_following_checkpoint_commits (OstreeRepo     *repo,
+                                               UpdateRefInfo  *out_update_ref_info,
+                                               GCancellable   *cancellable,
+                                               GError        **error)
+{
+  g_autofree gchar *upgrade_refspec = NULL;
+  g_autofree gchar *remote = NULL;
+  g_autofree gchar *ref = NULL;
+  g_autoptr(OstreeCollectionRef) collection_ref = NULL;
+  g_autofree gchar *new_refspec = NULL;
+  g_autofree gchar *version = NULL;
+  g_autofree gchar *checksum = NULL;
+  g_autoptr(GVariant) commit = NULL;
+
+  g_return_val_if_fail (out_update_ref_info != NULL, FALSE);
+  g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+
+  if (!get_refspec_to_upgrade_on (&upgrade_refspec, NULL, NULL, NULL, error))
+    return FALSE;
+
+  if (!fetch_latest_commit (repo,
+                            cancellable,
+                            upgrade_refspec,
+                            NULL,
+                            &checksum,
+                            &new_refspec,
+                            &version,
+                            error))
+    return FALSE;
+
+  out_update_ref_info->refspec = g_steal_pointer (&upgrade_refspec);
+  out_update_ref_info->remote = g_steal_pointer (&remote);
+  out_update_ref_info->ref = g_steal_pointer (&ref);
+  out_update_ref_info->collection_ref = g_steal_pointer (&collection_ref);
+  out_update_ref_info->new_refspec = g_steal_pointer (&new_refspec);
+  out_update_ref_info->checksum = g_steal_pointer (&checksum);
+  out_update_ref_info->version = g_steal_pointer (&version);
+  out_update_ref_info->commit = g_steal_pointer (&commit);
+
+  return TRUE;
+}
+
+static gboolean
+check_for_update_following_checkpoint_if_allowed (OstreeRepo     *repo,
+                                                  UpdateRefInfo  *out_update_ref_info,
+                                                  GCancellable   *cancellable,
+                                                  GError        **error)
+{
+  gboolean had_update_on_branch = FALSE;
+
+  g_return_val_if_fail (out_update_ref_info != NULL, FALSE);
+  g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+
+  /* First, check for an update on the booted refspec. If one exists,
+   * use that, since it may mean that we did some emergency fixes
+   * on the booted refspec after the checkpoint and we don't want
+   * to transition users on to the new branch just yet */
+  if (!check_for_update_using_booted_branch (repo,
+                                             &had_update_on_branch,
+                                             out_update_ref_info,
+                                             cancellable,
+                                             error))
+    return FALSE;
+
+  /* Did we have an update? If not, we can follow the checkpoint */
+  if (!had_update_on_branch)
+    {
+      if (!check_for_update_following_checkpoint_commits (repo,
+                                                          out_update_ref_info,
+                                                          cancellable,
+                                                          error))
+        return FALSE;
+    }
+
+  return TRUE;
+}
+
 /* May return NULL without setting an error if no updates were found. */
 static EosUpdateInfo *
 metadata_fetch_new (OstreeRepo    *repo,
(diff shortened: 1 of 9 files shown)


test-common/utils.c
```diff
@@ -579,14 +583,201 @@ maybe_hashtable_lookup (GHashTable *table,
...
+  GList *commit_keys = g_list_sort (g_hash_table_get_keys (commit_to_ref),
+                                    sort_commits);
+  GList *iter = commit_keys;
+
+  /* Clear the hash-table first */
+
+  for (; iter != NULL; iter = iter->next)
+    {
+      guint commit_id = GPOINTER_TO_UINT (iter->data);
+      OstreeCollectionRef *collection_ref =
```
smspillaz: This could probably be `const`

test-common/utils.c
```diff
@@ -615,34 +807,18 @@ prepare_commit (GFile *repo,
...
-    }
-  else
-    if (!prepare_sysroot_contents (repo,
-                                   tree_root,
-                                   error))
-      return FALSE;
+  if (!prepare_sysroot_contents (repo,
+                                 tree_root,
+                                 error))
+    return FALSE;
```
smspillaz: I am not sure if unconditionally preparing the sysroot contents here is always correct - it seems like we had the `commit_number > 0` check earlier for `prepare_commit` and only did this on commit 0. Then again, it seems harmless from running the tests...

test-common/utils.c
```diff
@@ -698,107 +877,136 @@ generate_delta_files (GFile *repo,
...
- *
- * Returns: the last known ref before wanted_commit_number
- */
-static OstreeCollectionRef *
-get_last_ref (GHashTable *ref_to_commit,
-              guint       wanted_commit_number)
+ * The implementation is a little awkward since we need to do
+ * a linear scan on the parents hashtable to figure out the
+ * children of a given commit. */
+gboolean
```
smspillaz: This could also be implemented as a forward adjacency list ... I guess there's just more stuff to keep track of then (every node will need to have a pointer array of children). Personally I don't mind the O(V) cost to expand children for each node since this is test code.

test-common/utils.c
```diff
@@ -698,107 +877,136 @@ generate_delta_files (GFile *repo,
...
+update_commits (EosTestSubserver *subserver,
+                GError **error)
+{
+  g_auto(CmdResult) cmd = CMD_RESULT_CLEARED;
+
+  if (!eos_test_updater_commit_graph_walk (subserver->commit_graph,
+                                           make_commit_if_not_available,
+                                           subserver,
+                                           error))
+    return FALSE;
```
smspillaz: This could also be implemented through the use of a generator a-la `GHashTableIter`, though I didn't think it was worth the extra complexity.

test-common/utils.h
```diff
@@ -55,6 +55,48 @@ extern const OstreeCollectionRef *default_collection_ref;
...
 
+typedef struct {
+  guint sequence_number;
+  guint parent;
+  OstreeCollectionRef *collection_ref;
+} EosTestUpdaterCommitInfo;
+
+EosTestUpdaterCommitInfo * eos_test_updater_commit_info_new (guint                     sequence_number,
+                                                             guint                     parent,
+                                                             const OstreeCollectionRef *collection_ref);
```
smspillaz: Alignment here

test-common/utils.h
```diff
@@ -55,6 +55,48 @@ extern const OstreeCollectionRef *default_collection_ref;
...
+
+void eos_test_updater_commit_info_free (EosTestUpdaterCommitInfo *info);
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (EosTestUpdaterCommitInfo, eos_test_updater_commit_info_free);
```
smspillaz: Trailing ;

test-common/utils.h
```diff
@@ -55,6 +55,48 @@ extern const OstreeCollectionRef *default_collection_ref;
...
+
+typedef struct {
+  GHashTable *parents;
+  GHashTable *commit_infos;
+  guint       ref_cnt;
+} EosTestUpdaterCommitGraph;
```
smspillaz: This could probably do with some gtk-doc

test-common/utils.h
```diff
@@ -55,6 +55,48 @@ extern const OstreeCollectionRef *default_collection_ref;
...
+
+typedef gboolean (*EosTestUpdaterCommitGraphWalkFunc) (EosTestUpdaterCommitInfo  *commit_info,
+                                                       EosTestUpdaterCommitInfo  *parent_commit_info,
+                                                       gpointer                   user_data,
+                                                       GError                   **error);
+
+gboolean eos_test_updater_commit_graph_walk (EosTestUpdaterCommitGraph          *commit_graph,
+                                             EosTestUpdaterCommitGraphWalkFunc   walk_func,
+                                             gpointer                            walk_func_data,
+                                             GError                            **error);
```
smspillaz: gtk-doc this

test-common/utils.h
```diff
@@ -55,6 +55,48 @@ extern const OstreeCollectionRef *default_collection_ref;
...
+EosTestUpdaterCommitGraph * eos_test_updater_commit_graph_new_from_leaf_nodes (GHashTable *leaf_nodes);
+EosTestUpdaterCommitGraph * eos_test_updater_commit_graph_ref (EosTestUpdaterCommitGraph *commit_graph);
+void eos_test_updater_insert_commit_steal_info (EosTestUpdaterCommitGraph *commit_graph,
+                                                guint                      parent_commit,
+                                                EosTestUpdaterCommitInfo  *commit_info);
+void eos_test_updater_populate_commit_graph_from_leaf_nodes (EosTestUpdaterCommitGraph *commit_graph,
+                                                             GHashTable                *leaf_nodes);
+void eos_test_updater_commit_graph_unref (EosTestUpdaterCommitGraph *commit_graph);
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (EosTestUpdaterCommitGraph, eos_test_updater_commit_graph_unref);
```
smspillaz: ;

test-common/utils.h
```diff
@@ -99,15 +150,18 @@ eos_test_subserver_ref_to_commit_new (void)
                                 (GDestroyNotify) ostree_collection_ref_free, NULL);
 }
 
+
```
smspillaz: Remove extra newline

tests/test-update-direct.c
```diff
@@ -235,9 +237,11 @@ test_cancel_update (EosUpdaterFixture *fixture,
                                   &local_error);
   g_assert_no_error (local_error);
 
-  g_hash_table_insert (subserver->ref_to_commit,
+  g_hash_table_insert (leaf_commit_nodes,
```
smspillaz: This can either be called `ref_to_commit` or `leaf_commit_nodes`. We should probably rename either `eos_test_subserver_ref_to_commit_new` or the `*_leaf_nodes` functions to be consistent here. Personally I prefer "leaf commit nodes" since it makes it clear that the caller needs to populate the graph elsewhere...

src/eos-updater-poll.c
```diff
@@ -257,6 +257,247 @@ async_result_cb (GObject      *source_object,
...
+                                                  gchar               **out_refspec_to_upgrade_on,
+                                                  gchar               **out_remote_to_upgrade_on,
+                                                  gchar               **out_ref_to_upgrade_on,
+                                                  OstreeCollectionRef **out_collection_ref_to_upgrade_on,
+                                                  gchar               **out_redirected_refspec_to_upgrade_on,
+                                                  gchar               **out_checksum,
+                                                  gchar               **out_version,
+                                                  GVariant            **out_commit,
+                                                  GCancellable         *cancellable,
+                                                  GError              **error)
```
smspillaz: Having a function with 11 arguments like this is not ideal :( Ideally what I wanted to be able to do here was have an abstraction that allows callers looking for which commits and branches to follow to do the right thing in a single function call ... without then having to go and fetch all the commit information again later on.

Previously the control flow here was a little more straightforward - we would work out which refspec to pull on, then pull that refspec and work out whether an update was available. Now we need to always pull the booted refspec in order to work out whether or not to even unconditionally use the booted refspec, then figure out if we need to use the checkpoint refspec if available, then pull commits from the right refspec.

Ideally I'd like to avoid pulling commits twice, hence the preservation of all the outparams...

test-common/utils.c
smspillaz: I decided to just revert this to how it was - seems to have not negatively impacted the tests..

test-common/utils.c
pwithnall: Sam, are you reviewing your own code? :octopus: :1st_place_medal:

src/eos-updater-poll.c
pwithnall: You could put the out-params into a structure, allocate that on the stack in the caller, and pass that in to this function by reference. That would reduce the parameter explosion fairly trivially.

test-common/utils.c
```diff
@@ -579,14 +583,201 @@ maybe_hashtable_lookup (GHashTable *table,
...
+eos_test_updater_insert_commit_steal_info (EosTestUpdaterCommitGraph *commit_graph,
+                                           guint                      parent_commit,
+                                           EosTestUpdaterCommitInfo  *commit_info)
+{
+  g_hash_table_insert (commit_graph->commit_infos,
+                       GUINT_TO_POINTER (commit_info->sequence_number),
+                       commit_info);
+  g_hash_table_insert (commit_graph->parents,
+                       GUINT_TO_POINTER (commit_info->sequence_number),
+                       GUINT_TO_POINTER (parent_commit));
```
pwithnall: Why not combine `commit_graph->commit_infos` and `commit_graph->parents` by adding a `parent` field to the `EosTestUpdaterCommitInfo` struct?

test-common/utils.c
```diff
@@ -579,14 +583,201 @@ maybe_hashtable_lookup (GHashTable *table,
...
+  /* Each of the key-value pairs in leaf_nodes points to a candidate
+   * leaf node for a given refspec. From there we recursively go down the
+   * tree and create new EosTestUpdaterCommitInfo objects, unless we
+   * see an entry for that commit in commit_to_ref (we'll start from
+   * that key the next time around */
+  GList *commit_keys = g_list_sort (g_hash_table_get_keys (commit_to_ref),
+                                    sort_commits);
+  GList *iter = commit_keys;
+
+  /* Clear the hash-table first */
```
pwithnall: There appears to be no hash table clearing going on here.

test-common/utils.c
```diff
@@ -579,14 +583,201 @@ maybe_hashtable_lookup (GHashTable *table,
...
+                                                           (CopyFunc) ostree_collection_ref_dup);
+
+  /* Each of the key-value pairs in leaf_nodes points to a candidate
+   * leaf node for a given refspec. From there we recursively go down the
+   * tree and create new EosTestUpdaterCommitInfo objects, unless we
+   * see an entry for that commit in commit_to_ref (we'll start from
+   * that key the next time around */
+  GList *commit_keys = g_list_sort (g_hash_table_get_keys (commit_to_ref),
+                                    sort_commits);
+  GList *iter = commit_keys;
```
pwithnall: Since we now live in the future, this could move into the `for` statement.

test-common/utils.c
```diff
@@ -579,14 +583,201 @@ maybe_hashtable_lookup (GHashTable *table,
...
+                             commit_to_ref);
+    }
+
+  g_list_free (commit_keys);
+}
+
+EosTestUpdaterCommitGraph *
+eos_test_updater_commit_graph_new_from_leaf_nodes (GHashTable *leaf_nodes)
+{
+  EosTestUpdaterCommitGraph *commit_graph = eos_test_updater_commit_graph_new ();
```
pwithnall: Why have `…_commit_graph_new()` and `…_commit_graph_new_from_leaf_nodes()` as separate functions? Just have one `…_commit_graph_new (GHashTable *leaf_nodes)` function which does the same as `…_new_from_leaf_nodes()` when called with a non-`NULL` `leaf_nodes`, and does the same as `…_new()` when called with a `NULL` parameter.

test-common/utils.c
```diff
@@ -698,107 +894,136 @@ generate_delta_files (GFile *repo,
...
+  while (!g_queue_is_empty (&queue))
     {
+      guint commit = GPOINTER_TO_UINT (g_queue_pop_head (&queue));
       GHashTableIter iter;
+      gpointer key, value;
+      EosTestUpdaterCommitInfo *commit_info =
+        g_hash_table_lookup (commit_graph->commit_infos,
+                             GUINT_TO_POINTER (commit));
+      guint parent_commit =
+        GPOINTER_TO_UINT(g_hash_table_lookup (commit_graph->parents,
```
pwithnall: Nitpick: Missing space before `(`.

tests/test-update-refspec-checkpoint.c
```diff
@@ -215,6 +215,340 @@ test_update_refspec_checkpoint (EosUpdaterFixture *fixture,
...
+
+  eos_test_client_has_commit (client,
+                              default_remote_name,
+                              2,
+                              &has_commit,
+                              &error);
+  g_assert_no_error (error);
+  g_assert_false (has_commit);
+
+  /* Now, lets say we screw up something and need to do an update on
```
pwithnall: Nitpick: s/lets/let’s/ (probably in a few other places too)

tests/test-update-refspec-checkpoint.c
```diff
@@ -215,6 +215,340 @@ test_update_refspec_checkpoint (EosUpdaterFixture *fixture,
...
+   * the old branch. Insert another commit, but this time without the
+   * metadata. */
+  eos_test_updater_insert_commit_steal_info (subserver->commit_graph,
+                                             1,
+                                             eos_test_updater_commit_info_new (3,
+                                                                               1,
+                                                                               default_collection_ref));
+  /* For completeness insert a new commit on the checkpoint branch */
```
pwithnall: Does this test work without the ‘for completeness’ commit too?

tests/test-update-refspec-checkpoint.c
```diff
@@ -215,6 +215,340 @@ test_update_refspec_checkpoint (EosUpdaterFixture *fixture,
...
+  eos_test_updater_insert_commit_steal_info (subserver->commit_graph,
+                                             3,
+                                             eos_test_updater_commit_info_new (4,
+                                                                               2,
+                                                                               next_collection_ref));
+  eos_test_subserver_update (subserver,
+                             &error);
+  g_assert_no_error (error);
+
+  /* Update the client client again. Even though we deployed
```
pwithnall: Nitpick: duplicated ‘client’

tests/test-update-refspec-checkpoint.c
```diff
@@ -215,6 +215,340 @@ test_update_refspec_checkpoint (EosUpdaterFixture *fixture,
...
+
+  /* Now, lets say we screw up something and need to do an update on
+   * the old branch. Insert another commit, but this time without the
+   * metadata. */
+  eos_test_updater_insert_commit_steal_info (subserver->commit_graph,
+                                             1,
+                                             eos_test_updater_commit_info_new (3,
+                                                                               1,
+                                                                               default_collection_ref));
+  /* For completeness insert a new commit on the checkpoint branch */
```
pwithnall: Similarly, does this test pass without the ‘for completeness’ commit?

tests/test-update-refspec-checkpoint.c
```diff
@@ -215,6 +215,340 @@ test_update_refspec_checkpoint (EosUpdaterFixture *fixture,
...
+  eos_test_updater_insert_commit_steal_info (subserver->commit_graph,
+                                             3,
+                                             eos_test_updater_commit_info_new (4,
+                                                                               2,
+                                                                               next_collection_ref));
+  eos_test_subserver_update (subserver,
+                             &error);
+  g_assert_no_error (error);
+
+  /* Update the client client again. Even though we deployed
```
pwithnall: Nitpick: client client here too

tests/test-update-refspec-checkpoint.c
```diff
@@ -215,6 +215,340 @@ test_update_refspec_checkpoint (EosUpdaterFixture *fixture,
...
+}
+
+/* Start with a commit, and then make a final commit on the first refspec
+ * which adds a new marker, such that when that commit is deployed, the updater
+ * will know to use a new refspec to upgrade with. However, say we screwed
+ * up and need to do a maintenance fix on the old branch. The commit from the
+ * old branch should be preferred on the next update such that the
+ * old refspec is still in use on reboot. However, later on we create
+ * another checkpoint commit on the newest commit in the old branch. That
+ * should take us to our new branch. */
```
pwithnall: It would be a bit easier to follow this description if there was a little diagram of the commit graph, using the commit numbers used in the code. Something like (numbers and graph totally don’t represent what’s actually going on in this test):
```
1
|
2
/\
3 4
| |\
5 6 7
```

test-common/utils.c
pwithnall: Indeed, as long as the tests run without timing out, I don’t care about performance. Also, we’re talking about ~10 test commits in the graph.

test-common/utils.c
pwithnall: Probably not worth it.

test-common/utils.c
smspillaz: Could do.

test-common/utils.c
smspillaz: Oops, I must have missed this.

smspillaz:
Jenkins passed. Squashed and rebasemerging

src/eos-updater-poll.c
```diff
@@ -257,6 +257,197 @@ async_result_cb (GObject      *source_object,
...
+  g_autoptr(OstreeCollectionRef) collection_ref = NULL;
+  g_autofree gchar *new_refspec = NULL;
+  g_autofree gchar *version = NULL;
+  g_autofree gchar *checksum = NULL;
+  g_autoptr(GVariant) commit = NULL;
+
+  g_return_val_if_fail (out_update_ref_info != NULL, FALSE);
+  g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+
+  if (!get_refspec_to_upgrade_on (&upgrade_refspec, NULL, NULL, NULL, error))
```
dbnicholson: @smspillaz why doesn't this fill in `remote`, `ref` and `collection_ref`? It seems like it will just end up hitting the `update_ref_info.collection_ref == NULL` check you added in `metadata_fetch_new`. Am I missing something?

smspillaz: Hmm, I see your point. The fact that the tests still work is ... concerning. I'll quickly look into what the tests are doing.

smspillaz: Okay, so it looks like the "new" metadata fetch code fails unconditionally (whoops) and we fall back to the old code, which is the reason why all the tests still pass.

I'll push a branch to fix this but I do wonder if we should set `EOS_UPDATER_DISABLE_FALLBACK_FETCHERS` in the tests to catch this in the future?

pwithnall: Fixed in #198. See my comments on https://github.com/endlessm/eos-updater/pull/198#pullrequestreview-116846421 about `EOS_UPDATER_DISABLE_FALLBACK_FETCHERS`.

PR #196, opened 2018-05-01
poll-common: Operate on checkpoint refs, not refspecs

dbnicholson:
The eos.checkpoint-target commit metadata is created on the server side
while the remote is a property of the client's configuration. Following
a remote specified in the metadata would almost certainly be wrong and
provide the server with the ability to switch a client's remote.

Instead, only take the ref from the metadata like the
ostree.endoflife-rebase handling. If a full refspec with remote is
included, the remote is ignored and a warning is issued.

https://phabricator.endlessm.com/T19437

### DIFF
--- a/src/eos-updater-poll-common.c
+++ b/src/eos-updater-poll-common.c
@@ -375,7 +375,6 @@ get_refspec_to_upgrade_on (gchar               **refspec_to_upgrade_on,
   g_autoptr(OstreeSysroot) sysroot = ostree_sysroot_new_default ();
   g_autoptr(OstreeDeployment) booted_deployment = NULL;
   g_autoptr(OstreeRepo) repo = NULL;
-  g_autoptr(GError) local_error = NULL;
 
   if (!ostree_sysroot_load (sysroot, NULL, error))
     return FALSE;
--- a/src/eos-updater-poll-common.c
+++ b/src/eos-updater-poll-common.c
@@ -308,20 +308,22 @@ get_booted_refspec (OstreeDeployment     *booted_deployment,
 }
 
 static gboolean
-get_refspec_to_upgrade_on_from_deployment (OstreeSysroot     *sysroot,
-                                           OstreeDeployment  *booted_deployment,
-                                           gchar            **out_refspec_to_upgrade_from_deployment,
-                                           GError           **error)
+get_ref_to_upgrade_on_from_deployment (OstreeSysroot     *sysroot,
+                                       OstreeDeployment  *booted_deployment,
+                                       gchar            **out_ref_to_upgrade_from_deployment,
+                                       GError           **error)
 {
   const gchar *checksum = ostree_deployment_get_csum (booted_deployment);
   g_autoptr(GVariant) commit = NULL;
   g_autoptr(GVariant) metadata = NULL;
   g_autoptr(GVariant) ref_for_deployment_variant = NULL;
   const gchar *refspec_for_deployment = NULL;
+  g_autofree gchar *remote = NULL;
+  g_autofree gchar *ref = NULL;
   g_autoptr(OstreeRepo) repo = NULL;
   g_autoptr(GError) local_error = NULL;
 
-  g_return_val_if_fail (out_refspec_to_upgrade_from_deployment != NULL, FALSE);
+  g_return_val_if_fail (out_ref_to_upgrade_from_deployment != NULL, FALSE);
 
   if (!ostree_sysroot_get_repo (sysroot, &repo, NULL, error))
    return FALSE;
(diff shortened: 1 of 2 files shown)


tests/test-update-refspec-checkpoint.c
```diff
@@ -561,6 +560,117 @@ test_update_refspec_checkpoint_continue_old_branch_then_new_branch (EosUpdaterFi
...
+
+  eos_test_client_has_commit (client,
+                              default_remote_name,
+                              2,
+                              &has_commit,
+                              &error);
+  g_assert_no_error (error);
+  g_assert_false (has_commit);
+
+  /* Update the client client again. Because we had deployed the
```
pwithnall: Nitpick: client client

Was this copied from somewhere? I thought it was fixed in the other places in the tests.

src/eos-updater-poll-common.c
```diff
@@ -386,41 +391,37 @@ get_refspec_to_upgrade_on (gchar               **refspec_to_upgrade_on,
...
-    return FALSE;
-  if (remote == NULL)
+  /* Handle the ref from the commit's metadata */
+  if (ref_for_deployment != NULL)
     {
-      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
-                   "Invalid refspec ‘%s’ in commit: did not contain a remote name",
-                   refspec);
-      return FALSE;
+      OstreeCollectionRef *orig_collection_ref = collection_ref;
```
smspillaz: Instead of freeing this later you can take ownership with `g_autoptr`

src/eos-updater-poll-common.c
```diff
@@ -386,41 +391,37 @@ get_refspec_to_upgrade_on (gchar               **refspec_to_upgrade_on,
...
+
+      /* Replace the ref, refspec and collection_ref */
+      g_free (ref);
+      ref = g_steal_pointer (&ref_for_deployment);
+      g_free (refspec);
+      refspec = g_strdup_printf ("%s:%s", remote, ref);
+      collection_ref = ostree_collection_ref_new (orig_collection_ref->collection_id,
+                                                  ref);
+      ostree_collection_ref_free (orig_collection_ref);
```
smspillaz: Doing this free/replace dance has a strange smell to it, I have a suggestion for how this could be a little neater.

src/eos-updater-poll-common.c
```diff
@@ -386,41 +391,37 @@ get_refspec_to_upgrade_on (gchar               **refspec_to_upgrade_on,
...
-  if (!get_refspec_to_upgrade_on_from_deployment (sysroot,
-                                                  booted_deployment,
-                                                  &refspec,
-                                                  error))
+  if (!get_booted_refspec (booted_deployment,
+                           &refspec,
+                           &remote,
+                           &ref,
+                           &collection_ref,
+                           error))
```
smspillaz: I think these can be renamed to `booted_refspec` `booted_remote`, `booted_ref`, `booted_collection_ref` etc. Then below in the branch where we check for the checkpoint, we can set the outparams to the overridden refspec, remote, and booted ref. If that branch isn't followed, then set the outparams to the `booted_refspec` and friends. That should obviate the need for the free/replace dance below.

tests/test-update-refspec-checkpoint.c
```diff
@@ -62,7 +61,7 @@ insert_update_refspec_metadata_for_commit (guint         commit,
 
   g_hash_table_insert (*out_metadata_hashtable,
                        GUINT_TO_POINTER (commit),
-                       create_checkpoint_target_metadata (new_refspec));
+                       create_checkpoint_target_metadata (new_ref));
 }
```
smspillaz: I think we should retain a test here that it doesn't crash when you give it a refspec with a remote. That said, such a test may not work considering that it uses `g_warning` and I'm pretty sure that gtester sets `G_DEBUG=fatal-warnings`

smspillaz: Oh wait, you did that in the test below. Ignore this comment then.

tests/test-update-refspec-checkpoint.c
smspillaz: Yeah, strange - I'll check `master` to see if there's other places where this happens....

smspillaz:
Created a PR here with my suggested changes: https://github.com/endlessm/eos-updater/pull/197 . Other than that tested this locally and it works fine.

tests/test-update-refspec-checkpoint.c
dbnicholson: This was completely cargo culted from the first test function in that file, `test_update_refspec_checkpoint()`. I'll fixup any of those in that file.

src/eos-updater-poll-common.c
dbnicholson: Ah, good call. I always think of `g_autoptr` as being related to allocation rather than an automatic finalizer.

src/eos-updater-poll-common.c
dbnicholson: Sure does. I couldn't decide a good scheme that let me keep part of the original call.

src/eos-updater-poll-common.c
dbnicholson: Makes sense. Keeping all the out params common is what pushed me what I had. Let me try reworking with them branched where necessary.

dbnicholson:
If you don't mind, I'll squash your changes from #197 into my commits.

dbnicholson:
CR ready again. I squashed/merged the changes from #197 here.

PR #201, opened 2018-05-07
T15816 Split eos-updater into subdirectories

pwithnall:
https://phabricator.endlessm.com/T15816


configure.ac
```diff
@@ -100,22 +100,19 @@ PKG_CHECK_MODULES([OSTREE],
 PKG_CHECK_MODULES([SOUP],
                   [libsoup-2.4])
 
-PKG_CHECK_MODULES([EOS_AUTOUPDATER],
-                  [libnm >= $NM_REQUIRED_VERSION])
-
+AX_PKG_CHECK_MODULES([EOS_AUTOUPDATER],[glib-2.0 >= $GLIB_REQUIRED_VERSION gio-2.0 gobject-2.0 libnm >= $NM_REQUIRED_VERSION ostree-1 >= $OSTREE_REQUIRED_VERSION])
```
smspillaz: As far as I can tell the autoupdater does not depend on ostree

pwithnall: Currently, it kind of does, in that `libeos-updater-util/util.h` included `ostree.h` in its header — so anything which used `util.h` (including the autoupdater) needed to have the `CFLAGS` for OSTree, but not actually link to it.

I’ve pushed another commit which rearranges the OSTree APIs so this isn’t necessary.

pwithnall:
> In general it might be good to split out commits that move things around from commits that make substantive changes to dependencies and the build process.

Not really possible given that the `#include` lines need to have the right paths, and automake doesn’t arrange build dependencies properly if you move them into sibling directories of where the `Makefile.am` is. :-(  I think this is the best I can get it.

smspillaz:
> Not really possible given that the #include lines need to have the right paths, and automake doesn’t arrange build dependencies properly if you move them into sibling directories of where the Makefile.am is. :-( I think this is the best I can get it.

Fair enough. We'll also need to fix the conflicts here.

pwithnall:
> Fair enough. We'll also need to fix the conflicts here.

git managed it automatically; I’ve checked the diffs. :confetti_ball:

pwithnall:
Rebased and merged.

PR #202, opened 2018-05-07
Fix checkpoint with no collection ID

dbnicholson:
Hit a segfault when constructing the update info for a checkpoint when the remote had no collection ID.

https://phabricator.endlessm.com/T19437

### DIFF
--- a/src/eos-updater-poll-common.c
+++ b/src/eos-updater-poll-common.c
@@ -419,7 +419,7 @@ get_refspec_to_upgrade_on (gchar               **refspec_to_upgrade_on,
       if (remote_to_upgrade_on != NULL)
         *remote_to_upgrade_on = g_steal_pointer (&booted_remote);
       if (ref_to_upgrade_on != NULL)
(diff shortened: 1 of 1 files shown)


src/eos-updater-poll-common.c
```diff
@@ -409,7 +409,7 @@ get_refspec_to_upgrade_on (gchar               **refspec_to_upgrade_on,
   if (checkpoint_ref_for_deployment != NULL)
     {
       /* Set outparams from the checkpoint ref instead */
-      if (collection_ref_to_upgrade_on != NULL)
+      if (collection_ref_to_upgrade_on != NULL && booted_collection_ref != NULL)
```
pwithnall: If `(collection_ref_to_upgrade_on != NULL && booted_collection_ref == NULL)`, then this will not set the `*collection_ref_to_upgrade_on` out parameter, which the caller may then check. In that case, it should set it to `NULL`. This needs a follow-up PR.

PR #206, opened 2018-05-08
eos-updater-apply: Set remote branches option after deploy

dbnicholson:
This is mostly cosmetic since the only time the remote branches option
is used is when a pull is done with no refs specified. Still, people
expect it to match the ref in the origin refspec, and changing it to
match prevents inadvertently pulling an unused ref that would occupy
disk space. Errors from the option update are ignored since it has no
effect on the actual upgrade.

https://phabricator.endlessm.com/T22477

### DIFF
--- a/eos-updater/apply.c
+++ b/eos-updater/apply.c
@@ -97,6 +97,66 @@ get_test_osname (void)
 }
 
 static gboolean
+update_remote_branches (OstreeRepo   *repo,
+                        const gchar  *refspec,
+                        GError      **error)
+{
+  g_autofree gchar *remote = NULL;
+  g_autofree gchar *ref = NULL;
+  g_autofree gchar *cur_branches = NULL;
+  g_autofree gchar *new_branches = NULL;
+
+  if (!ostree_parse_refspec (refspec, &remote, &ref, error))
+    return FALSE;
+  if (remote == NULL)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                   "Invalid refspec ‘%s’ in origin: did not contain a remote name",
+                   refspec);
+      return FALSE;
+    }
+
+  /* Update the remote branches option to include the origin ref.
+   * Ideally this would operate on a list of branches and add the new
+   * branch, but to keep things simple just set the option to the single
+   * origin ref. */
+  if (!ostree_repo_get_remote_option (repo, remote, "branches", NULL,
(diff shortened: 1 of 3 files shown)


eos-updater/apply.c
```diff
@@ -96,6 +96,66 @@ get_test_osname (void)
...
+       * that would prevent the repo from opening again. */
+      if (!g_key_file_has_group (config, remote_group))
+        {
+          GFile *repo_file = ostree_repo_get_path (repo);
+          g_autofree gchar *repo_path = g_file_get_path (repo_file);
+
+          g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                       "Remote ‘%s’ does not exist in %s/config",
+                       remote, repo_path);
+          return FALSE;
```
smspillaz: Does it make sense to have this fail the entire update process? In this case shouldn't we update the one that's in `remotes.d`?

eos-updater/apply.c
```diff
@@ -96,6 +96,66 @@ get_test_osname (void)
...
+    }
+
+  /* Update the remote branches option to include the origin ref.
+   * Ideally this would operate on a list of branches and add the new
+   * branch, but to keep things simple just set the option to the single
+   * origin ref. */
+  if (!ostree_repo_get_remote_option (repo, remote, "branches", NULL,
+                                      &cur_branches, error))
+    return FALSE;
+  new_branches = g_strdup_printf ("%s;", ref);
```
smspillaz: Do we want to unconditionally overwrite all the old branches here?

pwithnall: Only listing a single branch makes sense to me: the user is only ever going to be on a single OSTree branch apart from when they’ve just upgraded from (say) eos3.3 to eos3.4 and hence have one deployment from each. Even then, I can only see them wanting to pull eos3.4 going forwards.

dbnicholson: FWIW, I worked out how to merge the lists. If you're still interested in that, I can push my branch out.

The use case for multiple branches is more like a developer that's doing something very specific. Still, since the only use for the `branches` option is an `ostree pull` with no specified refs, it's very unlikely that just putting the single branch is the wrong thing to do.

pwithnall: I don’t think it’s worth the churn, unless you think it’s worth it.

dbnicholson: Not really, but since I did the work I thought I'd offer :)

PR #208, opened 2018-05-22
Don't prune the repository when updating or uninstalling flatpaks

mariospr:
This is not a fix, but a workaround as an attempt to prevent taking too long
to boot when doing OS upgrades that imply updating a flatpak deployment.

https://phabricator.endlessm.com/T22621


mariospr:
Of course, CI failed because this PR depends on https://github.com/endlessm/flatpak/pull/95 being merged

eos-updater/fetch.c
```diff
@@ -468,7 +468,7 @@ static FlatpakUpdateFlags
 update_flags_for_action_flags (EuuFlatpakRemoteRefActionFlags action_flags)
 {
   return !(action_flags & EUU_FLATPAK_REMOTE_REF_ACTION_FLAG_IS_DEPENDENCY) ?
-      FLATPAK_UPDATE_FLAGS_NO_DEPLOY :
```
cosimoc: IIRC the logic here is that before reboot, we fully install/update all dependencies and do a NO_DEPLOY update for the actual ref we need to install during the next boot, where we will do a subsequent NO_PULL update to complete it. (https://github.com/endlessm/eos-updater/blob/master/libeos-updater-flatpak-installer/perform-flatpak-actions.c#L166 and https://github.com/endlessm/eos-updater/blob/master/libeos-updater-flatpak-installer/perform-flatpak-actions.c#L52)

In other words, I am pretty sure that this code path is only hit during the pre-reboot preparation, and you should instead add the NO_PRUNE options to other call sites mentioned above, since that's the code that runs at boot.
(That's also what makes me think that the upstream logic is buggy in the other PR.)

pwithnall: We talked about this in London today, and we concluded that the prune operations which flatpak does should be conditional on `!NO_PULL`, rather than anything to do with `NO_DEPLOY`. flatpak’s pruning is based on reachability from refs, and deploying/not deploying an app doesn’t affect its refs. Pulling/Not pulling an app does.

pwithnall: This likely needs fixing in other places in flatpak too.

pwithnall:
> This is not a fix, but a workaround as an attempt to prevent taking too long
> to boot when doing OS upgrades that imply updating a flatpak deployment.

I would say it counts as a legitimate improvement, not a workaround.

Remember to bump the flatpak dependency in `debian/control` so that we get the right APIs.

eos-updater/fetch.c
mariospr: Yes, I think you're right. But also, talking to @pwithnall he seems to be happy on not pruning on here either, so I think I'll keep setting a `NO_PRUNE` flag here anyway, regardless of whether it's not really run at boot time.

eos-updater/fetch.c
```diff
@@ -464,12 +464,13 @@ install_flags_for_action_flags (EuuFlatpakRemoteRefActionFlags action_flags)
       FLATPAK_INSTALL_FLAGS_NONE;
 }
 
-static FlatpakInstallFlags
+static FlatpakUpdateFlags
```
smspillaz: I am surprised that the compiler did not catch this earlier! Is there a warning level we can use to get strongly typed enums?

smspillaz:
> I would say it counts as a legitimate improvement, not a workaround.

Not sure I would want to keep this permanently. To me, pruning is a feature since we want to minimize the chance that the installer fails due to `ENOSPACE`. That said, it is doing whatever it is doing very inefficiently. Once we fix the performance of it upstream, this should be revisited (or at least, group all this stuff into a transaction and prune at the start and end)

eos-updater/fetch.c
mariospr: I was surprised too, really

pwithnall:
> Not sure I would want to keep this permanently. To me, pruning is a feature since we want to minimize the chance that the installer fails due to `ENOSPACE`. That said, it is doing whatever it is doing very inefficiently. Once we fix the performance of it upstream, this should be revisited (or at least, group all this stuff into a transaction and prune at the start and end)

Pruning always happens after the peak disk space usage has been reached. If installing/updating a particular app is going to give you `ENOSPACE`, no amount of pruning before/after the install/update operation is going to help. It can only help ensure that situation isn’t reached for future app installs/updates — and that’s already taken care of by the prune at the end of eos-updater’s `apply` phase.

pwithnall:
The [flatpak changes](https://github.com/endlessm/flatpak/pull/95) have been merged. Waiting for [OBS to rebuild](https://obs-master.endlessm-sf.com/package/show?package=flatpak&project=eos%3Amaster%3Aendless), then we can poke Jenkins here and merge.

pwithnall:
Jenkins, retest this please

PR #209, opened 2018-05-23
T22054: Logging improvements

smspillaz:
https://phabricator.endlessm.com/T22054

### DIFF
--- a/libeos-updater-util/flatpak.c
+++ b/libeos-updater-util/flatpak.c
@@ -921,10 +921,10 @@ read_flatpak_ref_actions_from_node (JsonNode      *node,
                                     EOS_UPDATER_ERROR,
                                     EOS_UPDATER_ERROR_UNKNOWN_ENTRY_IN_AUTOINSTALL_SPEC))
             {
-              g_debug ("%s while parsing %s. Skipping this action and it "
-                       "will not be reapplied later. System may be in an "
-                       "inconsistent state from this point forward.",
-                       local_error->message, filename);
+              g_warning ("%s while parsing %s. Skipping this action and it "
+                         "will not be reapplied later. System may be in an "
+                         "inconsistent state from this point forward.",
+                         local_error->message, filename);
               g_ptr_array_add (skipped_action_entries, json_node_to_string (element_node));
               g_clear_error (&local_error);
               continue;
@@ -1564,9 +1564,9 @@ list_all_remote_refs_in_flatpak_installation (FlatpakInstallation  *installation
        * what collection to install it from. */
       if (collection_id == NULL)
         {
-          g_debug ("Ignoring dependencies from remote %s as "
-                   "no collection-id set",
-                   remote_name);
+          g_warning ("Ignoring dependencies from remote %s as "
+                     "no collection-id set",
+                     remote_name);
           continue;
         }
 
@@ -1683,12 +1683,12 @@ fetch_runtime_ref_for_source_ref (FlatpakInstallation *installation,
       if (g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
         {
           g_autofree gchar *formatted_source_ref = flatpak_ref_format_ref (source_ref);
-          g_debug ("Remote %s did not have a summary cache "
-                   "or %s was not found in it, assuming that "
-                   "%s does not have a runtime",
-                   source_ref_remote_name,
-                   formatted_source_ref,
-                   formatted_source_ref);
+          g_warning ("Remote %s did not have a summary cache "
+                     "or %s was not found in it, assuming that "
+                     "%s does not have a runtime",
+                     source_ref_remote_name,
+                     formatted_source_ref,
+                     formatted_source_ref);
 
           *out_runtime_ref = NULL;
           return TRUE;
--- a/libeos-updater-util/tests/flatpak.c
+++ b/libeos-updater-util/tests/flatpak.c
@@ -27,6 +27,31 @@
 #include <libeos-updater-util/types.h>
 #include <locale.h>
 
+static guint n_warnings = 0;
+
+static GLogWriterOutput
+log_writer_cb (GLogLevelFlags   log_level,
+               const GLogField *fields,
+               gsize            n_fields,
+               gpointer         user_data)
+{
+  if (log_level & G_LOG_LEVEL_WARNING)
+    n_warnings++;
+  return G_LOG_WRITER_HANDLED;
+}
+
+static void
+reset_n_warnings (void)
+{
+  n_warnings = 0;
+}
+
+static void
+assert_n_warnings (guint expected_n_warnings)
+{
+  g_assert_cmpuint (n_warnings, ==, expected_n_warnings);
+}
+
 typedef struct
 {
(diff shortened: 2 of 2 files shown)


libeos-updater-util/flatpak.c
```diff
@@ -921,10 +921,10 @@ read_flatpak_ref_actions_from_node (JsonNode      *node,
                                     EOS_UPDATER_ERROR,
                                     EOS_UPDATER_ERROR_UNKNOWN_ENTRY_IN_AUTOINSTALL_SPEC))
             {
-              g_debug ("%s while parsing %s. Skipping this action and it "
-                       "will not be reapplied later. System may be in an "
-                       "inconsistent state from this point forward.",
-                       local_error->message, filename);
+              g_message ("%s while parsing %s. Skipping this action and it "
```
pwithnall: I’d go with `g_warning()`. Was there any discussion about the logging level in use for these messages in the initial PR?

libeos-updater-util/flatpak.c
```diff
@@ -1564,9 +1564,9 @@ list_all_remote_refs_in_flatpak_installation (FlatpakInstallation  *installation
        * what collection to install it from. */
       if (collection_id == NULL)
         {
-          g_debug ("Ignoring dependencies from remote %s as "
-                   "no collection-id set",
-                   remote_name);
+          g_message ("Ignoring dependencies from remote %s as "
```
pwithnall: Similarly, `g_warning()` here.

libeos-updater-util/flatpak.c
```diff
@@ -1683,12 +1683,12 @@ fetch_runtime_ref_for_source_ref (FlatpakInstallation *installation,
       if (g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
         {
           g_autofree gchar *formatted_source_ref = flatpak_ref_format_ref (source_ref);
-          g_debug ("Remote %s did not have a summary cache "
-                   "or %s was not found in it, assuming that "
-                   "%s does not have a runtime",
-                   source_ref_remote_name,
-                   formatted_source_ref,
-                   formatted_source_ref);
+          g_message ("Remote %s did not have a summary cache "
```
pwithnall: And here. Basically, I’d want these to be really obvious in the systemd journal. (And if we ever get automated QA, they could potentially be caught by a test which checks for warning (or higher) entries in the journal.)

smspillaz: I'm not really sure how I feel about warning. If we test this behaviour `g_warning` and tests don't mix :/

libeos-updater-util/flatpak.c
smspillaz: I'm not sure. I think `debug` was suggested somewhere and then it got copied.

smspillaz:
One thing I am still potentially concerned about is how a warning loglevel will behave in the presence of subprocesses where we dont call `g_log_set_writer_func`. AFAIK, `G_DEBUG` is still set to `fatal-warnings` on those so the subprocess would still crash if it tried to emit a warning. I have not been able to test this however, so I could be wrong

pwithnall:
> Optional: Could pass a pointer to a fixture here (by calling this in the test setup func) instead of relying on global state.

Nope, since `g_log_set_writer_func()` should only be called once per process (see [its documentation](https://developer.gnome.org/glib/stable/glib-Message-Logging.html#g-log-set-writer-func)).

> I would call reset_n_wanings as part of a fixture teardown hook just in case a future test triggers a warning (which will be suppressed above), causing later tests to fail due to potentially stale global state.

No other tests can fail due to warning messages, since no other tests call `assert_n_warnings()`. However, this does mean that we will silently ignore warnings from other tests, which is not good (that might be what you were getting at). I’ll update the branch to improve our behaviour there.

> One thing I am still potentially concerned about is how a warning loglevel will behave in the presence of subprocesses where we dont call `g_log_set_writer_func`.

The tests have all passed, so I don’ŧ think there’s a problem there.

> AFAIK, `G_DEBUG` is still set to `fatal-warnings` on those so the subprocess would still crash if it tried to emit a warning. I have not been able to test this however, so I could be wrong

Actually, `G_DEBUG` is only set to `gc-friendly` in the tests (see `glib-tap.mk`). We should be setting it to `fatal-warnings` though, definitely. I’ll handle that as a separate PR.

smspillaz:
Optional: Could pass a pointer to a fixture here (by calling this in the test setup func) instead of relying on global state.

    Nope, since g_log_set_writer_func() should only be called once per process (see its documentation).

        I would call reset_n_wanings as part of a fixture teardown hook just in case a future test triggers a warning (which will be suppressed above), causing later tests to fail due to potentially stale global state.

    No other tests can fail due to warning messages, since no other tests call assert_n_warnings(). However, this does mean that we will silently ignore warnings from other tests, which is not good (that might be what you were getting at). I’ll update the branch to improve our behaviour there.


Ah, that sucks about only being able to set the callback once.

The case I'm thinking of is more along the lines of:

```
happy_test() { /* something that warns unexpectedly, reset_count not called afterwards, warning not detected */ }
test_that_expect_not_to_warn() { /* asserts that there were no warnings 💥 */ }
```

Basically its a classic potential leakage of global state into other tests issue. I'd like to see us avoid the global state or at least always clean it up after each test.

> Actually, G_DEBUG is only set to gc-friendly in the tests (see glib-tap.mk). We should be setting it to fatal-warnings though, definitely. I’ll handle that as a separate PR.

Given the discussion we just had, I'm not sure if we should be doing that? I'm thinking of https://github.com/ostreedev/ostree/pull/1542 for example.

If we ever want to test what the behaviour is during a case that warns in an acceptance test, the child process has to not warn (and thus terminate unexpectedly).

smspillaz:
> Shouldn’t review my own code!

Since I proposed this PR, I cant review it either 🙃

pwithnall:
> Ah, that sucks about only being able to set the callback once.

It’s a deliberate way of how the logging infrastructure is designed. I don’t think it’s a problem. (We do want the same log writer function for each test after all.) We just need to make sure any global state touched by that log writer function is appropriately reset.

> Basically its a classic potential leakage of global state into other tests issue. I'd like to see us avoid the global state or at least always clean it up after each test.

The warning count does get reset at the start of each test which later asserts on the warning count. We could do with an assertion before exiting which checks there were no other warnings though. I’d rather not introduce a test fixture just to assert on the warning count; that‘s quite an invasive change.

> Given the discussion we just had, I'm not sure if we should be doing that? I'm thinking of ostreedev/ostree#1542 for example.
> 
> If we ever want to test what the behaviour is during a case that warns in an acceptance test, the child process has to not warn (and thus terminate unexpectedly).

I’ve filed https://phabricator.endlessm.com/T22734 about it. I think we should use `G_DEBUG=fatal-warnings`. If it becomes a problem for subprocesses, we can remove it from the subprocess environment and do some checks on the `stderr` coming from that subprocess instead.

pwithnall:
Updated and ready for review. You could just review it in a comment, @smspillaz.

smspillaz:
Github wont let me review this but after our discussion I am fine with this for now

pwithnall:
Thanks.

PR #212, opened 2018-06-06
T22763 Prune before fetch

pwithnall:
https://phabricator.endlessm.com/T22763

This should **not** be merged, as it has not been tested yet. Let’s see what Jenkins says, and we probably want to do some manual testing in a VM too before merging. Ready for code review though.


pwithnall:
Closed (without prejudice) as per https://phabricator.endlessm.com/T22763#598482. We might want to look at this again in future, but only as part of a larger, more coherent plan for making updates succeed on low disk space systems.

PR #214, opened 2018-06-08
T22805: eos-updater: Allow upgrades even if the current commit is missing

pwithnall:
https://phabricator.endlessm.com/T22805

Rebased and ready for full review and merging.

### DIFF
--- a/tests/test-update-missing-deployed-commit.c
+++ b/tests/test-update-missing-deployed-commit.c
@@ -1,0 +1,89 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*-
+ *
+ * Copyright © 2018 Endless Mobile, Inc.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * Authors:
+ *  - Philip Withnall <withnall@endlessm.com>
+ */
+
+#include <gio/gio.h>
+#include <libeos-updater-util/util.h>
+#include <locale.h>
+#include <test-common/convenience.h>
+
+
+/* Delete the commit object representing the currently deployed commit, and try
+ * to do an update. The update should succeed (but should warn about the missing
+ * commit object). */
+static void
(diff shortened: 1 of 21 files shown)


tests/test-update-missing-deployed-commit.c
```diff
@@ -0,0 +1,89 @@
...
+{
+  g_auto(EtcData) real_data = { NULL, };
+  EtcData *data = &real_data;
+
+  g_test_bug ("T22805");
+
+  /* We could get OSTree working by setting OSTREE_BOOTID, but shortly
+   * afterwards we hit unsupported syscalls in qemu-user when running in an
+   * ARM chroot (for example), so just bail. */
+  if (!eos_test_has_ostree_boot_id ())
```
wjt: This (existing) function is misleadingly-named. I expected it would mean "is this test running on an OSTree-booted system?", and my follow-up question would have been: given that neither development, nor Jenkins runs, nor OBS builds happen on unconverted systems, when would this test run?

But in fact it refers to `/proc/sys/kernel/random/boot_id`. I happened to have read [Lennart's post “On IDs”](http://0pointer.de/blog/projects/ids.html) earlier today and it has this to say (emphasis mine):

> `/proc/sys/kernel/random/boot_id`: A random ID that is regenerated on each boot. As such it can be used to identify the local machine's current boot. It's **universally available** on any recent Linux kernel. It's a good and safe choice if you need to identify a specific boot on a specific booted kernel.

So instead my question is: in which environment we care about is this not present?

wjt: And a follow-up: this block (and comment) is copy-pasted in many places. `skip_test_on_ostree_boot_id` is defined in `test-update-direct.c` and then each test in that file uses:

```
if (skip_test_on_ostree_boot_id ())
  return;
```

Maybe it could be hoisted into `test-common/utils` and the rest of the copy-pasta removed? 🍝🍝🍝

pwithnall: > So instead my question is: in which environment we care about is this not present?

An ARM chroot where you can’t reasonably bind `/proc` into the chroot (for reasons I can’t remember; probably because you don’t want to expose anything from the host’s `/proc`). It was something I was using last year to debug an insidious issue with the ARM eos-updater code, which required running the unit tests in such a chroot.

See commit 5032d0a879bb5b220a52c3cc40dcf75260d64fca.

> Maybe it could be hoisted into test-common/utils and the rest of the copy-pasta removed? spaghettispaghettispaghetti

Yes, 100%. I’ll get out my spaghetti wrangling tools.

pwithnall: Fix pushed.

dbnicholson: I could be wrong, but I'm pretty sure both `jenkins-obs` and `obs-build` mount `/proc` in the buildroots. Many things are broken without it. What I suspect happened is that the older kernels that the ARM builders were running didn't have `/proc/sys/kernel/random/boot_id`. They're all running linux 4.4 from xenial and it looks like they all have `boot_id` present. So, I'd guess that this isn't an issue anymore, but it's still obviously safe to ensure it's present before running the test.

pwithnall: IIRC it was never anything to do with the Jenkins or OBS builders, it was for running an ARM chroot locally on my machine ([as per the wiki](https://phabricator.endlessm.com/w/howto/development_chroot/)).

PR #215, opened 2018-06-08
eupv: Update the summary file before calling create-usb

mwleeds:
The `ostree create-usb` command requires an updated summary file to
function properly; see its man page.

https://phabricator.endlessm.com/T22895

### DIFF
--- a/eos-updater-prepare-volume/eos-updater-prepare-volume
+++ b/eos-updater-prepare-volume/eos-updater-prepare-volume
@@ -257,15 +257,15 @@ class VolumePreparer:
         # Eliminate duplicates.
         collection_refs = list(set(collection_refs))
 
-        # Pass the heavy lifting off to `ostree create-usb`.
+        # Pass the heavy lifting off to `ostree create-usb`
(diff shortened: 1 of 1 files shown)


eos-updater-prepare-volume/eos-updater-prepare-volume
```diff
@@ -257,11 +257,13 @@ class VolumePreparer:
...
 
-        # Pass the heavy lifting off to `ostree create-usb`.
+        # Pass the heavy lifting off to `ostree create-usb`
+        # which requires an updated summary file
         # TODO: Verify that it adds GPG keys where appropriate.
-        # TODO: Verify that a summary file is present.
         flattened_collection_refs = [e for l in collection_refs for e in l]
         _, repo = self.sysroot.get_repo()
+        self.__run(['ostree', 'summary', '--update',
+                    '--repo', os.path.realpath(repo.get_path().get_path())])
```
pwithnall: Might want to factor out the repeated `os.path.realpath(repo.get_path().get_path())` from this line and the one below.

PR #217, opened 2018-06-12
poll: Fix an assertion failure on the fallback code path

mwleeds:
In metadata_fetch_internal(), there are two code paths for getting
metadata, a "new" one that supports pulling from peer sources, and a
fallback one that only fetches from the Internet. On the fallback code
path, we are passing to run_fetchers() only one fetcher but the complete
configured list of sources. Since the fallback path is supposed to only
be used for fetching from the Internet, we should only pass that source
to run_fetchers(). This commit does that, which avoids an assertion
failure from the line "g_return_val_if_fail (fetchers->len ==
sources->len, NULL);".


pwithnall:
Cherry picked to eos3.4 and eos3.3. The eos3.4 build [failed](https://ci.endlessm-sf.com/job/obs-eos-updater-eos3.4/13/), but it seems to be unrelated. I’m investigating.

pwithnall:
Seems it was a transient failure: retrying the eos3.4 build succeeded. The eos3.3 build succeeded first time round.

PR #219, opened 2018-06-13
T22192

mwleeds:


eos-updater/poll-common.c
```diff
@@ -551,6 +551,7 @@ fetch_latest_commit (OstreeRepo *repo,
       g_clear_pointer (&ref, g_free);
       g_clear_pointer (&new_refspec, g_free);
       g_clear_pointer (&checksum, g_free);
+      g_clear_pointer (out_version, g_free);
```
pwithnall: It’s a bit suspect to clear the `out_version` (what if the caller hadn’t initialised it to `NULL` before calling `fetch_latest_commit()`? — they don’t have to). It would be better to assign to a local `version` variable inside the loop, then steal that to `out_version` after the loop, just like is done for `out_checksum` and `out_new_refspec`.

mwleeds:
>Review of “poll: Don't leak the OstreeRepoFinderResult array” — Good catch. Did you find this with valgrind, or via inspection?

I didn't use valgrind; I just noticed it.

eos-updater/poll-common.c
mwleeds: Pushed a fixup

mwleeds:
The test that's failing on the CI server isn't failing on my machine :/

pwithnall:
The direct failure is a bit of a red herring; the test only fails because `eos-updater` crashes:
```
# eos-updater-tests-DEBUG: Trying to cancel state None
# eos-updater-tests-DEBUG: Error cancelling None: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name com.endlessm.Updater was not provided by any .service files
# eos-updater-tests-DEBUG: State changed None
# eos-updater-tests:ERROR:../../../tests/test-update-direct.c:278:test_cancel_update: 'timed_out' should be FALSE
```

The crash is higher up in [the log](https://ci.endlessm-sf.com/job/obs-eos-updater-pr/285/console):
```
eos-updater-dbus-Message: Changing to error state: com.endlessm.Updater.Error.Cancelled, 19, Operation was cancelled
eos-updater-dbus-Message: Clearing error state and changing to state Polling
eos-updater-Message: Using product branch REF

(/build/src/eos-updater_1.3.0~rc3+dev1097.37bf036/_build/sub/eos-updater/.libs/eos-updater:7990): OSTree-CRITICAL **: ostree_repo_prepare_transaction: assertion 'self->in_transaction == FALSE' failed
```

This looks very similar to something I [fixed recently in OSTree](https://github.com/ostreedev/ostree/commit/2d2f218669e6230edaf6329913a2b2c1cda590ed). The fix was [merged into endlessm/ostree.git](https://github.com/endlessm/ostree/commit/d77e69cc079cc266d55389cc6be4b2e132c75071) 20 hours ago, and this Jenkins run was 8 hours ago, so it’s likely the fix is incomplete, erroneous, or perhaps not actually relevant.

The fact that the test cancels various eos-updater states and then tries again means that the likely mechanism for hitting the OSTree assertion failure is that a pull transaction was started, cancelled, and then another transaction was attempted to be started (but `self->in_transaction` was still `TRUE`). So I’d be looking for places in the OSTree code where it could be possible to cancel a pull but (accidentally) not abort the transaction as well.

mwleeds:
Indeed, the build log says it's using commit d77e69c of libostree which is the latest commit.

mwleeds:
So it appears there's some non-determinism in that unit test failure because it succeeded this time after I added a debug statement. That also means the failure is not necessarily unique to this branch.

pwithnall:
> So it appears there's some non-determinism in that unit test failure because it succeeded this time after I added a debug statement. That also means the failure is not necessarily unique to this branch.

Yeah, I’ve just managed to reproduce the failure locally on master (without this branch). I’ve filed https://phabricator.endlessm.com/T23091 about it and will take a look.

I’ll review this branch soon, and will ignore the test failure for it.

tests/test-update-refspec-endoflife.c
```diff
@@ -0,0 +1,243 @@
...
+                                      &error);
+  g_assert_no_error (error);
+  g_assert_cmpuint (server->subservers->len, ==, 1u);
+
+  subserver = g_object_ref (EOS_TEST_SUBSERVER (g_ptr_array_index (server->subservers, 0)));
+  client_root = g_file_get_child (fixture->tmpdir, "client");
+  client = eos_test_client_new (client_root,
+                                default_remote_name,
+                                subserver,
+                                default_collection_ref,
```
pwithnall: Can you also please do a version of this test without collection IDs in the configuration? I don’t think we currently have any coverage for `ostree.endoflife-rebase`.

tests/test-update-refspec-endoflife.c
```diff
@@ -0,0 +1,243 @@
...
+
+const OstreeCollectionRef _default_collection_ref_no_id = { NULL, (gchar *) "REF" };
+const OstreeCollectionRef *default_collection_ref_no_id = &_default_collection_ref_no_id;
+
+static GHashTable *
+create_eol_rebase_metadata (const gchar *ref_to_upgrade)
+{
+  GHashTable *ht = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
+  g_hash_table_insert (ht,
+                       g_strdup ("ostree.endoflife-rebase"),
```
pwithnall: You can use `OSTREE_COMMIT_META_KEY_ENDOFLIFE_REBASE` for this.

eos-updater/poll.c
```diff
@@ -568,109 +558,23 @@ metadata_fetch_new (OstreeRepo    *repo,
...
-          g_clear_pointer (&upgrade_collection_ref, ostree_collection_ref_free);
-          upgrade_collection_ref = g_steal_pointer (&new_collection_ref);
-        }
+      g_message ("Poll: Couldn’t find any updates");
+      return NULL;
     }
-  while (redirect_followed);
-
-  /* Final checks on the commit we found. */
-  if (!is_checksum_an_update (repo,
```
pwithnall: So the `is_checksum_an_update()` checks are done as part of the call to `check_for_update_following_checkpoint_if_allowed()`, right?

eos-updater/poll-common.c
```diff
@@ -540,40 +555,118 @@ fetch_latest_commit (OstreeRepo *repo,
...
+                return FALSE;
+            }
+
+          g_clear_pointer (&checksum, g_free);
+          g_clear_pointer (&version, g_free);
+          g_clear_pointer (&new_refspec, g_free);
+          g_clear_pointer (&new_collection_ref, ostree_collection_ref_free);
+
+          /* Parse the commit and check there’s no redirection to a new ref. */
+          if (!parse_latest_commit (repo, upgrade_refspec, &redirect_followed,
```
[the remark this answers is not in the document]
mwleeds: good point

eos-updater/poll.c
mwleeds: Right. In the case that no update was available you'll have `update_ref_info.commit == NULL` after the call to `check_for_update_following_checkpoint_if_allowed`.

tests/test-update-refspec-endoflife.c
mwleeds: :+1:

tests/test-update-refspec-endoflife.c
mwleeds: done (on the client only because the server will forever have a collection ID set)

pwithnall:
Looks good to merge once the fixups have been auto-squashed into the branch.

PR #224, opened 2018-07-23
eupv: Port to using flatpak create-usb

mwleeds:
Now that we have a `flatpak create-usb` tool that takes care of copying
flatpaks to USBs along with their dependencies, related refs, repo
metadata, etc. there's no reason for eos-updater-prepare-volume to try
to emulate that work. So this commit changes the syntax accepted by
e-u-p-v so that it's just flatpak refs (flatpak will figure out the
collection IDs on its own) and calls out to `flatpak create-usb` after
using `ostree create-usb` to copy the OS ref.

https://phabricator.endlessm.com/T23118

### DIFF
--- a/eos-updater-prepare-volume/eos-updater-prepare-volume
+++ b/eos-updater-prepare-volume/eos-updater-prepare-volume
@@ -190,8 +190,9 @@ class VolumePreparer:
         self.__run(['ostree', 'summary', '--update', '--repo', repo_path])
         self.__run(['ostree', 'create-usb', '--repo', repo_path,
                     self.volume_path, os_collection_ref[0], os_collection_ref[1]])
-        self.__run(['flatpak', '--system', 'create-usb',
-                    self.volume_path] + all_flatpak_refs)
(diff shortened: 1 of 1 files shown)


mwleeds:
I haven't been able to test this yet because of the build error:
```
./configure: line 12890: syntax error near unexpected token `git-directory'
./configure: line 12890: `AX_IS_RELEASE(git-directory)'
```

mwleeds:
> I haven't been able to test this yet because of the build error:
>
>./configure: line 12890: syntax error near unexpected token `git-directory'
>./configure: line 12890: `AX_IS_RELEASE(git-directory)'

For the record this was fixed by installing `autoconf-archive`. Now this is blocked on the broken eos-updater build which is preventing testing this (fixed in #225).

treitter:
@mwleeds based on the Jenkins [output](https://ci.endlessm-sf.com/job/obs-eos-updater-pr/301/console), it looks like two test cases need to be updated to pass the ref(s) to `flatpak create-usb`. Currently, only `--system` and the mount point is being passed.

PR #225, opened 2018-07-24
test-common: Use the flatpak --asumeyes option

mwleeds:
Recent verisons of flatpak prompt the user for confirmation during
install/update/uninstall operations. So when the eos-updater unit tests
try to install the test runtime, there's an "Is this ok? [y/n]:"
prompt, and no is assumed since there's no user input, which eventually
leads to a unit test failure.

So this commit adds the --asumeyes option when the tests call out to
`flatpak install` or `flatpak uninstall`, which is meant for the purpose
of scripting and answers yes to questions (or the default option if it's
multiple choice).

https://phabricator.endlessm.com/T23138

### DIFF
--- a/test-common/flatpak-spawn.c
+++ b/test-common/flatpak-spawn.c
@@ -91,6 +91,7 @@ flatpak_install (GFile        *updater_dir,
       { NULL, FLATPAK_BINARY },
       { NULL, "install" },
       { "user", NULL },
+      { "assumeyes", NULL },
       { NULL, remote },
       { NULL, app_id },
       { NULL, NULL }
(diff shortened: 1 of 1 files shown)


test-common/flatpak-spawn.c
```diff
@@ -91,6 +91,7 @@ flatpak_install (GFile        *updater_dir,
       { NULL, FLATPAK_BINARY },
       { NULL, "install" },
       { "user", NULL },
+      { "assumeyes", NULL },
```
wjt: In the commit message this flag is typoed.

wjt: In what release of Flatpak was this flag added? Do we need to bump a build/test-time dependency?

wjt: Ah, I see 44db804b58b1a5471e7099357ed6f424486c0f61 already added a dependency on the very newest version.

mwleeds:
Seems like Will reviewed this so I'll merge it

PR #226, opened 2018-07-25
T23347: Drop old refspec and prune after following checkpoint

smspillaz:
https://phabricator.endlessm.com/T23385

### DIFF
--- a/eos-updater/apply.c
+++ b/eos-updater/apply.c
@@ -156,6 +156,35 @@ update_remote_branches (OstreeRepo   *repo,
   return TRUE;
 }
 
+static void
+prune_repo_and_report_statistics (OstreeRepo           *repo,
+                                  OstreeRepoPruneFlags  flags,
+                                  GCancellable         *cancellable)
+{
+  gint objects_total;
+  gint objects_pruned;
+  guint64 pruned_size_total;
+  g_autoptr(GError) local_error = NULL;
+
+  if (!ostree_repo_prune (repo,
+                          flags,
+                          -1,
+                          &objects_total,
+                          &objects_pruned,
+                          &pruned_size_total,
+                          cancellable,
+                          &local_error))
+    {
+      g_warning ("Failed to prune ostree repo: %s", local_error->message);
+      return;
+    }
+
+  g_message ("Pruned %d objects, %llu bytes total. %d remain",
+             objects_pruned,
+             pruned_size_total,
+             objects_total);
+}
+
 static gboolean
 apply_internal (ApplyData     *apply_data,
                 gboolean      *out_bootversion_changed,
@@ -246,6 +275,12 @@ apply_internal (ApplyData     *apply_data,
     g_warning ("Failed to clean up the sysroot after successful deployment: %s",
                local_error->message);
   g_clear_error (&local_error);
+
+  /* Prune any unused objects in the repo. This is a necessary step
+   * after removing old refs. */
+  prune_repo_and_report_statistics (repo,
+                                    OSTREE_REPO_PRUNE_FLAGS_REFS_ONLY,
+                                    cancellable);
 
   /* Try to update the remote branches option to use the new refspec.
    * This option is almost never used and has no impact on future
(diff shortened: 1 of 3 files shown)


eos-updater/poll.c
```diff
@@ -456,13 +474,22 @@ check_for_update_following_checkpoint_commits (OstreeRepo     *repo,
                               error))
     return FALSE;
 
-  out_update_ref_info->refspec = g_steal_pointer (&upgrade_refspec);
+  /* The "refspec" member is the *currently booted* refspec
+   * which may get cleaned up later if we change away from it. */
+  out_update_ref_info->refspec = g_steal_pointer (&booted_refspec);
+
+  /* The "remote", "ref" and "collection_ref" refer here to the
+   * ref and remote that we should be following given checkpoints. */
```
smspillaz: I'm not super happy with this, but I basically took the approach of "it doesn't seem to be broken and I don't know if anything depends on this, so I'm not touching it". Ideally `ref` and `remote` would be the booted ref and remote.

ramcq:
I've had a quick look, I don't quite understand how the test is proving that the old ref gets pruned. Might just be being thick though. I think the prune commit can be dropped because ostree_sysroot_cleanup calls a prune for us.

eos-updater/poll.c
```diff
@@ -470,7 +474,9 @@ check_for_update_following_checkpoint_commits (OstreeRepo     *repo,
                               error))
     return FALSE;
 
-  out_update_ref_info->refspec = g_steal_pointer (&upgrade_refspec);
+  /* The "refspec" member is the *currently booted* refspec
+   * which may get cleaned up later if we change away from it. */
+  out_update_ref_info->refspec = g_steal_pointer (&booted_refspec);
```
dbnicholson: So, is this the actual bug fix? Because `update_ref_info.refspec` gets passed to `eos_update_info_new` as the `old_refspec` argument?

eos-updater/apply.c
```diff
@@ -247,6 +276,12 @@ apply_internal (ApplyData     *apply_data,
                local_error->message);
   g_clear_error (&local_error);
 
+  /* Prune any unused objects in the repo. This is a necessary step
+   * after removing old refs. */
+  prune_repo_and_report_statistics (repo,
+                                    OSTREE_REPO_PRUNE_FLAGS_REFS_ONLY,
+                                    cancellable);
```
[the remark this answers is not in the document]
smspillaz: Yeah, I had this since there didn't seem to be anything about pruning in the documentation for `ostree_sysroot_cleanup`, but checking here, it does a prune as you suggest: https://github.com/resin-io/ostree/blob/master/src/libostree/ostree-sysroot-cleanup.c#L474 . I'll dro the commit.

eos-updater/poll.c
smspillaz: Yup, that's the fix. Essentially, we weren't cleaning up the booted refspec, but rather the refspec we initially considered for pulling (eg, we were measuring the difference between the refspec to initially pull vs eol-rebase refspecs, not the booted refspec vs the refspec that actually got pulled).

smspillaz:
@ramcq you are probably most interested in https://github.com/endlessm/eos-updater/pull/226/files#diff-0140425ccfb636a439827974f267d1c8R358 

    +  ostree_repo_list_refs (repo, NULL, &refs, NULL, &error);
    +  g_assert_no_error (error);
    +
    +  has_original_ref = g_hash_table_contains (refs, "REMOTE:REF");
    +  has_new_ref = g_hash_table_contains (refs, "REMOTE:REFv2");
    +
    +  g_assert_true (has_new_ref);
    +  g_assert_false (has_original_ref);

`REMOTE:REF` is the booted refspec, `REMOTE:REFv2` is the checkpoint refspec. After updating from the checkpoint, we should only retain the latter, not the former.

smspillaz:
Considering that I pushed this to the wrong branch, continuing in https://github.com/endlessm/eos-updater/pull/227

eos-updater/apply.c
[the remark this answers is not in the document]
dbnicholson: Yeah, you are definitely right that the documentation is not clear on that.

PR #227, opened 2018-07-27
T23385: Drop old refspec and prune after following checkpoint

smspillaz:
https://phabricator.endlessm.com/T23385


smspillaz:
This is the continuation of #226 , which was pushed to the wrong branch.

PR #230, opened 2018-07-30
Follow-up from #229

mwleeds:


pwithnall:
> if you try to pass that to the second function that expects a GHashTable you get an error

What was the error? IIRC pygobject should always be able to transparently convert a `dict` to a `GHashTable`. Is it a problem with the boxed types within the hash table?

mwleeds:
>What was the error?

`ValueError: Item 0: Pointer arguments are restricted to integers, capsules, and None. See: https://bugzilla.gnome.org/show_bug.cgi?id=683599`

pwithnall:
I guess it’s because of the nested generics in `@ref_actions_table: (element-type filename GPtrArray<EuuFlatpakRemoteRefAction>)` :'-(

Guess we have to go with the wrapper API. Thanks for sorting this out.

PR #234, opened 2018-09-18
apply: Add a collection ID to the OS remote config

mwleeds:
This commit makes eos-updater add a collection ID to the OS remote
config during the apply stage, in a similar way that the "branches"
config key is currently added. This has the effect of enabling LAN/USB
OS updates when the eos-updater config file has those sources enabled.

After the next checkpoint this patch can be dropped, because at that
point every user's config will have been updated.

https://phabricator.endlessm.com/T19363

### DIFF
--- a/eos-updater/apply.c
+++ b/eos-updater/apply.c
@@ -97,6 +97,82 @@ get_test_osname (void)
 }
 
 static gboolean
+update_remote_collection_id (OstreeRepo   *repo,
+                             const gchar  *refspec,
+                             GError      **error)
+{
+  g_autofree gchar *remote = NULL;
+  g_autofree gchar *ref = NULL;
+  g_autofree gchar *cur_collection_id = NULL;
+  g_autofree gchar *checksum = NULL;
+  g_autofree gchar *remote_group = NULL;
+  g_autoptr(GVariant) commit = NULL;
+  g_autoptr(GVariant) metadata = NULL;
+  g_autoptr(GKeyFile) config = NULL;
+  const char *collection_id;
+
+  if (!ostree_parse_refspec (refspec, &remote, &ref, error))
+    return FALSE;
+  if (remote == NULL)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                   "Invalid refspec ‘%s’ in origin: did not contain a remote name",
+                   refspec);
+      return FALSE;
+    }
+
+  /* If the remote already has a collection ID set, there's nothing to do */
+  if (!ostree_repo_get_remote_option (repo, remote, "collection-id", NULL,
+                                      &cur_collection_id, error))
+    return FALSE;
+  if (cur_collection_id != NULL && *cur_collection_id != '\0')
+    return TRUE;
+
+  if (!ostree_repo_resolve_rev (repo, refspec, FALSE, &checksum, error))
+    return FALSE;
+  if (!ostree_repo_load_variant (repo,
+                                 OSTREE_OBJECT_TYPE_COMMIT,
+                                 checksum,
+                                 &commit,
+                                 error))
+    return FALSE;
+
+  metadata = g_variant_get_child_value (commit, 0);
+  if (!g_variant_lookup (metadata, OSTREE_COMMIT_META_KEY_COLLECTION_BINDING, "&s", &collection_id))
+    {
+      g_debug ("No ostree.collection-binding metadata key on commit; not setting a collection ID locally");
+      return TRUE;
+    }
+
+  /* FIXME: Update the config file directly. If ostree ever gains
+   * sane remote modification, use that. */
+  config = ostree_repo_copy_config (repo);
+  remote_group = g_strdup_printf ("remote \"%s\"", remote);
+
+  /* Make sure that the remote group exists in the config file and
+   * the remote isn't defined in a remotes.d conf file. That
+   * shouldn't happen, but adding the group (which
+   * g_key_file_set_value() will do) would create a duplicate remote
+   * that would prevent the repo from opening again. */
+  if (!g_key_file_has_group (config, remote_group))
+    {
+      GFile *repo_file = ostree_repo_get_path (repo);
+      g_autofree gchar *repo_path = g_file_get_path (repo_file);
+
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                   "Remote ‘%s’ does not exist in %s/config",
+                   remote, repo_path);
+      return FALSE;
+    }
+
+  g_key_file_set_value (config, remote_group, "collection-id", collection_id);
+  if (!ostree_repo_write_config (repo, config, error))
+    return FALSE;
+
+  return TRUE;
+}
+
+static gboolean
 update_remote_branches (OstreeRepo   *repo,
                         const gchar  *refspec,
                         GError      **error)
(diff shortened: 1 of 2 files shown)


mwleeds:
Hmm, just realized this will add a collection ID in the update after 3.4.8, e.g. 3.5.0, because the updater doing the 3.4.8 update doesn't have this patch.

eos-updater/apply.c
```diff
@@ -96,6 +96,85 @@ get_test_osname (void)
...
+   * sane remote modification, use that. */
+  config = ostree_repo_copy_config (repo);
+  remote_group = g_strdup_printf ("remote \"%s\"", remote);
+
+  /* Make sure that the remote group exists in the config file and
+   * the remote isn't defined in a remotes.d conf file. That
+   * shouldn't happen, but adding the group (which
+   * g_key_file_set_value() will do) would create a duplicate remote
+   * that would prevent the repo from opening again. */
+  if (!g_key_file_has_group (config, remote_group))
```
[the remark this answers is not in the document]
dbnicholson: The problem is finding which file has that definition. That's where the lack of reasonable OSTree API comes in. Almost certainly our OS remote is in the main config file, though.

eos-updater/apply.c
```diff
@@ -96,6 +96,85 @@ get_test_osname (void)
...
+                   refspec);
+      return FALSE;
+    }
+
+  /* If the remote already has a collection ID set, there's nothing to do */
+  if (!ostree_repo_get_remote_option (repo, remote, "collection-id", NULL,
+                                      &cur_collection_id, error))
+    return FALSE;
+  if (cur_collection_id != NULL && *cur_collection_id != '\0')
+    return TRUE;
```
dbnicholson: If someone switches between repo stages, then this might not be accurate anymore.

eos-updater/apply.c
```diff
@@ -96,6 +96,85 @@ get_test_osname (void)
...
+    return FALSE;
+
+  metadata = g_variant_get_child_value (commit, 0);
+  collection_id_v = g_variant_lookup_value (metadata, "ostree.collection-binding", G_VARIANT_TYPE_STRING);
+  if (collection_id_v == NULL ||
+      (collection_id = g_variant_get_string (collection_id_v, NULL)) == NULL)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED, "Failed to read ostree.collection-binding metadata key");
+      return FALSE;
+    }
```
dbnicholson: Would it be better to look at the collection ID in the summary file (since that will definitely match what the remote specifies) rather than the binding in the commit?

eos-updater/apply.c
```diff
@@ -255,6 +334,15 @@ apply_internal (ApplyData     *apply_data,
                local_error->message);
   g_clear_error (&local_error);
 
+  /* Try to add a collection ID to the OS remote, which enables LAN/USB OS
+   * updates when combined with eos-updater.conf changes. Don't treat any
+   * errors as fatal to the update. */
+  /*FIXME: Remove this function call after the next checxpoint */
+  if (!update_remote_collection_id (repo, update_refspec, &local_error))
```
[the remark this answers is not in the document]
mwleeds: I don't think so, because keeping them separate means we can easily drop this after the next checkpoint, and the performance difference between the two implementations is probably paltry.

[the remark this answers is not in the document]
mwleeds: Not that it's hard to revert a patch, just that in terms of readable you can clearly tell this way that the two operations are separate

eos-updater/apply.c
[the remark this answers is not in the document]
mwleeds: Yeah, I don't think it's worth adding code to support an unsupported config which we're not doing in `update_remote_branches()`

eos-updater/apply.c
mwleeds: When you switch changes, `eos-stage-ostree` changes the collection ID for you. I don't think that's eos-updater's job

eos-updater/apply.c
mwleeds: How could the one in the commit not be the correct one? I don't think we can fetch the summary here in the apply stage, which I think is supposed to work offline

mwleeds:
> Hmm, just realized this will add a collection ID in the update after 3.4.8, e.g. 3.5.0, because the updater doing the 3.4.8 update doesn't have this patch.

But I don't think this is a reason not to merge this patch. Enabling USB OS updates for reflashes in one release and for upgrade users in the next means they'll get more testing

eos-updater/apply.c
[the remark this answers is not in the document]
pwithnall: I wasn’t thinking from a performance point of view, more from a ‘doing edits atomically is good’ point of view. It’s a bit moot since if anything else is writing the `config` file at the same time as the updater we’re doomed anyway, regardless of whether we do one write/replace or two.

eos-updater/apply.c
```diff
@@ -255,6 +334,15 @@ apply_internal (ApplyData     *apply_data,
                local_error->message);
   g_clear_error (&local_error);
 
+  /* Try to add a collection ID to the OS remote, which enables LAN/USB OS
+   * updates when combined with eos-updater.conf changes. Don't treat any
+   * errors as fatal to the update. */
+  /*FIXME: Remove this function call after the next checxpoint */
```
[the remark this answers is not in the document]
pwithnall: there’s a typo in ‘checkpoint’ too, but whatever.

tests/test-update-from-main.c
```diff
@@ -135,6 +136,12 @@ test_update_from_main (EosUpdaterFixture *fixture,
                                  &branches_option, &error);
   g_assert_no_error (error);
   g_assert_cmpstr (branches_option, ==, expected_branches);
+
+  ostree_repo_get_remote_option (repo, default_remote_name,
+                                 "collection-id", NULL,
+                                 &collection_id_option, &error);
+  g_assert_no_error (error);
+  g_assert_cmpstr (collection_id_option, ==, default_collection_ref->collection_id);
```
pwithnall: Does this actually test that the collection ID has been set, and wasn’t set before? From a quick reading of the code, since `default_collection_ref` is non-`NULL`, it looks like `eos_test_client_new()` will set the collection ID in the repo `config` when it’s first created.

eos-updater/apply.c
```diff
@@ -255,6 +334,15 @@ apply_internal (ApplyData     *apply_data,
                local_error->message);
   g_clear_error (&local_error);
 
+  /* Try to add a collection ID to the OS remote, which enables LAN/USB OS
+   * updates when combined with eos-updater.conf changes. Don't treat any
+   * errors as fatal to the update. */
+  /* FIXME: Remove this function call after the next checxpoint */
```
pwithnall: Nitpick: Probably should also include a version number here, because at any point in time we will never actually reach the ‘next checkpoint’. i.e. `FIXME: Remove this function call after the first checkpoint after 3.4.8`.

eos-updater/apply.c
pwithnall: Agreed, that should definitely be fixed by `eos-stage-ostree`. I really don’t want any code in the system overwriting collection IDs when they are non-empty or set, as the collection ID is the root of a lot of trust. Anything automated should only ever be able to set a collection ID when it was previously unset, and should not be able to change or remove set collection IDs.

eos-updater/apply.c
```diff
@@ -96,6 +96,85 @@ get_test_osname (void)
...
+    return FALSE;
+  if (!ostree_repo_load_variant (repo,
+                                 OSTREE_OBJECT_TYPE_COMMIT,
+                                 checksum,
+                                 &commit,
+                                 error))
+    return FALSE;
+
+  metadata = g_variant_get_child_value (commit, 0);
+  collection_id_v = g_variant_lookup_value (metadata, "ostree.collection-binding", G_VARIANT_TYPE_STRING);
```
pwithnall: Nitpick: Use `OSTREE_COMMIT_META_KEY_COLLECTION_BINDING` instead of hard-coding the string.

pwithnall: Also, you can simplify this using `g_variant_lookup()`:
```
metadata = g_variant_get_child_value (commit, 0);
if (!g_variant_lookup (metadata, OSTREE_COMMIT_META_KEY_COLLECTION_BINDING, "&s", &collection_id))
  {
    g_debug (…);
    return TRUE;
  }
```

tests/test-update-from-main.c
mwleeds: Pushed a fixup so that we do check the transition from NULL to non-NULL

PR #235, opened 2018-09-22
T23391

mwleeds:


mwleeds:
This is made possible by https://github.com/flatpak/flatpak/pull/2124 and https://github.com/ostreedev/ostree/pull/1733

PR #236, opened 2018-09-24
T23851

mwleeds:


mwleeds:
This also has a couple commits that fix comments in addition to https://phabricator.endlessm.com/T23851

eos-updater-prepare-volume/eos-updater-prepare-volume
```diff
@@ -109,6 +109,10 @@ class VolumePreparer:
         _, remote_name, ref_name = OSTree.parse_refspec(refspec)
         if not remote_name:
             return None
+
+        # Technically the collection ID configured for updates might not match
+        # the one of the booted OSTree, but we can't use the correct one here
+        # until https://github.com/ostreedev/ostree/issues/1487 is fixed
```
ramcq: I'm not sure I agree this is blocked by that - in the specific case of one commit we're going to read anyway, can't we read the collection-id binding off the commit?

mwleeds: yeah, but then `create-usb` would fail since `/ostree/repo/summary` specifies the other (possibly wrong) collection ID along with the ref in question

pwithnall:
The Jenkins test failure was due to the `--commit` option not being supported yet due to https://github.com/ostreedev/ostree/pull/1735 not yet being backported. Merged anyway.

PR #237, opened 2018-09-27
eupv: cope with OSTree lock contention from other processes

treitter:
If another processes, such as `ostree commit`, has an exclusive lock on
the target repository, `e-u-p-v` will fail during `ostree summary
--update` or `ostree create-usb` if it hits the timeout (which is
currently 30 seconds as defined in OSTree's ostree-repo.c).

This fix simply retries these commands indefinitely until they succeed.

Without this fix in place, USB OS copies in Gnome Software (the App
Center) may fail unexpectedly as we have some processes, launched
independently, which can take an exclusive lock of the OSTree repo.

https://phabricator.endlessm.com/T23422

### DIFF
--- a/eos-updater-prepare-volume/eos-updater-prepare-volume
+++ b/eos-updater-prepare-volume/eos-updater-prepare-volume
@@ -244,7 +244,6 @@ class VolumePreparer:
         handler = signal.signal(signal.SIGINT, self._sigint_handler)
         self._ostree_set_lock_timeout(8*60)
 
-        self.__run(['ostree', 'summary', '--update', '--repo', repo_path])
         self.__run(['ostree', 'create-usb', '--repo', repo_path,
                     '--commit={}'.format(os_collection_ref[2]),
                     self.volume_path, os_collection_ref[0], os_collection_ref[1]])
--- a/eos-updater-prepare-volume/eos-updater-prepare-volume
+++ b/eos-updater-prepare-volume/eos-updater-prepare-volume
@@ -20,6 +20,7 @@
 
 import argparse
 import os
+import signal
 import subprocess
 import sys
 
@@ -45,12 +46,16 @@ class VolumePreparer:
     EXIT_FAILED = 1
     EXIT_INVALID_ARGUMENTS = 2
     EXIT_RUN_AS_ROOT = 3
+    # as of this writing, this is the default hard-coded in OSTree but there's
+    # no apparent way to retrieve it
+    OSTREE_LOCK_TIMEOUT_S_DEFAULT = 30
 
     def __init__(self, volume_path, flatpak_refs, quiet=False, debug=False):
         self.volume_path = volume_path
         self.flatpak_refs = flatpak_refs
         self.quiet = quiet
         self.debug = debug
+        self.ostree_lock_timeout_s = self.OSTREE_LOCK_TIMEOUT_S_DEFAULT
 
         self.sysroot = OSTree.Sysroot.new_default()
 
@@ -143,6 +148,40 @@ class VolumePreparer:
 
         return autoinstalls
 
+    def _ostree_get_lock_timeout(self):
+        """
+        Retrieve the OSTree lock timeout.
+
+        This determines how long commands like 'ostree summary --update' and
+        'ostree create-usb' will wait to get an exclusive lock on the repo
+        before failing.
+        """
+
+        timeout_s = self.OSTREE_LOCK_TIMEOUT_S_DEFAULT
+        try:
+            out = subprocess.check_output(['ostree', 'config', 'get',
+                                          'core.lock-timeout-secs'])
+            timeout_s = int(out)
+        except Exception:
+            print("no value set for OSTree's core.lock-timeout-secs config;"
+                  " using fallback of %d", timeout_s)
+
+        return timeout_s
+
+    def _ostree_set_lock_timeout(self, timeout_s):
+        try:
+            subprocess.check_call(['ostree', 'config', 'set',
+                                   'core.lock-timeout-secs', str(timeout_s)])
+        except subprocess.CalledProcessError:
+                sys.stderr.write("failed to set OSTree's core.lock-timeout-secs"
+                                 " config value")
+
+    def _ostree_restore_lock_timeout(self):
+        self._ostree_set_lock_timeout(self.ostree_lock_timeout_s)
+
+    def _sigint_handler(self, signal, frame):
+        self._ostree_restore_lock_timeout()
+
     def prepare_volume(self):
         # We need to be root in order to read all the files in the OSTree repo
         # (unless we’re running the unit tests). */
@@ -194,6 +233,17 @@ class VolumePreparer:
         # TODO: Verify that it adds GPG keys where appropriate.
         _, repo = self.sysroot.get_repo()
         repo_path = os.path.realpath(repo.get_path().get_path())
+
+        # Increase the OSTree exclusive lock timeout significantly for the
+        # following OSTree commands.
+        #
+        # This reduces (but does not eliminate) the risk of another process
+        # holding the lock so long one of these commands gives up and this
+        # entire program fails.
+        self.ostree_lock_timeout_s = self._ostree_get_lock_timeout()
+        handler = signal.signal(signal.SIGINT, self._sigint_handler)
+        self._ostree_set_lock_timeout(8*60)
+
         self.__run(['ostree', 'summary', '--update', '--repo', repo_path])
         self.__run(['ostree', 'create-usb', '--repo', repo_path,
                     '--commit={}'.format(os_collection_ref[2]),
@@ -201,6 +251,12 @@ class VolumePreparer:
         if len(all_flatpak_refs) > 0:
             self.__run(['flatpak', '--system', 'create-usb',
                         self.volume_path] + all_flatpak_refs)
+
+        # restore the original timeout so other OSTree processes wait the
+        # amount of time they normally would
+        self._ostree_restore_lock_timeout()
+        signal.signal(signal.SIGINT, handler)
+
         # Ensure that this USB drive can later be written to by non-root users
         self.__run(['chmod', '-R', '777', GLib.build_filenamev([self.volume_path, '.ostree'])])
 
--- a/eos-updater/apply.c
+++ b/eos-updater/apply.c
@@ -97,6 +97,82 @@ get_test_osname (void)
 }
 
 static gboolean
+update_remote_collection_id (OstreeRepo   *repo,
+                             const gchar  *refspec,
+                             GError      **error)
+{
+  g_autofree gchar *remote = NULL;
+  g_autofree gchar *ref = NULL;
+  g_autofree gchar *cur_collection_id = NULL;
+  g_autofree gchar *checksum = NULL;
+  g_autofree gchar *remote_group = NULL;
+  g_autoptr(GVariant) commit = NULL;
+  g_autoptr(GVariant) metadata = NULL;
+  g_autoptr(GKeyFile) config = NULL;
+  const char *collection_id;
+
+  if (!ostree_parse_refspec (refspec, &remote, &ref, error))
+    return FALSE;
+  if (remote == NULL)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                   "Invalid refspec ‘%s’ in origin: did not contain a remote name",
+                   refspec);
+      return FALSE;
+    }
+
+  /* If the remote already has a collection ID set, there's nothing to do */
+  if (!ostree_repo_get_remote_option (repo, remote, "collection-id", NULL,
+                                      &cur_collection_id, error))
+    return FALSE;
+  if (cur_collection_id != NULL && *cur_collection_id != '\0')
+    return TRUE;
+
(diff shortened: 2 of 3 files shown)


mwleeds:
The thing that worries me about this approach is that summary generation could still fail in the other places that it happens, like [here](https://github.com/endlessm/flatpak/blob/07a40735da62be5458b21a132bfe116ecc48bfb4/common/flatpak-dir.c#L8993) and [here](https://github.com/endlessm/eos-updater/blob/965462541be42a9e9d756426b88f6b5533cde7d1/libeos-update-server/repo.c#L809).

Maybe we should instead set a value higher than 30 for `core.lock-timeout-secs` in the default repo config, or stop whatever process is holding the lock for so long from doing that?

treitter:
I agree that a longer timeout would be preferable to this work-around (especially given it will never terminate if either the summary generation or `create-usb` fail though making it give up after a timeout or number of tries felt like another arbitrary number to deal with).

But if we do increase the timeout, it will likely have to be much longer (on the order of several minutes) to make the issue very hard to hit. And that carries the risk that we'll just trade this issue for another if we have implicit expectations that the timeout stays close to 30 seconds.

If there is another process holding the lock longer than it "should", it'd be good to fix that but we might have a hard time tracking that down and we'd retain the risk that any process that suddenly holds the lock for a longer-than-conventional time could re-trigger this bug.

ramcq:
The most likely holder of the lock is a prune operation, which suggests that the updater has finished deploying a new ostree and is tidying up the previous rollback, or that flatpak has done an install/update and is clearing away stuff.

In the end upstream I think we actually need to split the lock between refs and commits/repo contents. What summary generation/updating/etc needs is a shared reader lock on the refs, which can be shared with a prune operation - the prune needs an exclusive lock on the commits (ie repo content) but just needs the refs to be unchanged - other people can read refs/commits in parallel to the prune. Maybe file an up to

Prunes can take effing ages though, so surely the best way to reduce the chance of the lock contention is to arrange for GNOME Software to suspend the processing of the queue of installs, and maybe even queue the USB creation operation at the head of this queue so it doesn't even start until the current operation has finished.

I also don't think that e-u-p-v can take the lock and inherit it in the summary creation / usb creation unless we add an --inherit-lock or something? Or we make those API rather than scripts, but it's a bit late in the day for that.

Finally... I also don't really like the idea of looping indefinitely. If something is wedged taking the lock, it's better to fail the operation eventually rather than just fall down a well.

dbnicholson:
> It would be better to take the lock once at a high level, explicitly, from e-u-p-v itself.

I tried to convince Colin that it would be useful to make the lock APIs public, but he didn't buy it. He might be more open to it now that the locking has been in for a while and has broken everything.

ramcq:
> > It would be better to take the lock once at a high level, explicitly, from e-u-p-v itself.
> 
> I tried to convince Colin that it would be useful to make the lock APIs public, but he didn't buy it. He might be more open to it now that the locking has been in for a while and has broken everything.

Right, but this isn't even API calls, it's separate process invocations - the create USB is just a command, not an API.

dbnicholson:
> If another processes, such as `ostree commit`, has an exclusive lock on
> the target repository, `e-u-p-v` will fail during `ostree summary --update` or `ostree create-usb` if it hits the timeout (which is
> currently 30 seconds as defined in OSTree's ostree-repo.c).

I didn't see any description of the error here or in phabricator, so let me make sure I understand what's happening.

* `ostree summary -u` is called in the system repo. This is likely to have contention because it takes an exclusive lock, so anything with a shared (pull) or exclusive (prune) lock is going to block it from happening. When @mwleeds proposed upstream to have summary generation take an exclusive lock (https://github.com/ostreedev/ostree/pull/1681) I didn't envision it being run out of band by random applications. Quite possibly that should be eased to a shared lock. If the locking API was also public and you were paranoid about the repo contents, you could first take an exclusive lock before regenerating the summmary.

* `ostree create-usb` is called to pull from the system repo to the USB repo and then updates the summary on the USB repo. I can't see this failing due to lock contention. Local pulls don't attempt to lock the source repo (although that wouldn't be a bad idea), and I can't see any contention on the USB repo. Are you actually seeing locking errors in `ostree create-usb`?

Did I understand that correctly?

dbnicholson:
> > > It would be better to take the lock once at a high level, explicitly, from e-u-p-v itself.
> > 
> > 
> > I tried to convince Colin that it would be useful to make the lock APIs public, but he didn't buy it. He might be more open to it now that the locking has been in for a while and has broken everything.
> 
> Right, but this isn't even API calls, it's separate process invocations - the create USB is just a command, not an API.

Right, I was talking about API, but this is something for a builtin. I actually had a fancy `--lock-timeout` option that got added to all the builtins in my original submittal - https://github.com/ostreedev/ostree/pull/1292/commits/cf2824ac61334cd6264cf398219c3246f078e0bf.

pwithnall:
> I also don't think that e-u-p-v can take the lock and inherit it in the summary creation / usb creation unless we add an --inherit-lock or something?

Yes, that was implicit in my suggestion.

> Finally... I also don't really like the idea of looping indefinitely. If something is wedged taking the lock, it's better to fail the operation eventually rather than just fall down a well.

Me too.

treitter:
I don't think I've seen anyone discuss the idea of simply increasing the lock wait timeout (and skipping the script changes I've proposed) as a near-term work-around while we figure out a better locking solution upstream.

What do people think? Do we expect risks if we increase it to, eg, 8 minutes?

mwleeds:
> Prunes can take effing ages though, so surely the best way to reduce the chance of the lock contention is to arrange for GNOME Software to suspend the processing of the queue of installs, and maybe even queue the USB creation operation at the head of this queue so it doesn't even start until the current operation has finished.

Couldn't eos-updater be the one doing the prune, if you're updating at the same time that you're copying to the USB? I suppose that's a less likely scenario than gnome-software parallelizing install/update operations. Or is the OS update one of those operations since gnome-software orchestrates OS updates these days?

ramcq:
> I don't think I've seen anyone discuss the idea of simply increasing the lock wait timeout (and skipping the script changes I've proposed) as a near-term work-around while we figure out a better locking solution upstream.
> 
> What do people think? Do we expect risks if we increase it to, eg, 8 minutes?

Well... the risk is the experience is a bit ass, because it just sits there for ages rather than making any visible headway. But if we're not able to do anything to stop the other activity or show the user we're not doing anything - I guess it would work...? Can you preserve it, set it, restore it afterwards?

ramcq:
> > Prunes can take effing ages though, so surely the best way to reduce the chance of the lock contention is to arrange for GNOME Software to suspend the processing of the queue of installs, and maybe even queue the USB creation operation at the head of this queue so it doesn't even start until the current operation has finished.
> 
> Couldn't eos-updater be the one doing the prune, if you're updating at the same time that you're copying to the USB? I suppose that's a less likely scenario than gnome-software parallelizing install/update operations. Or is the OS update one of those operations since gnome-software orchestrates OS updates these days?

I was writing with the assumption that gnome-software was controlling it for both - but now you ask, I don't know this 100% for sure. I actually don't see any changes to eos-updater to remove the timer, so maybe we're just scheduling the updates from g-s in addition to our 14 day timer?

Also - is it possible that this is a pretty unlikely race for normal users, but shows up in QA because the system is switched to dev and you start futzing around with it, so it's gonna be updating more often and pulling daily (or more frequent) builds from staging?

dbnicholson:
> I don't think I've seen anyone discuss the idea of simply increasing the lock wait timeout (and skipping the script changes I've proposed) as a near-term work-around while we figure out a better locking solution upstream.
> 
> What do people think? Do we expect risks if we increase it to, eg, 8 minutes?

I think you're going to be better off, but you're always going to see a lot of lock contention if regenerating a summary takes an exclusive lock. I think if you change this to a shared lock (as I did in my initial implementation in https://github.com/ostreedev/ostree/pull/1292/commits/eba8a175e9b9a8241b0e5b6eec5a29621d250df0), most of the contention will go away.

That said, increasing the default lock timeout a bit might be a good idea. I doubt most pruning operations could finish in less than 1 minute on our hardware.

treitter:
> > I don't think I've seen anyone discuss the idea of simply increasing the lock wait timeout (and skipping the script changes I've proposed) as a near-term work-around while we figure out a better locking solution upstream.
> > What do people think? Do we expect risks if we increase it to, eg, 8 minutes?
> 
> Well... the risk is the experience is a bit ass, because it just sits there for ages rather than making any visible headway. But if we're not able to do anything to stop the other activity or show the user we're not doing anything - I guess it would work...? Can you preserve it, set it, restore it afterwards?

From the user's perspective, the copy duration will just be larger (and, without a frame of reference, if someone thinks 30 minutes is a long time, I don't think they'd think 22 minutes is a short time). I'd definitely prefer a very-reliable, slower OS copy to one that's unreliable.

And Dan's right, this isn't a total fix because we're just making it less likely to fail. I agree we should still come up with a better long-term fix with upstream.

I'll test out reading the value, bumping it up, doing the copy, then re-setting it.

mwleeds:
> you're always going to see a lot of lock contention if regenerating a summary takes an exclusive lock. I think if you change this to a shared lock (as I did in my initial implementation in ostreedev/ostree@eba8a17), most of the contention will go away.

That would protect against commits being deleted but wouldn't protect against refs being added/removed/updated, right? Is the idea that the latter race doesn't often matter in practice?

dbnicholson:
> > you're always going to see a lot of lock contention if regenerating a summary takes an exclusive lock. I think if you change this to a shared lock (as I did in my initial implementation in [ostreedev/ostree@eba8a17](https://github.com/ostreedev/ostree/commit/eba8a17)), most of the contention will go away.
> 
> That would protect against commits being deleted but wouldn't protect against refs being added/removed/updated, right? Is the idea that the latter race doesn't often matter in practice?

Right. Currently exclusive locks are taken when pruning objects and deltas, but not refs. I think it would be more robust to require an exclusive lock when generating the summary to ensure the refs aren't changed. However, even if you were able to block refs from being updated, a subsequent ref update would immediately invalidate the summary. The summary generation is effectively 2 reads (refs and deltas) followed by an atomic write to the summary file. What the locking would gain you from this perspective is to help ensure you don't get failures traversing the repo. Perhaps the locking doesn't really gain anything in this context.

On the other hand, there is one case where I would advise having locking during summary generation. If you're going to sign the summary, then you really don't want the summary to change while you're generating the signature. Otherwise you could publish mismatched summary and signature. That just inflicts pain on clients. That's the real reason why I had shared lock for summary generation and exclusive lock for summary signing in my locking PR even though I don't think that's what I put in the commit message or comment.

So, circling back to this issue, I think it would be fine to reduce the summary generation locking to shared or remove it entirely until the summary signing gains locking.

treitter:
I've force-pushed a new version which increases the value of `core.lock-timeouts-secs` while the critical commands are running then restores afterward (or in case the process gets SIGINT).

It seems to work reliably in my testing.

If everyone is OK with this being the work-around for 3.4.8, I think we should merge it to `master` and `eos3.4` and test. If that works, create a follow-up task to implement a longer-term fix.

eos-updater-prepare-volume/eos-updater-prepare-volume
```diff
@@ -143,6 +148,40 @@ class VolumePreparer:
...
+        This determines how long commands like 'ostree summary --update' and
+        'ostree create-usb' will wait to get an exclusive lock on the repo
+        before failing.
+        """
+
+        timeout_s = self.OSTREE_LOCK_TIMEOUT_S_DEFAULT
+        try:
+            out = subprocess.check_output(['ostree', 'config', 'get',
+                                          'core.lock-timeout-secs'])
+            timeout_s = float(out)
```
dbnicholson: The timeout is an integer, not a float.

treitter: It actually seems to be an arbitrary value based on `ostree-repo.c` but I'd be fine with treating it as an `int` here.

treitter: Fixed and tested in the version I just pushed

eos-updater-prepare-volume/eos-updater-prepare-volume
```diff
@@ -194,13 +233,30 @@ class VolumePreparer:
...
                     self.volume_path, os_collection_ref[0], os_collection_ref[1]])
+
         if len(all_flatpak_refs) > 0:
             self.__run(['flatpak', '--system', 'create-usb',
                         self.volume_path] + all_flatpak_refs)
+
+        # restore the original timeout so other OSTree processes wait the
+        # amount of time they normally would
+        self._ostree_restore_lock_timeout()
+        signal.signal(signal.SIGINT, signal.SIG_DFL)
```
dbnicholson: Instead of setting this to `signal.SIG_DFL`, you should save the earlier call to `signal.signal()` and reset it here. Python uses a different `SIGINT` handler to manage the `KeyboardInterrupt` exception.

treitter: Fixed in the newly-pushed version

treitter:
@dbnicholson 's suggestion to simply skip summary generation here and in `flatpak create-usb` seems to work, and it's a cleaner solution, so I'm going to close this pull request in favor of these:

https://github.com/endlessm/flatpak/pull/125
https://github.com/endlessm/eos-updater/pull/238

PR #238, opened 2018-10-02
eupv: don't manually generate the summary

treitter:
Generating the summary requires acquiring an exclusive lock on the
OSTree repository which risks this process failing if another process,
like `ostree commit` or `ostree prune`, holds an exclusive lock for
longer than OSTree's core.lock-timeout-secs config option (which
defaults to only 30 seconds which is not always sufficient).

In testing directly and with Gnome Software calling this script to copy
the OS and dependencies to USB, this step seems to be unnecessary, so
we're best off skipping it.

https://phabricator.endlessm.com/T23422

### DIFF
--- a/eos-updater-prepare-volume/eos-updater-prepare-volume
+++ b/eos-updater-prepare-volume/eos-updater-prepare-volume
@@ -190,11 +190,9 @@ class VolumePreparer:
         all_flatpak_refs = list(set(all_flatpak_refs))
 
         # Pass the heavy lifting off to `ostree create-usb` and `flatpak create-usb`
-        # which require an updated summary file
         # TODO: Verify that it adds GPG keys where appropriate.
         _, repo = self.sysroot.get_repo()
         repo_path = os.path.realpath(repo.get_path().get_path())
-        self.__run(['ostree', 'summary', '--update', '--repo', repo_path])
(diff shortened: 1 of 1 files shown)


eos-updater-prepare-volume/eos-updater-prepare-volume
```diff
@@ -190,11 +190,8 @@ class VolumePreparer:
         all_flatpak_refs = list(set(all_flatpak_refs))
 
         # Pass the heavy lifting off to `ostree create-usb` and `flatpak create-usb`
-        # which require an updated summary file
-        # TODO: Verify that it adds GPG keys where appropriate.
```
mwleeds: I think this comment should be deleted if I understand it correctly. GPG keys are not copied to the USB but they're not supposed to be. The destination computer has to already have the remote (and its keys) configured, and if the keys need to be updated that happens via `xa.gpg-keys` which is copied as part of `ostree-metadata`.

**But** I think you should delete the comment in a separate commit since it's unrelated to the summary update stuff.

treitter: I've addressed this in the new version I force-pushed.

Please verify my new commit message is accurate.

PR #240, opened 2018-10-15
eos-updater: Avoid scheduler for offline updates

mwleeds:
Currently eos-updater asks the download scheduler (mogwai) before
fetching an update, so that the user's configured update schedule can be
respected. However that system is designed to help users with limited
Internet connections; it doesn't make sense to use it when the update
comes from local sources, such as the local network or a USB drive. So
this commit makes changes to eos-updater so that offline OS updates can
happen regardless of whether or not automatic updates are turned on.

The way this is implemented is that in the Poll() stage, we first check
offline sources for updates, and then only check the Internet if none
were found offline. If we simply check all the sources at once as we
previously were, you'd have to make significant changes to the Fetch()
stage, because the scheduler can't make a decision when the offline and
online results are mixed together.

There are two notable implications of this implementation:

1. This means that even if the LAN or USB drive has an older update than
what's available on the Internet, we update from it first (then on the
next check for updates the Internet would be used). In my opinion, this
is desired behavior because after the offline update is complete, the
delta for the online update may be smaller.

2. Before this commit if the fetch from an offline source fails we try
to fetch from the Internet (if they both provide the latest commit).
After this commit, if the offline fetch fails the whole update fails.
This might be less than ideal but I don't think it will matter in
practice. The alternative would be to rearchitect eos-updater or the
libostree API used for the fetch in significant ways.

A mounted filesystem or a peer on the network could conceivably prevent
you from updating, but that's true both before and after this commit;
see ostreedev/ostree#1527

https://phabricator.endlessm.com/T23972

### DIFF
--- a/eos-updater/poll.c
+++ b/eos-updater/poll.c
@@ -155,14 +155,16 @@ object_unref0 (gpointer obj)
   g_object_unref (obj);
 }
 
-static GPtrArray *
+static void
 get_finders (SourcesConfig          *config,
              GMainContext           *context,
+             GPtrArray             **out_offline_finders,
+             GPtrArray             **out_online_finders,
              OstreeRepoFinderAvahi **out_finder_avahi)
 {
   g_autoptr(OstreeRepoFinderAvahi) finder_avahi = NULL;
-  g_autoptr(GPtrArray) finders = g_ptr_array_new_full (config->download_order->len,
-                                                       object_unref0);
+  g_autoptr(GPtrArray) offline_finders = g_ptr_array_new_full (0, object_unref0);
+  g_autoptr(GPtrArray) online_finders = g_ptr_array_new_full (0, object_unref0);
   g_autoptr(GError) local_error = NULL;
   gsize i;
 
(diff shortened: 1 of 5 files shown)


eos-updater/poll.c
```diff
@@ -211,7 +218,10 @@ get_finders (SourcesConfig          *config,
         }
     }
 
-  g_ptr_array_add (finders, NULL);  /* NULL terminator */
+  if (offline_finders->len > 0)
+    g_ptr_array_add (offline_finders, NULL);  /* NULL terminator */
```
wjt: Why's the NULL terminator needed? These are `GPtrArray`s which know their own length.

wjt: Ah, because their `pdata` ends up being passed as `OstreeRepoFinder **` to `ostree_repo_find_remotes_async`.

eos-updater/poll.c
```diff
@@ -200,8 +202,13 @@ get_finders (SourcesConfig          *config,
     {
       g_autoptr(OstreeRepoFinderOverride) finder_override = ostree_repo_finder_override_new ();
 
-      g_ptr_array_set_size (finders, 0);  /* override everything */
-      g_ptr_array_add (finders, g_object_ref (finder_override));
+      g_ptr_array_set_size (offline_finders, 0);  /* override everything */
+      g_ptr_array_set_size (online_finders, 0);  /* override everything */
+
+      /* We don't know if the URIs are online or offline; assume online so we
```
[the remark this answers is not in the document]
wjt: Agreed, this doesn't seem worthwhile.

pwithnall:
Mogwai should always be allowing download requests to `file:` URIs or URIs on the local network, though. If that’s not working, it needs to be fixed in Mogwai.

Implication 1 is a good enough reason for this PR to be accepted on its own merits, though.

PR #242, opened 2019-03-26
T16499 Add documentation to D-Bus interface

pwithnall:
https://phabricator.endlessm.com/T16499


eos-updater/com.endlessm.Updater.xml
```diff
@@ -2,49 +2,297 @@
...
+    @short_description: Endless OS updater control interface
+
+    Interface to query the state of the OS updater, and to control it: checking
+    for, downloading and applying updates.
+
+    The updater moves through a series of states (see the `State` property and
+    `StateChanged` signal), with state transitions triggered by the appropriate
+    method call, or by an error in an operation. It will return a
+    `com.endlessm.Updater.Error.WrongState` error if you attempt to perform an
+    invalid state transition.
```
wjt: Not one for this PR but I do sometimes wonder whether `Poll` in state `Polling`, `Fetch` in state `Fetching`, etc. shouldn't be no-ops rather than errors.

eos-updater/com.endlessm.Updater.xml
```diff
@@ -2,49 +2,297 @@
...
+
+    <!--
+      PollVolume:
+      @path: Absolute path to the root directory of a volume to check.
+
+      Like `Poll`, except this polls a specific removable volume (such as a USB
+      stick), rather than the internet.
+
+      If a `.ostree/repo` directory is available beneath @path, and it contains
+      a valid OSTree repository, that repository will be checked for updates.
```
wjt: And if not? Does the state change to Error?

eos-updater/com.endlessm.Updater.xml
```diff
@@ -2,49 +2,297 @@
...
+    <!--
+      FetchFull:
+      @options: Potentially empty dictionary of options.
+
+      Like `Fetch`, except options may be provided to affect its behaviour.
+      Currently, the following options are supported (unsupported options are
+      ignored):
+
+        * `force` (type: `b`): If true, force the download without scheduling
+          it through the system’s metered data scheduler.
```
wjt: Perhaps worth saying:

> Typically, this would be true in response to an explicit user action, and false otherwise.

eos-updater/com.endlessm.Updater.xml
```diff
@@ -2,49 +2,297 @@
...
-      -->
-    <property name="ErrorName"        type="s" access="read"/>
-    <!-- a human-readable, albeit unlocalized, error message -->
-    <property name="ErrorMessage"     type="s" access="read"/>
 
+    <!--
+      ErrorCode:
+
+      Error code of the current error, or `0` if no error has been reported.
+      This is in an unspecified error doman, and hence is useless.
```
wjt: :ok_hand:

eos-updater/com.endlessm.Updater.xml
```diff
@@ -2,49 +2,297 @@
...
+    -->
+    <property name="ErrorCode" type="u" access="read"/>
+
+    <!--
+      ErrorName:
+
+      A fully-qualified D-Bus error name, as might be returned from a D-Bus
+      method. `com.endlessm.Updater.Error.LiveBoot` when the system is a
+      read-only live USB, where updates are not supported.
```
wjt: I think this sentence is redundant with the list below.

eos-updater/com.endlessm.Updater.xml
```diff
@@ -293,8 +293,11 @@
       @expected: Number of bytes expected.
 
       Unused.
+
+      Deprecated: Use `org.freedesktop.DBus.Properties.PropertiesChanged`.
```
wjt: If it's not emitted by the daemon, or consumed by any existing clients, can we just drop it here?

wjt:
Build is failing because of the GCC 8 transition:

```
eos-updater/fetch.c: In function ‘check_scheduler’:
eos-updater/fetch.c:1026:46: error: cast between incompatible function types from ‘gboolean (*)(GCancellable *, void *)’ {aka ‘int (*)(struct _GCancellable *, void *)’} to ‘gboolean (*)(void *)’ {aka ‘int (*)(void *)’} [-Werror=cast-function-type]
   g_source_set_callback (cancellable_source, (GSourceFunc) cancellable_source_cb, NULL, NULL);
                                              ^
mv -f eos-updater/.deps/eos_updater-live-boot.Tpo eos-updater/.deps/eos_updater-live-boot.Po
```

One for a `G_SOURCE_FUNC()` cast?

eos-updater/com.endlessm.Updater.xml
pwithnall: Probably. It’s not used by gnome-software; not sure where else it could be used.

eos-updater/com.endlessm.Updater.xml
pwithnall: :ok:

eos-updater/com.endlessm.Updater.xml
pwithnall: :ok_hand:

eos-updater/com.endlessm.Updater.xml
pwithnall: It’ll go to `Ready`, afaik, since no update is available.

eos-updater/com.endlessm.Updater.xml
pwithnall: Returning an error gives the caller a bit more information. They should be handling errors discriminately in any case, so should be able to distinguish it from other, more error-y, errors.

pwithnall:
> One for a `G_SOURCE_FUNC()` cast?

Indeed, done.

wjt:
Looks fine… but the build is now failing for new reasons :(

pwithnall:
> I think this implicitly assumes that `sizeof (guint) < sizeof (gint64)`.

It does. I’m finding it quite frustrating that I can’t think of a way to add an `int` to an `unsigned int` without getting either a `-Wsign-conversion` warning, or a `-Wduplicated-branches` warning.

I’ve reworked it as per your suggestion, thanks. Not 100% happy with that either, since it makes the function less general. Oh well.

eos-updater/com.endlessm.Updater.xml
pwithnall: Dropped.

PR #244, opened 2019-03-28
tests: Add mock eos-updater service

pwithnall:
This is a mock implementation of eos-updater written using
python-dbusmock. It’s distributed by default, but not used by anything
at the moment; eventually some automated tests for eos-updater clients
could be written with it (using python-unittest), but for the moment
it’s only being used manually (using `python3 -m dbusmock --template
./path/to/mock-eos-updater.py`).

https://phabricator.endlessm.com/T26011

### DIFF
--- a/eos-updater/tests/eos_updater.py
+++ b/eos-updater/tests/eos_updater.py
@@ -1,0 +1,436 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
+# Copyright © 2019 Endless Mobile, Inc.
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+# MA  02110-1301  USA
+
+'''eos-updater mock template
+
+This creates a mock eos-updater interface (com.endlessm.Updater), with several
+methods on the Mock sidecar interface which allow its internal state flow to be
+controlled.
+
+A typical call chain for this would be:
+ - Test harness calls SetPollAction('update', {}, '', '')
+ - SUT calls Poll()
+ - Test harness calls FinishPoll()
+ - SUT calls Fetch()
+ - Test harness calls FinishFetch()
+ - SUT calls Apply()
+ - Test harness calls FinishApply()
+
+Errors can be simulated by specifying an `early-error` or `late-error` as the
+action in a Set*Action() call. `early-error` will result in the associated
+Poll() call (for example) transitioning to the error state. `late-error` will
+result in a transition to the error state only once (for example) FinishPoll()
+is called.
+
+See the implementation of each Set*Action() method for the set of actions it
+supports.
+
+Usage:
+   python3 -m dbusmock \
+       --template ./eos-updater/tests/eos_updater.py
+'''
+
+from enum import IntEnum
+from gi.repository import GLib
+import time
+
+import dbus
+import dbus.mainloop.glib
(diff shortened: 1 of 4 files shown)


pwithnall:
This is marked as WIP until the gnome-software plugin lands upstream; I don’t want to merge this to eos-updater until I’m sure it’s the final version we’re using.

This PR is essentially a reminder for that.

pwithnall:
Updated with the latest state of the mock service from https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/197. This fixes a problem where `FinishFetch()` would block any incoming `Cancel()` calls until it finished.

pwithnall:
This can land! It’s been in use upstream in gnome-software for testing the eos-updater plugin for a while.

I’ve reformatted it slightly, added it to meson and added REUSE information, but the PR is otherwise unchanged from 3.5 years ago.

eos-updater/tests/eos_updater.py
```diff
@@ -0,0 +1,428 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
+# Copyright © 2019 Endless Mobile, Inc.
```
wjt, line 4:
```suggestion
# Copyright © 2019 Endless OS Foundation LLC
```

eos-updater/tests/eos_updater.py
```diff
@@ -0,0 +1,428 @@
...
+import time
+
+import dbus
+import dbus.mainloop.glib
+from dbusmock import MOCK_IFACE
+
+
+__author__ = 'Philip Withnall'
+__email__ = 'pwithnall@endlessos.org'
+__copyright__ = '© 2019 Endless Mobile Inc.'
```
wjt, line 63:
```suggestion
__copyright__ = '© 2019 Endless OS Foundation LLC'
```

eos-updater/tests/meson.build
```diff
@@ -0,0 +1,12 @@
+# Copyright 2022 Endless OS Foundation, LLC
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+# python-dbusmock doesn’t currently provide a way to install out-of-tree templates
+
+# lint check
+test(
+  'flake8',
+  py3,
+  args: ['-m', 'flake8', files('eos_updater.py')],
```
wjt, line 10: Might we want to check for the presence of flake8? Two ways to do this:

1. Invoke it via its wrapper
```
test(
  'flake8',
  find_program('flake8'),
  args: [files('eos_updater.py')],
  ...
)
```

2. One can use the `python3` module to find a python3 installation that has the desired modules (`flake8` and `dbusmock` here I guess)

eos-updater/tests/eos_updater.py
```diff
@@ -0,0 +1,428 @@
...
+        UpdaterState.READY,
+        UpdaterState.UPDATE_AVAILABLE,
+        UpdaterState.UPDATE_READY,
+        UpdaterState.ERROR,
+    ]))
+
+    self.__change_state(self, UpdaterState.POLLING)
+
+    if self.__poll_action == 'early-error':
+        time.sleep(0.5)
```
wjt, line 191: What's the purpose of the `time.sleep()` calls?

eos-updater/tests/eos_updater.py
pwithnall: At the time, we weren’t yet Endless OS Foundation (that happened in 2020), which is why I kept this as Endless Mobile. I believe that’s correct, although it’s all a bit moot because the copyright ownership has transferred.

eos-updater/tests/meson.build
pwithnall: `python-flake8` is listed in `Build-Depends` in `debian/control`, and this meson snippet is the same as the other two uses of `flake8` in the source tree.

Using `find_program()` might make meson fail a little earlier if it’s not found, but that would require installing the [`flake8`](https://packages.debian.org/bullseye/all/flake8/filelist) package as well as [`python3-flake8`](https://packages.debian.org/bullseye/all/python3-flake8/filelist) (the latter only provides the module). So that seems a little over the top.

I’ll look at adding a commit to search for the module, which will at least make meson fail a bit more gracefully when building it without Debian packaging on a system which doesn’t have `python3-flake8` installed.

eos-updater/tests/eos_updater.py
pwithnall: To simulate network activity, iirc. I’ll add a comment.

eos-updater/tests/meson.build
wjt: OK, I didn't realise it's already used in this project.

pwithnall: I’ve added an additional commit which checks for it when finding the Python installation.

wjt (approved):

PR #245, opened 2019-04-10
eupv: Use collection ID from commit metadata

mwleeds:
Currently eos-updater-prepare-volume uses the collection ID from the
remote config when deciding how to specify the OS ref to `ostree
create-usb`. However this is not always accurate since for example an
image built using a staging ostree (com.endlessm.Dev.Os) will still have the production URL
and collection ID configured for updates (com.endlessm.Os). So this
commit changes e-u-p-v to use the collection ID from the commit metadata
instead, which is always the correct one.

When I wrote this I thought it wouldn't matter because I thought `ostree
create-usb` uses the collection ID from the local repo summary file,
which would still be the wrong one. However I later found out (thanks to
Dan) that the local repo summary is not needed or used for pulls between
local repos (in this case from the system repo to the one on the USB
drive).

So now copying a USB OS update on a staging image should work regardless
of where the eos remote is pointed, as long as the computer receiving
the update is pointed at staging.

https://phabricator.endlessm.com/T24034

### DIFF
--- a/eos-updater-prepare-volume/eos-updater-prepare-volume
+++ b/eos-updater-prepare-volume/eos-updater-prepare-volume
@@ -80,10 +80,17 @@ class VolumePreparer:
             # case libflatpak changes it in future.
             return False
 
-    def _get_collection_id_for_remote(self, remote_name):
-        """Get the configured collection ID for the given remote, or None."""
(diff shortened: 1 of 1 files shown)


pwithnall:
LGTM.

PR #246, opened 2019-04-16
debian: Update to compat level 12

pwithnall:
From https://manpages.debian.org/testing/debhelper/debhelper.7.en.html:

 • dh_systemd_start has been replaced by dh_installsystemd.
 • dh_install --fail-missing has been replaced by dh_missing.
 • dh autoreconf and systemd helpers are enabled by default.
 • dh uses --parallel by default for build systems which support it.


debian/compat
```diff
@@ -1 +1 @@
-9
+12
```
dbnicholson: You also should change the `debhelper` `Build-Depends` to ensure a new enough debhelper is installed. The proper way to do this is to set it to `debhelper-compat (=12)` per https://manpages.debian.org/unstable/debhelper/debhelper.7.en.html#COMPATIBILITY_LEVELS.

pwithnall: Thanks ⇒ https://github.com/endlessm/eos-updater/pull/248

PR #248, opened 2019-04-17
More Debian compatibility updates

pwithnall:
Following up from https://github.com/endlessm/eos-updater/pull/246#pullrequestreview-227246591.


debian/compat
```diff
@@ -1 +0,0 @@
-12
```
dbnicholson: I didn't realize you could drop the `compat` file after depending on `debhelper-compat`, but indeed that's what the manual says now.

PR #255, opened 2019-05-23
libeos-updater-util: Check for new flatpak error code

mwleeds:
libflatpak now returns FLATPAK_ERROR_REF_NOT_FOUND instead
of G_IO_ERROR_NOT_FOUND (and sometimes other errors) when a
ref can't be found either locally or in a remote. So update
some code which checks for G_IO_ERROR_NOT_FOUND.

See these commits:
https://github.com/flatpak/flatpak/commit/c0d5f1bfc8f09d5c7038422ecd948f26eaad747c
https://github.com/flatpak/flatpak/commit/c028e0dc1b7861a507c39f76c5bb582ee2150da1

https://phabricator.endlessm.com/T26678

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -9,7 +9,7 @@ Build-Depends:
  dh-autoreconf,
  dh-python,
  eos-metrics-0-dev,
- flatpak (>= 1.1.2),
+ flatpak (>= 1.3.3+dev37.20fb7d1-0),
  gir1.2-flatpak-1.0,
  gir1.2-glib-2.0,
  gir1.2-ostree-1.0,
--- a/libeos-updater-util/flatpak.c
+++ b/libeos-updater-util/flatpak.c
@@ -1820,18 +1820,25 @@ fetch_runtime_ref_for_source_ref (FlatpakInstallation *installation,
 
   if (bytes == NULL)
     {
-      /* G_IO_ERROR_NOT_FOUND here means that either the remote did not have
-       * a summary cache or that the source ref was not found in the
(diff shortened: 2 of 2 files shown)


mwleeds:
This is dependent on https://github.com/endlessm/flatpak/pull/158

pwithnall:
> This is dependent on [endlessm/flatpak#158](https://github.com/endlessm/flatpak/pull/158)

Could bump the version dependency in `debian/control` to reflect that.

mwleeds:
> > This is dependent on [endlessm/flatpak#158](https://github.com/endlessm/flatpak/pull/158)
> 
> Could bump the version dependency in `debian/control` to reflect that.

I think this would require doing a release of our fork of Flatpak which doesn't seem worth it.

pwithnall:
> I think this would require doing a release of our fork of Flatpak which doesn't seem worth it.

Shouldn’t do — if the fork already has the right API in it, then just use the current flatpak package version from OBS (`1.3.3+dev33.49c8941-0`). Next time we update the fork, that version number should increase monotonically.

mwleeds:
> > I think this would require doing a release of our fork of Flatpak which doesn't seem worth it.
> 
> Shouldn’t do — if the fork already has the right API in it, then just use the current flatpak package version from OBS (`1.3.3+dev33.49c8941-0`). Next time we update the fork, that version number should increase monotonically.

Okay, I made that change to `debian/control` so after OBS publishes the new build of flatpak we can merge this.

PR #256, opened 2019-08-06
Move eos-updater executable to /usr/libexec

wjt:
I constantly run `eos-updater update` rather than `eos-updater-ctl
update` thanks to tab completion. There's no reason for this executable
to be in $PATH.

https://phabricator.endlessm.com/T27471

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -221,7 +221,7 @@ eos_updater_libeos_updater_dbus_la_SOURCES = \
 	$(NULL)
 nodist_eos_updater_libeos_updater_dbus_la_SOURCES = $(eos_updater_codegen_sources)
 
-bin_PROGRAMS += eos-updater/eos-updater
+libexec_PROGRAMS += eos-updater/eos-updater
 
 eos_updater_eos_updater_CPPFLAGS = \
 	$(AM_CPPFLAGS) \
--- a/debian/eos-updater.install
+++ b/debian/eos-updater.install
@@ -6,7 +6,7 @@ lib/systemd/system/eos-updater-flatpak-installer.service
 lib/systemd/system/eos-updater.service
 lib/systemd/system/eos-update-server.service
 lib/systemd/system/eos-update-server.socket
-usr/bin/eos-updater
+usr/libexec/eos-updater
 usr/libexec/eos-autoupdater
 usr/libexec/eos-updater-avahi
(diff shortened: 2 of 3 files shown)


wjt:
Tested as follows:

* Built the package locally
* `eos-dev-unlock`
* `dpkg -i eos-updater_0-1_amd64.deb eos-updater-tools_0-1_amd64.deb gir1.2-eos-updater-0_0-1_amd64.deb`
* `systemctl restart eos-updater.service`
* Verified with `pgrep -a eos-updater` that it is running from `/usr/libexec/eos-updater`
* `eos-updater-ctl update`
* Booted into the update

wjt:
The only one which comes to mind is:

```
wjt@camille:~$ cat /proc/`pidof eos-updater`/cgroup 
12:memory:/system.slice/eos-updater.service
11:rdma:/
10:pids:/system.slice/eos-updater.service
9:freezer:/
8:devices:/system.slice/eos-updater.service
7:cpuset:/
6:perf_event:/
5:net_cls,net_prio:/
4:blkio:/
3:cpu,cpuacct:/
2:hugetlb:/
1:name=systemd:/system.slice/eos-updater.service
0::/system.slice/eos-updater.service
```

Probably harder to forge?

.oO( should D-Bus grow `GetConnectionUnixPidfd() → (h)`? cf https://lwn.net/Articles/794707/ )

pwithnall:
> Probably harder to forge?

The issue is not forging the identifying information; it’s the race between `dbus-daemon` returning an identifier, and the client process then looking that identifier up again. Particularly with PIDs, this is wide open to races with PID reuse. Currently, [we can only get a limited amount of information](https://dbus.freedesktop.org/doc/dbus-specification.html#bus-messages-get-connection-credentials) about a peer race-free.

Supporting pidfds is a nice idea. I’ve filed [an issue](https://gitlab.freedesktop.org/dbus/dbus/issues/274).

wjt:
> > Probably harder to forge?
> 
> The issue is not forging the identifying information; it’s the race between `dbus-daemon` returning an identifier, and the client process then looking that identifier up again. Particularly with PIDs, this is wide open to races with PID reuse. Currently, [we can only get a limited amount of information](https://dbus.freedesktop.org/doc/dbus-specification.html#bus-messages-get-connection-credentials) about a peer race-free.

Oh, right!

PR #257, opened 2019-09-06
eos-autoupdater: Add a timer for PAYG systems

mwleeds:
On PAYG Endless images, if eos-paygd hits a fatal error condition (such
as a seg fault), the whole system is brought down. This creates a
reliability issue, because if we ever have such a bug, and it happens on
every boot, and it happens early enough in the boot, it may prevent the
machine from ever updating and receiving the bug fix. The automatic
updater is currently configured to run 15 minutes after boot up, to give
NTP a chance to correct the clock. But this shouldn't be a concern for
PAYG computers, which should have their clocks set correctly during
provisioning. So introduce a new timer, eos-autoupdater-payg.timer,
which activates eos-autoupdater.service immediately after boot (really
after the network comes up since the service has After=network.target).
Use a kernel command line parameter to ensure the new timer only runs on
PAYG images, and the existing timer only runs on non-PAYG images.

Both timers also have RandomizedDelaySec=30m, to reduce the chances that
many computers on the same network will update at once and overload the
network. This means the update won't often happen immediately after the
network comes up, but seems like a worthwhile trade-off.

https://phabricator.endlessm.com/T27051

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -166,6 +166,7 @@ nodist_eos_autoupdater_eos_autoupdater_SOURCES = \
 eos_autoupdater_in = \
 	eos-autoupdater/eos-autoupdater.service.in \
 	eos-autoupdater/eos-autoupdater.timer.in \
+	eos-autoupdater/eos-autoupdater-payg.timer.in \
 	$(NULL)
 
 systemdsystemunit_DATA += $(eos_autoupdater_in:%.in=%)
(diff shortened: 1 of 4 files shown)


mwleeds:
I filed this task about the deprecation warnings causing the Jenkins build to fail: https://phabricator.endlessm.com/T27723

pwithnall:
Rather than add a new timer which is used for PAYG, I think we could probably fix this by fixing https://phabricator.endlessm.com/T16671, which is a way of removing the arbitrary 15 minute timeout on the main timer.

treitter:
> Rather than add a new timer which is used for PAYG, I think we could probably fix this by fixing https://phabricator.endlessm.com/T16671, which is a way of removing the arbitrary 15 minute timeout on the main timer.

I agree this sounds like a better fix than adding another arbitrary timer.

PR #259, opened 2019-10-03
Allow compilation against upstream OSTree

valentindavid:
`ostree_repo_get_commit_sizes` is not available on upstream OSTree. As
a result, it failed to compile against it.  When the function is not
available, only the fallback to the call to
`ostree_repo_get_commit_sizes` is executed.


pwithnall:
@eos-shell-build, test this please

PR #260, opened 2019-10-05
eos-autoupdater: Lower OnBootSec from 15m to 3m

mwleeds:
This commit makes eos-autoupdater.timer potentially start 3 minutes
after boot, rather than 15 minutes after boot. Because we have
RandomizedDelaySec=30m, in practice it will start between 3 and 33
minutes after boot. The primary reason for the change is that we will
soon be shipping PAYG Endless computers, on which if the eos-paygd
daemon crashes or exits with an error code, the whole computer shuts
down after a 20 minute grace period. This poses a risk of making
computers stuck in a boot loop and never receiving an update which might
fix the problem, in case the hypothetical crash happens before
eos-updater runs. So make such a boot loop less likely by having the
updater run automatically earlier after boot.

The reason the updater currently runs 15 minutes after boot is to give
ntp a chance to update an incorrect clock to the correct time, because a
sufficiently incorrect clock would prevent GPG signature verification
from working.[1] One might think this problem could be solved by adding
After=time-sync.target to the unit, but in practice that target is
reached before the clock is synchronized.[2] And if we were to enable
systemd-time-wait-sync.service to make time-sync.target work as
advertised, it's not clear when updates would be triggered on offline
machines, and according to the man page, systemd-time-wait-sync.service
is unreliable when used in conjunction with ntp rather than
systemd-timesyncd.service (and Endless uses ntp).[3] I think we should
just accept that there's some chance updates will fail due to a wrong
clock or for other reasons, and the updater will just try again when
OnUnitInactiveSec=1h is reached or a reboot occurs.

By starting 3 minutes after boot, we're giving the machine some time to
get online and get the clock corrected, without blocking on the
completion of either of those tasks. It wouldn't make sense to wait for
an Internet connection because we want LAN/USB updates to happen
automatically.[4] Also note that After=network-online.target doesn't
guarantee a working Internet connection, because
NetworkManager-wait-online.service only waits until NetworkManager has
started up.

[1] https://github.com/endlessm/eos-shell/issues/1478
[2] https://github.com/systemd/systemd/issues/5097
[3] https://www.freedesktop.org/software/systemd/man/systemd-time-wait-sync.service.html
[4] https://phabricator.endlessm.com/T21250

https://phabricator.endlessm.com/T27051

### DIFF
--- a/eos-autoupdater/eos-autoupdater.timer.in
+++ b/eos-autoupdater/eos-autoupdater.timer.in
@@ -5,7 +5,9 @@ ConditionKernelCommandLine=!endless.live_boot
 ConditionKernelCommandLine=ostree
 
 [Timer]
-OnBootSec=15m
+# After 3 minutes the network connection has likely been set up, but we don't
+# have a hard requirement; offline updates need to work too
+OnActiveSec=3m
 OnUnitInactiveSec=1h
 RandomizedDelaySec=30min
 


dbnicholson:
I didn't realize the purpose of that was for ntp. Thanks for digging that up. I always thought it was just so we didn't start immediately hogging your bandwidth as soon as the system starts. I guess in general I don't think there's any real harm in trying to start earlier. As @mwleeds notes, there will always be a possibility that updates fail because of clock skew, and the autoupdater will just try again later.

I think @dsd would have more information about how this would affect slow booting machines. One thing that might help on slow booting machines is using `OnActiveSec` rather than `OnBootSec`. This would not include the super slow early booting part in the timing. So, if you had a machine that took several minutes to boot, `OnBootSec` might trigger the update immediately. But that change is a bit orthogonal and I don't know exactly how that interacts with the paygd timing.

pwithnall:
> One thing that might help on slow booting machines is using `OnActiveSec` rather than `OnBootSec`. This would not include the super slow early booting part in the timing. So, if you had a machine that took several minutes to boot, `OnBootSec` might trigger the update immediately. But that change is a bit orthogonal and I don't know exactly how that interacts with the paygd timing.

If that interacts OK with the payg stuff, that does sound like a good improvement to make. That’s something new learned today :D

mwleeds:
> In any case, I think there should be some follow-up tasks about fixing those systemd issues so that we can trigger an update check when the clock has settled, and get rid of the hard-coded value permanently.

I don't think so. If we ever have to wait for the clock to settle, that means we're not allowing offline updates to happen. An offline machine might never reach an NTP server, but should still be allowed to automatically update from LAN/USB. I don't think a hard-coded value is such a bad thing in this case.


> > One thing that might help on slow booting machines is using `OnActiveSec` rather than `OnBootSec`. This would not include the super slow early booting part in the timing. So, if you had a machine that took several minutes to boot, `OnBootSec` might trigger the update immediately. But that change is a bit orthogonal and I don't know exactly how that interacts with the paygd timing.
> 
> If that interacts OK with the payg stuff, that does sound like a good improvement to make. That’s something new learned today :D

Using `OnActiveSec` does mean we'll be starting later relative to when eos-paygd starts (in the initramfs), but it also means the update is more likely to be successful (since it's more likely that we're online). I'll change the PR.

pwithnall:
> An offline machine might never reach an NTP server, but should still be allowed to automatically update from LAN/USB. I don't think a hard-coded value is such a bad thing in this case.

I wonder if offline machines should always advertise their clocks as synchronised, in the sense that they are as synchronised as they are ever going to get? Because otherwise `systemd-time-wait-sync.service` is useless, and that will (eventually, if it doesn’t now) affect services other than `eos-paygd` which are ordered based on `systemd-time-wait-sync.service`.

So I’m proposing that `eos-paygd` *does* depend on `systemd-time-wait-sync.service`, and we add another service, only on known-offline machines, which touches `/run/systemd/timesync/synchronized` on boot.

Yes/No?

mwleeds:
> > An offline machine might never reach an NTP server, but should still be allowed to automatically update from LAN/USB. I don't think a hard-coded value is such a bad thing in this case.
> 
> I wonder if offline machines should always advertise their clocks as synchronised, in the sense that they are as synchronised as they are ever going to get? Because otherwise `systemd-time-wait-sync.service` is useless, and that will (eventually, if it doesn’t now) affect services other than `eos-paygd` which are ordered based on `systemd-time-wait-sync.service`.
> 
> So I’m proposing that `eos-paygd` _does_ depend on `systemd-time-wait-sync.service`, and we add another service, only on known-offline machines, which touches `/run/systemd/timesync/synchronized` on boot.
> 
> Yes/No?
> 
> (If ‘yes’, https://phabricator.endlessm.com/T16671 should probably be reopened.)

To be clear I don't know what the behavior is if you order your service after `systemd-time-wait-sync.service` and you're offline. But even if currently it's okay (e.g. there's a timeout and your service is started after 90 seconds or something) that seems like the type of thing that isn't guaranteed to continue to be the case; systemd could decide to change that behavior.

I don't agree that `systemd-time-wait-sync.service` has to work offline and otherwise it's a bug. Perhaps services should only require that one if they expect to only work in an online environment and truly require a synchronized clock to work. I would argue that eos-updater doesn't strictly require a synchronized clock to work; if the clock is a bit off the update will still work and if the clock is far off from reality we hit a non-fatal error and try again later. It's not as if we're providing a service like Time-based OTP that only works if the clock is correct.

I also don't think it's a good idea to make a service that's only used on offline machines and have it pretend the clock is synchronized because (a) it's not feasible to know which computers will be offline at image build time (we only sometimes know and even then don't need the extra burden of keeping that config option accurate) and (b) maybe some OS component actually does need the clock to be synchronized, and falsely claiming it is by touching `/run/systemd/timesync/synchronized` might cause that component to malfunction.

PR #261, opened 2019-11-06
eos-updater-flatpak-installer: skip parental controls checks

wjt:
Without this, deploying Flatpaks after a reboot fails with the following
in the journal (line-wrapped for legibility):

    Attempting to install eos-apps:app/com.hack_computer.Clubhouse/x86_64/eos3
    Failed to install eos-apps:app/com.hack_computer.Clubhouse/x86_64/eos3:
      Could not connect: No such file or directory
    eos-updater-flatpak-installer: Couldn’t apply some flatpak update actions for this boot:
      Could not connect: No such file or directory

This problem does not occur if I trigger this service once the system is
fully booted.

I believe this is because deploying a Flatpak now involves a parental
controls check, which involves talking to accountsservice on the system
bus, neither of which are running at the point where this service runs.
Work around this by using the same workaround we used in the image
builder to skip the parental controls check.

https://phabricator.endlessm.com/T28530

### DIFF
--- a/eos-updater-flatpak-installer/eos-updater-flatpak-installer.service.in
+++ b/eos-updater-flatpak-installer/eos-updater-flatpak-installer.service.in
@@ -19,6 +19,11 @@ RemainAfterExit=true
 ExecStart=@libexecdir@/eos-updater-flatpak-installer
 Restart=no
 
+# Flatpak checks parental controls at deploy time. In order to do this, it
+# needs to talk to accountsservice on the system bus, neither of which are
+# running when this job runs.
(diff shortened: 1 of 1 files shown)


wjt:
I tested this change by adding that line to a copy of the file in `/etc/systemd/system`, deleting `/var/.updated` then rebooting. Hack appeared on my desktop.

ramcq:
> LGTM - maybe in the future we should reorder the service to run after the required deps though.

We use this system (hopefully) very rarely and only when we Really Want To, so I think it's reasonable for the system to circumvent the parental checks. The risk of not doing that is that we get into the danger zone of having boot dependency cycles where fixup jobs are trying to run on the config files / state of services which have been started. The "needs update" jobs run very early for this reason.

PR #262, opened 2019-12-06
Fix updates from 3.3.19

mwleeds:
https://phabricator.endlessm.com/T28626


mwleeds:
I know adding a debug message print on an old stable branch might seem strange, but if for some reason this doesn't fully fix updates from 3.3.19, it will be helpful to have.

libeos-updater-util/flatpak.c
```diff
@@ -1294,11 +1301,26 @@ squash_ref_actions_ptr_array (GPtrArray *ref_actions  /* (element-type EuuFlatpa
...
+      if (g_ptr_array_find_with_equal_func (squashed_ref_actions,
+                                            action,
+                                            euu_flatpak_remote_ref_action_equal_ref_only,
+                                            NULL))
+        continue;
+
+      squashed_action_for_ref = g_hash_table_lookup (hash_table, action->ref->ref);
+      g_assert (squashed_action_for_ref != NULL);
+      g_ptr_array_add (squashed_ref_actions,
+                       euu_flatpak_remote_ref_action_ref (squashed_action_for_ref));
```
pwithnall: I think this will produce an array which is sometimes not actually the same order as the input array, but I can’t think of a situation where that would cause a problem. It makes me nervous nonetheless.

An example of this would be where the same ref shows up twice in the input list, with two actions, and actions on some other ref in between. For example:
```
ref_actions = [ install A, install B, uninstall A ]
```
will get compressed to the following with this PR:
```
squashed_ref_actions = [ uninstall A, install B ]
```
but I think the original intent of the `squash_ref_actions_ptr_array()` function was to produce:
```
squashed_ref_actions = [ install B, uninstall A ]
```

What do you think?

pwithnall:
This looks like a good collection of fixes to me, aside from my question above.

libeos-updater-util/flatpak.c
mwleeds: Good point. I doubt that could cause a problem in practice, but it seems worth fixing anyway.

I think there's another potential issue here which is that if in your example A is a dependency of B, the installation of B will fail with either of the two sequences of squashed actions. That seems unlikely to happen since we'd have to add an action which uninstalls a runtime needed by a previous action. I don't think we need to address that here but I added a comment to https://phabricator.endlessm.com/T23150

Pushed fixups to address this and add a unit test. Also spent a while thinking about whether a dependency could be installed too late if it's needed by more than one action, but it looks like the later dependency will be ignored by determine_action_for_related_ref().

mwleeds:
I checked that updates still work from 3.3.19 with the fixup commit.

libeos-updater-util/flatpak.c
```diff
@@ -1294,11 +1301,37 @@ squash_ref_actions_ptr_array (GPtrArray *ref_actions  /* (element-type EuuFlatpa
...
+      EuuFlatpakRemoteRefAction *squashed_action_for_ref = NULL;
+
+      squashed_action_for_ref = g_hash_table_lookup (hash_table, action->ref->ref);
+      g_assert (squashed_action_for_ref != NULL);
+
+      /* Check that the action type matches so that
+       * [ install A, install B, uninstall A ] gets squashed into
+       * [ install B, uninstall A ] not [ uninstall A, install B ]
+       */
+      if (squashed_action_for_ref->type != action->type)
```
pwithnall: I’m not sure about this either: I think it would fail in the situation where `ref_actions = [ install A, uninstall A, install B, install A, uninstall A ]`, though I haven’t worked the scenario through in detail.

Could we use a pointer comparison `if (squashed_action_for_ref != action) continue` here instead? Basically what we want to achieve is to ignore all the instances of any action for a given ref, unless it’s the `squashed_action_for_ref` for that ref.

libeos-updater-util/flatpak.c
pwithnall: woohoo, unit tests! Would be good to add a couple more unit tests along the same lines to check the other ways that actions could be squashed and end up out of order. The test you added checks the case where an uninstall action squashes an install action, but there are also the situations:
 * Install squashes uninstall
 * Update squashes update
 * Install squashes update
 * Uninstall squashes updates

mwleeds: Honestly I don't think we need a unit test for every one of those. You'd have to make pretty obviously wrong changes in order to break those cases without breaking the test we have now, and each test we add does have some cost (to write and to maintain).

libeos-updater-util/flatpak.c
mwleeds: Agreed, pointer comparison is better. I also updated the unit test to cover this case.

libeos-updater-util/flatpak.c
pwithnall: > You'd have to make pretty obviously wrong changes in order to break those cases without breaking the test we have now

Well, you made some fairly subtle changes in this PR to go from not working for all the ways actions could be squashed, to working for more of them. So I’d argue that equally subtle changes could break them. The test harness is set up to allow easily adding more actions to a test, as you’ve already done to check the pointer comparison fix.

pwithnall:
I recommend that the extra unit test coverage is added, but I’m not going to block on that.

In any case, the fixup commits need to be squashed in; feel free to merge once that’s done.

libeos-updater-util/flatpak.c
mwleeds: But we are testing for the way in which this was subtly broken (namely, the case described in https://github.com/endlessm/eos-updater/pull/262#discussion_r355992045).

The order preservation code in squash_ref_actions_ptr_array() doesn't do anything that's conditional on action types, so if we added more tests for each type we'd just be re-testing the squashing code that's already covered by other tests.

Anyway, seems like we both agree it's not important either way, so merging this.

PR #263, opened 2019-12-06
T28626 forwardport

mwleeds:
Cherry-pick of commits from #262. Only one minor difference which is noted in the commit message.

https://phabricator.endlessm.com/T28626


mwleeds:
(this will need to be updated when #262 is finalized)

mwleeds:
Now this matches the commits on the `eos3.3` branch.

PR #266, opened 2020-01-27
poll: Use ostree_commit_get_object_sizes if available

dbnicholson:
OSTree upstream recently gained an API to return the parsed entries in
the `ostree.sizes` metadata[1]. Split out a helper function and use it
if it's available to sum the various commit size properies. There is one
slight behavior change here - the updater only continues on error if the
error is `G_IO_ERROR_NOT_FOUND`. This is what both APIs return when
`ostree.sizes` does not exist as well as what's now set when neither
APIs are available.

1. https://github.com/ostreedev/ostree/pull/1957

https://phabricator.endlessm.com/T18171

### DIFF
--- a/eos-updater/poll-common.c
+++ b/eos-updater/poll-common.c
@@ -1186,6 +1186,86 @@ string_to_download_source (const gchar *str,
   return TRUE;
 }
 
+/* Macros for handling saturated incrementing and clamping */
+#define SATURATED_INCREMENT_GUINT64(a, b)       \
+  G_STMT_START {                                \
+    if ((a) < G_MAXUINT64 - (b))                \
+      (a) += (b);                               \
+    else                                        \
+      (a) = G_MAXUINT64;                        \
+  } G_STMT_END
+#define CLAMP_GUINT64_TO_GINT64(a) \
+  G_STMT_START {                   \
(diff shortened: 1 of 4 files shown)


eos-updater/poll-common.c
```diff
@@ -1186,6 +1186,69 @@ string_to_download_source (const gchar *str,
...
+
+      if (!ostree_repo_has_object (repo, entry->objtype, entry->checksum,
+                                   &exists, cancellable, error))
+        return FALSE;
+
+      if (!exists)
+        {
+          /* Object not in local repo */
+          *new_archived += (gint64) entry->archived;
+          *new_unpacked += (gint64) entry->unpacked;
```
pwithnall: We should probably use saturated addition here. I’d also be tempted to use `guint64` for the types in `get_commit_sizes()` and then clamp down to `gint64` in the calling function, to give that function the scope for how it handles the possibility of sizes which are greater than `G_MAXINT64`.

dbnicholson:
After adding the saturating math I found out there was no testing for this, so I added the necessary infrastructure. I'm glad I did because I initially had the saturating math backwards...

Let me know if you had something else in mind.

PR #274, opened 2020-09-09
T25765 Fix transaction operation error handling in eufi

pwithnall:
Turns out that the `operation-error` signal needs to be handled in order for transactions to behave as you’d expect them to (for example, not failing on skipped operations, returning the error code from a failed operation).

See the individual commit messages for details.

This branch should improve our error reporting from EUFI. It may fix the upgrade-time installation of Totem, but I haven’t explicitly tested that (yet).

https://phabricator.endlessm.com/T25765

### DIFF
--- a/eos-updater-avahi/eos-updater-avahi.c
+++ b/eos-updater-avahi/eos-updater-avahi.c
@@ -38,8 +38,7 @@ get_refs (OstreeRepo            *repo,
 {
   g_autoptr(GHashTable) refs = NULL;
   GHashTableIter iter;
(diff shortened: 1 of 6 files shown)


pwithnall:
Marking as a draft as I haven’t tested this beyond running the (new and old) unit tests yet. It shouldn’t be merged until I’ve done some testing on images, and tested that it fixes the Totem issue on T25765.

mwleeds (approved):
LGTM, thanks for catching this

PR #279, opened 2020-11-24
eupv: Don't error out on missing flatpaks

mwleeds:
In a perfect world, USB drives created with eos-updater-prepare-volume
could be used to update Endless computers on any previous OS versions,
but alas, this is not a perfect world. One of the reasons we can't keep
that promise is checkpoints ("eos.checkpoint-target"), but we also can't
guarantee we'll have all the flatpaks needed by
eos-updater-flatpak-installer on the target machine receiving the
update. This is because flatpaks may be uninstalled by the user after
being auto-installed, or they may be installed and later uninstalled by
eos-updater-flatpak-installer, as will be the case for
com.hack_computer.ProjectLibrary in 3.9.1.

So make failures of the `flatpak create-usb` command non-fatal for
flatpaks not explicitly specified by the user. We may also want to
install any missing flatpaks before doing the copy when we have an
Internet connection, but that improvement is left as an exercise for the
reader.

https://phabricator.endlessm.com/T30368

### DIFF
--- a/eos-updater-prepare-volume/eos-updater-prepare-volume
+++ b/eos-updater-prepare-volume/eos-updater-prepare-volume
@@ -56,11 +56,17 @@ class VolumePreparer:
 
         self.sysroot = OSTree.Sysroot.new_default()
 
-    def __run(self, cmd):
+    def __run(self, cmd, fatal_error=True):
         """Run cmd locally."""
         print('# {}'.format(cmd))
 
-        subprocess.check_call(cmd)
+        try:
+            subprocess.check_call(cmd)
+        except subprocess.CalledProcessError:
+            if fatal_error:
+                raise
+            else:
+                sys.stderr.write('Warning: command failed: {}\n'.format(cmd))
 
     def __fail(self, exit_status, message):
         """Print an error to stderr and exit with the given error status."""
@@ -192,13 +198,6 @@ class VolumePreparer:
                                'OSTree deployment ref could not be found; '
                                'are you on an OSTree system?')
 
-        all_flatpak_refs = \
-            list(self.flatpak_refs) + \
-            list(autoinstall_flatpaks)
-
-        # Eliminate duplicates.
-        all_flatpak_refs = list(set(all_flatpak_refs))
-
         # Pass the heavy lifting off to `ostree create-usb` and
         # `flatpak create-usb`
         _, repo = self.sysroot.get_repo()
@@ -208,9 +207,18 @@ class VolumePreparer:
                         '--commit={}'.format(os_collection_ref[2]),
                         self.volume_path, os_collection_ref[0],
                         os_collection_ref[1]])
-        if len(all_flatpak_refs) > 0:
+        # For specified flatpaks, an error is fatal
+        for flatpak_ref in self.flatpak_refs:
             self.__run(['flatpak', '--system', 'create-usb',
-                        self.volume_path] + all_flatpak_refs)
+                        self.volume_path, flatpak_ref],
+                       fatal_error=True)
+        # For autoinstall flatpaks, an error is a warning. Some may have been
+        # manually or automatically uninstalled, and the drive can still be
+        # used to update some OS versions.
+        for flatpak_ref in autoinstall_flatpaks:
+            self.__run(['flatpak', '--system', 'create-usb',
+                        self.volume_path, flatpak_ref],
+                       fatal_error=False)
         # Ensure that this USB drive can later be read by any user and the
         # .ostree directory can be written to by any user
         if not self.preserve_permissions:
--- a/eos-updater-prepare-volume/eos-updater-prepare-volume
+++ b/eos-updater-prepare-volume/eos-updater-prepare-volume
@@ -156,6 +156,8 @@ class VolumePreparer:
             if action.type != \
                EosUpdaterUtil.FlatpakRemoteRefActionType.UNINSTALL:
                 autoinstalls.add(action.ref.ref.format_ref())
+            else:
+                autoinstalls.discard(action.ref.ref.format_ref())
 
         return autoinstalls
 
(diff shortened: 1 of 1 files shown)


eos-updater-prepare-volume/eos-updater-prepare-volume
```diff
@@ -56,11 +56,17 @@ class VolumePreparer:
...
-    def __run(self, cmd):
+    def __run(self, cmd, fatal_error=True):
         """Run cmd locally."""
         print('# {}'.format(cmd))
 
-        subprocess.check_call(cmd)
+        try:
+            subprocess.check_call(cmd)
+        except subprocess.CalledProcessError:
+            if (fatal_error):
```
wjt, line 66:
```suggestion
            if fatal_error:
```

wjt (changes requested):
If I understand correctly, this will allow the OS commit D on the USB stick to be used to update a machine from some previous commit C provided that the autoinstall file changes between C and D do not require any of the missing Flatpaks to be installed. But if C → D requires that one of the missing Flatpaks be installed, that update will still fail. Right?

So, this seems like an improvement, but it's another nail in the box of "we should make e-u-f-i failures non-fatal" IMO.

In the specific example you hit: ProjectLibrary was previously installed as part of an old update, and is now removed in 3.9.1. So this is a non-issue: it's missing, but 3.9.1 requires it be removed, so that's OK. If we had removed the old stanza that installed it from the autoinstall, would this change have been necessary? Also, does `_get_autoinstall_flatpaks()` ignore install actions which are subsequently countermanded by an uninstall action?

eos-updater-prepare-volume/eos-updater-prepare-volume
mwleeds: I'm too used to C code :laughing:

mwleeds:
> If I understand correctly, this will allow the OS commit D on the USB stick to be used to update a machine from some previous commit C provided that the autoinstall file changes between C and D do not require any of the missing Flatpaks to be installed. But if C → D requires that one of the missing Flatpaks be installed, that update will still fail. Right?
> 
Right
> So, this seems like an improvement, but it's another nail in the box of "we should make e-u-f-i failures non-fatal" IMO.
> 
I would tend to agree
> In the specific example you hit: ProjectLibrary was previously installed as part of an old update, and is now removed in 3.9.1. So this is a non-issue: it's missing, but 3.9.1 requires it be removed, so that's OK. If we had removed the old stanza that installed it from the autoinstall, would this change have been necessary?

No, it wouldn't have been necessary for this case (though as you point out it's still a useful change for the case when something has been manually uninstalled).

>Also, does `_get_autoinstall_flatpaks()` ignore install actions which are subsequently countermanded by an uninstall action?

No, but that would be an easy change to make. I'll do that.

mwleeds:
There is an edge case that's quite hard to support: if there's a checkpoint commit at or before a flatpak is auto-installed, and that flatpak is later auto-uninstalled, the update USB won't be able to provide the flatpak for a receiving computer that needs to follow the checkpoint. However, the USB also won't have the checkpoint itself, so it's a broader problem than we can fix here.

Probably another reason to avoid checkpoints unless truly necessary, for now.

mwleeds:
Checked that this script still behaves as expected with the new minor changes

eos-updater-prepare-volume/eos-updater-prepare-volume
```diff
@@ -150,6 +156,8 @@ class VolumePreparer:
             if action.type != \
                EosUpdaterUtil.FlatpakRemoteRefActionType.UNINSTALL:
                 autoinstalls.add(action.ref.ref.format_ref())
+            else:
+                autoinstalls.discard(action.ref.ref.format_ref())
```
wjt, line 160: I have never encountered `set.discard` before! I have always used:

```python3
                autoinstalls -= {action.ref.ref.format_ref()}
```

when I've wanted this, which has always felt clunky. Nice.

eos-updater-prepare-volume/eos-updater-prepare-volume
```diff
@@ -208,9 +209,18 @@ class VolumePreparer:
                         '--commit={}'.format(os_collection_ref[2]),
                         self.volume_path, os_collection_ref[0],
                         os_collection_ref[1]])
-        if len(all_flatpak_refs) > 0:
+        # For specified flatpaks, an error is fatal
+        for flatpak_ref in self.flatpak_refs:
+            self.__run(['flatpak', '--system', 'create-usb',
+                        self.volume_path] + [flatpak_ref],
```
wjt, line 215:
```suggestion
                        self.volume_path, flatpak_ref],
```

eos-updater-prepare-volume/eos-updater-prepare-volume
```diff
@@ -208,9 +209,18 @@ class VolumePreparer:
...
+                       fatal_error=True)
+        # For autoinstall flatpaks, an error is a warning. Some may have been
+        # manually uninstalled, and the drive can still be used to update some
+        # OS versions.
+        for flatpak_ref in autoinstall_flatpaks:
             self.__run(['flatpak', '--system', 'create-usb',
-                        self.volume_path] + all_flatpak_refs)
+                        self.volume_path] + [flatpak_ref],
```
wjt, line 222:
```suggestion
                        self.volume_path, flatpak_ref],
```

wjt (changes requested):
LGTM except for a pair of minor changes I missed before

mwleeds:
> > So, this seems like an improvement, but it's another nail in the box of "we should make e-u-f-i failures non-fatal" IMO.
> 
> I would tend to agree

Filed https://phabricator.endlessm.com/T31149

eos-updater-prepare-volume/eos-updater-prepare-volume
mwleeds: Yeah just discovered that too

PR #281, opened 2021-02-08
test-common: Use standalone ostree-trivial-httpd

dbnicholson:
The `trivial-httpd` builtin for the `ostree` command has been deprecated
for a couple years and is not included on default builds of ostree. On
the other hand, the standalone `ostree-trivial-httpd` has been available
since ostree 2017.1 in `<libexecdir>/libostree/ostree-trivial-httpd`.
Try to use that instead of the builtin.

https://phabricator.endlessm.com/T31384

### DIFF
--- a/test-common/meson.build
+++ b/test-common/meson.build
@@ -19,11 +19,27 @@ libeos_updater_test_common_headers = [
   'utils.h',
 ]
 
+# ostree-trivial-httpd is in libexecdir/libostree or
+# libexecdir/installed-tests/libostree if
+# https://github.com/ostreedev/ostree/pull/1633 is merged.
+# Unfortunately, but we don't know where ostree's libexecdir is. Try to
+# build a path for typical places.
+ostree_trivial_httpd = find_program(
+    'ostree-trivial-httpd',
+    join_paths(libexecdir, 'libostree/ostree-trivial-httpd'),
+    join_paths(libexecdir, 'installed-tests/libostree/ostree-trivial-httpd'),
+    '/usr/libexec/libostree/ostree-trivial-httpd',
+    '/usr/libexec/installed-tests/libostree/ostree-trivial-httpd',
+    '/usr/lib/libostree/ostree-trivial-httpd',
+    '/usr/lib/installed-tests/libostree/ostree-trivial-httpd',
+)
+
 libeos_updater_test_common_cppflags = [
   '-DFLATPAK_BINARY="@0@"'.format(find_program('flatpak').path()),
   '-DG_LOG_DOMAIN="eos-updater-test-common"',
   '-DGPG_BINARY="@0@"'.format(find_program('gpg').path()),
   '-DOSTREE_BINARY="@0@"'.format(find_program('ostree').path()),
+  '-DOSTREE_TRIVIAL_HTTPD_BINARY="@0@"'.format(ostree_trivial_httpd.path()),
   '-DOSTREE_WITH_AUTOCLEANUPS',
 ]
 
--- a/test-common/ostree-spawn.c
+++ b/test-common/ostree-spawn.c
@@ -28,6 +28,10 @@
 #error OSTREE_BINARY is not defined
 #endif
 
+#ifndef OSTREE_TRIVIAL_HTTPD_BINARY
+#error OSTREE_TRIVIAL_HTTPD_BINARY is not defined
+#endif
+
 static void
 copy_strv_to_ptr_array (const gchar * const *strv,
                         GPtrArray *array)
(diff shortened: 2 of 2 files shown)


test-common/meson.build
```diff
@@ -19,11 +19,21 @@ libeos_updater_test_common_headers = [
   'utils.h',
 ]
 
+# ostree-trivial-httpd is in libexecdir/libostree, but we don't know
+# where libexecdir is. Try to build a path for typical places.
+ostree_trivial_httpd = find_program(
+    'ostree-trivial-httpd',
+    join_paths(libexecdir, 'libostree/ostree-trivial-httpd'),
+    '/usr/libexec/libostree/ostree-trivial-httpd',
+    '/usr/lib/libostree/ostree-trivial-httpd'
```
pwithnall, line 28: Would be good to file an issue against OSTree and ask them to list their `libexecdir` in `ostree-1.pc`, then use that here and fall back to this lot if it’s not set.

pwithnall (commented):
You may want to comment on [this upstream PR](https://github.com/ostreedev/ostree/pull/1633) and point out that we use `ostree-trivial-httpd`. The PR’s a bit old, but they were proposing to make it builddir-only.

dbnicholson:
> You may want to comment on [this upstream PR](https://github.com/ostreedev/ostree/pull/1633) and point out that we use `ostree-trivial-httpd`. The PR’s a bit old, but they were proposing to make it builddir-only.

Thanks. Yeah, telling upstream that we no longer use it was a follow up. I already had to unbreak this case once in the past when someone else waded into this area

test-common/meson.build
dbnicholson: I was thinking about that but was a little concerned that `ostree-1.pc` really describes libostree and not the CLI or test programs. I.e., you could have `ostree-tests` installed without `libostree-dev` in Debian. Systemd has a similar issue where the pc files are split up and the one containing the unitdir setting is shipped in the main systemd package as opposed to the libsystemd-dev package. But I guess it would be better than nothing.

dbnicholson:
> > You may want to comment on [this upstream PR](https://github.com/ostreedev/ostree/pull/1633) and point out that we use `ostree-trivial-httpd`. The PR’s a bit old, but they were proposing to make it builddir-only.
> 
> Thanks. Yeah, telling upstream that we no longer use it was a follow up. I already had to unbreak this case once in the past when someone else waded into this area

I was looking at this again in the context of rebasing ostree. It actually would still be installed so long as you build with `--enable-installed-tests`, which debian does. In that case it would be in `$libexecdir/installed-tests/libostree/ostree-trivial-httpd` and part of the `ostree-tests` debian package. I think the point still stands about letting upstream know we use it.

Alternatively, I could just drop the dependency and use python's `http.server` which I've used many times in the past.

pwithnall:
> I was looking at this again in the context of rebasing ostree. It actually would still be installed so long as you build with `--enable-installed-tests`, which debian does. In that case it would be in `$libexecdir/installed-tests/libostree/ostree-trivial-httpd` and part of the `ostree-tests` debian package. I think the point still stands about letting upstream know we use it.

So should the `installed-tests` path be added to the search path for the httpd server?

dbnicholson:
> > I was looking at this again in the context of rebasing ostree. It actually would still be installed so long as you build with `--enable-installed-tests`, which debian does. In that case it would be in `$libexecdir/installed-tests/libostree/ostree-trivial-httpd` and part of the `ostree-tests` debian package. I think the point still stands about letting upstream know we use it.
> 
> So should the `installed-tests` path be added to the search path for the httpd server?

Sure. If the upstream PR ever goes in it would still find it.

test-common/meson.build
```diff
@@ -19,11 +19,21 @@ libeos_updater_test_common_headers = [
   'utils.h',
 ]
 
+# ostree-trivial-httpd is in libexecdir/libostree, but we don't know
+# where libexecdir is. Try to build a path for typical places.
+ostree_trivial_httpd = find_program(
+    'ostree-trivial-httpd',
+    join_paths(libexecdir, 'libostree/ostree-trivial-httpd'),
```
pwithnall, line 26: Unless I’m misinterpreting [your comment](https://github.com/endlessm/eos-updater/pull/281#issuecomment-805113627), I think this needs an additional `join_paths(libexecdir, 'installed-tests/libostree/ostree-trivial-httpd'),` entry then.

dbnicholson: Sorry, I was a bit terse. I'll update with that location so that it works if/when upstream moves it there.

pwithnall: Seems easier to me to add the entry now, then we never have to think about it again, but LGTM either way. Merge whenever you like :)

dbnicholson: Poor communication on my part. It was my intention to update the PR to reflect the new potential future path. I just hadn't gotten around to it. Updated now.

pwithnall (approved):

PR #282, opened 2021-03-03
e-u-f-i: Unset RemainAfterExit on -fallback.service

wjt:
Quoth systemd.timer(5):

> […] services with RemainAfterExit= set […] are usually not suitable
> for activation via repetitive timers, as they will only be activated
> once, and then stay around forever.

And indeed before this patch,
eos-updater-flatpak-installer-fallback.timer would only fire once
(assuming the .service succeeds) because the .service is considered to
be still running after it exits.

https://phabricator.endlessm.com/T31564

### DIFF
--- a/eos-updater-flatpak-installer/eos-updater-flatpak-installer-fallback.service.in
+++ b/eos-updater-flatpak-installer/eos-updater-flatpak-installer-fallback.service.in
@@ -15,7 +15,6 @@ Conflicts=shutdown.target
 
 [Service]
 Type=oneshot
-RemainAfterExit=true
 ExecStart=@libexecdir@/eos-updater-flatpak-installer --pull --mode=perform
 Restart=no
 


wjt:
Tested by adding an override:

```
# /etc/systemd/system/eos-updater-flatpak-installer-fallback.service.d/override.conf
[Service]
RemainAfterExit=false
```

and verifying that the timer still planned to fire – but I confess that I didn't wait for the first timer to actually expire. I'll un-draft this tomorrow once that's happened "organically".

wjt:
```
NEXT                        LEFT         LAST                        PASSED       UNIT                                         ACTIVATES                                     
Fri 2021-03-05 09:49:45 GMT 22h left     Thu 2021-03-04 09:30:15 GMT 1h 55min ago eos-updater-flatpak-installer-fallback.timer eos-updater-flatpak-installer-fallback.service
```

System seems to work

pwithnall (approved):

PR #283, opened 2021-06-04
T31918 Teach eos-updater not to cross eos3a→eos4 checkpoint on certain systems

pwithnall:
See the commit messages.

This adds a load of checks for systems which should not be migrated to `eos4`. The checks are based on searching around tasks and code for ways to distinguish these systems, rather than actual testing on these systems. The basis for the checks still needs to be checked against the actual systems.

### DIFF
--- a/eos-updater/poll-common.c
+++ b/eos-updater/poll-common.c
@@ -26,12 +26,16 @@
 
 #include <eos-updater/object.h>
 #include <eos-updater/poll-common.h>
+#include <gio/gunixmounts.h>
+#include <glib.h>
+#include <glib/gi18n.h>
 #include <libeos-updater-util/metrics-private.h>
 #include <libeos-updater-util/ostree-util.h>
 #include <libeos-updater-util/util.h>
 #include <libsoup/soup.h>
 #include <ostree.h>
 #include <string.h>
+#include <sys/utsname.h>
 
 #ifdef HAS_EOSMETRICS_0
 #include <eosmetrics/eosmetrics.h>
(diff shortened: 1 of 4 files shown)


tests/test-update-refspec-checkpoint.c
```diff
@@ -1389,6 +1389,214 @@ test_update_refspec_checkpoint_ignore_remote (EosUpdaterFixture *fixture,
...
+      "model		: 142\n"
+      "model name	: Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz\n"
+      "stepping	: 12\n";
+      /* etc */
+  const gchar *cpuinfo_not_i8565u =
+      "processor	: 0\n"
+      "vendor_id	: NotIntel\n"
+      "cpu family	: 6\n"
+      "model		: 123\n"
+      "model name	: Some collection of ASICs\n"
```
dbnicholson, line 1418: :laughing:

dbnicholson (changes requested):
Looks good to me, but the tests are failing. Two of them seem related to flatpak changes, but the third is from here:

```
(/build/src/_build/tests/../eos-updater/eos-updater:10116): GLib-GIO-CRITICAL **: 16:13:53.809: g_unix_mount_compare: assertion 'mount1 != NULL && mount2 != NULL' failed
```

I don't understand why the `GUnixMountEntry` would be `NULL`. This probably has something to do with the chroot setup in jenkins.

eos-updater/poll-common.c
```diff
@@ -383,9 +387,206 @@ get_booted_refspec (OstreeDeployment     *booted_deployment,
...
+{
+  g_autoptr(GUnixMountEntry) extra_mount = NULL, root_mount = NULL;
+
+  extra_mount = g_unix_mount_for ("/var/endless-extra", NULL);
+  root_mount = g_unix_mount_for ("/", NULL);
+
+  if (g_strcmp0 (g_getenv ("EOS_UPDATER_TEST_IS_SPLIT_DISK"), "1") == 0)
+    return TRUE;
+
+  return (g_unix_mount_compare (extra_mount, root_mount) != 0);
```
dbnicholson, line 404: I think this would be more straightforward to use `g_unix_mount_at ("/var/endless-extra", NULL)` and check if the result is `NULL` or not. I actually think this would produce false positives on non-split Endless systems, too. `/var` is a bind mount, so the entry returned by `g_unix_mount_for` will be `/var` in the non-split case, which will not match the mount at `/`. I'm pretty sure all systems will be regarded as split disk systems.

pwithnall: > I don't understand why the `GUnixMountEntry` would be `NULL`. This probably has something to do with the chroot setup in jenkins.

I neglected to consider the case where either of the mount entries were `NULL`. It didn’t help that [GLib didn’t document this](https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2145) until an introspection annotation was added recently.

Your suggestion is good though, and avoids a load of faffing with handling different things possibly being `NULL`.

pwithnall:
Currently wrestling with the flatpak test failures, which I think are unrelated but we should fix anyway. So far, all I’ve been able to determine is that it’s something to do with the new summary code in flatpak not finding a summary for the `ostree-metadata` branch, and therefore being unhappy. Consequently, I am also unhappy. Will continue to sink life energy into this tomorrow.

pwithnall:
See https://github.com/endlessm/flatpak/pull/268 for the flatpak test failures. I don’t think the fix requires changes in eos-updater.

dbnicholson:
Changes look good now. I'll wait until the flatpak build is completed and kick the CI here again.

dbnicholson:
retest this please

dbnicholson:
The CI is running again with the updated flatpak:
```
Get:73 http://obs-master.endlessm-sf.com:82/shared/eos master/endless amd64 flatpak amd64 1.10.2+dev35.5b28741-0bem1 [1281 kB]
```

dbnicholson:
Still hitting an issue in the flatpak tests. I'm not sure if the issue is related, but if you want to split that off to a separate task, I'm pretty sure these changes are working correctly.

pwithnall:
Rebased and fixup squashed in. Let’s see if this passes the tests now that #285 is in.

pwithnall:
woohoo, tests passed :partying_face:

dbnicholson (approved):
Woot!

PR #284, opened 2021-06-14
tests: Drop test for collection ID not being set on the remote server

pwithnall:
Since flatpak has almost entirely dropped collection ID support, this
test is no longer relevant.

It was originally added in T20812.

### DIFF
--- a/tests/test-update-install-flatpaks.c
+++ b/tests/test-update-install-flatpaks.c
@@ -1178,85 +1178,6 @@ test_update_install_flatpaks_in_repo_error_no_branch_name (EosUpdaterFixture *fi
   g_assert_false (g_strv_contains ((const gchar * const *) flatpaks_in_repo, flatpaks_to_install[0].app_id));
 }
 
-/* Insert a list of flatpaks to automatically install on the commit, this
- * time with a collection-id specified, a collection-id is configured
- * on the remote config, but the collection-id is not set up on the remote
- * end. This is an invalid configuration and should fail. */
-static void
-test_update_install_flatpaks_in_repo_error_if_collection_invalid (EosUpdaterFixture *fixture,
-                                                                  gconstpointer      user_data)
-{
-  g_auto(EtcData) real_data = { NULL, };
-  EtcData *data = &real_data;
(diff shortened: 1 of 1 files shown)


pwithnall:
Sigh, tests still fail but in a different way this time. Looking into it.

pwithnall:
I think the latest test failure is caused by https://github.com/flatpak/flatpak/pull/4316, which is an additional fix related to the previous set of flatpak fixes I backported. I’ll comment on the Phab task to tie this all together.

dbnicholson:
Merged the flatpak backport in https://github.com/endlessm/flatpak/pull/269. I'll wait until that gets built and restart this test.

dbnicholson:
I was curious about how the collection ID was handled with the new index summary partially for a personal preview of what will happen when a newer flatpak is deployed to the server and starts generating the new indexed summaries.

It appears that the collection ID is only validated when the [full summary is fetched](https://github.com/endlessm/flatpak/blob/master/common/flatpak-dir.c#L12172) (`state->summary` is only set when the full summary is downloaded). I think that the eos-updater tests only generate the full summary, so I'm not sure why this doesn't fail.

dbnicholson:
retest this please

dbnicholson:
Oh, it's because the tests use `flatpak build-export`, which by default updates the summary, which in turn generates `summary.idx`. So, it is indeed not fetching the full summary. Incidentally, the tests make another OS commit (to ship the autoinstall.json update?) after the flatpak is built, which then regenerates a summary with `ostree summary --update`.

I think the tests would be a little more robust if they used `flatpak build-update-repo` instead of `ostree summary --update` when the repo contained flatpaks so that the summary data is consistent. This would also allow passing `--no-summary-index` (which removes stale `summary.idx` files), which would allow being explicit about whether a full summary or indexed summary would be fetched by the client.

dbnicholson:
Well, since our server will continue to use full summaries for some time, removing this test does hide a bug. I created #285 to continue simulating the behavior with older flatpak on the server.

pwithnall:
Closing strongly in favour of #285

PR #285, opened 2021-06-17
tests: Remove indexed summary for invalid collection ID test

dbnicholson:
With newer flatpak, an indexed summary file will be created on the
server and the client will try to download `summary.idx` first. This
breaks the test that expects to fail when there's no collection ID in
the remote summary since flatpak only validates its presence on a full
`summary` file.

Since the Endless servers aren't providing indexed summaries yet, it's
useful to continue simulating this test with a full summary that doesn't
contain a collection ID. If the summary index exists, delete it to go
back to older flatpak behavior.

### DIFF
--- a/tests/test-update-install-flatpaks.c
+++ b/tests/test-update-install-flatpaks.c
@@ -33,6 +33,7 @@
 #include <flatpak.h>
 #include <json-glib/json-glib.h>
 
+#include <glib/gstdio.h>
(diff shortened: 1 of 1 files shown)


dbnicholson:
An alternative to #284. It passed that particular test for me locally but a later one failed. Let's see if the full test suite passes.

pwithnall (approved):

PR #286, opened 2021-06-24
eos-updater: Tweak eos3a → eos4 checkpoint tests on aarch64 hosts

pwithnall:
Since the checkpoint redirect should not be followed on aarch64 hosts,
the tests are going to behave slightly differently on that architecture:
we should expect the checkpoint to never be followed unless the
architecture is mocked, or following redirects is forced.

The alternative approach would be to always mock the architecture, and
set it to something other than aarch64 (unless explicitly testing the
architecture checks), but then we’d be throwing away an opportunity to
test the checkpoint code on exactly the platform it’s meant to catch.

### DIFF
--- a/tests/test-update-refspec-checkpoint.c
+++ b/tests/test-update-refspec-checkpoint.c
@@ -27,6 +27,7 @@
 
 #include <gio/gio.h>
 #include <locale.h>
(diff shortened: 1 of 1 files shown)


dbnicholson (approved):
I think that works.

PR #289, opened 2021-08-18
Backport eos3a → eos4 conditional checkpoint work

dbnicholson:
These are the 3 commits representing the conditional checkpoint implementation and tests. On my local machine none of the flatpak fixes were needed since 3.9 is still using flatpak 1.8. We'll see if Jenkins agrees.


pwithnall (approved):
:clap:

PR #290, opened 2021-08-20
eos-updater: Fix ref matching for eos3a → eos4 checkpoint tests

dbnicholson:
The eos3a → eos4 checkpoint tests were only being applied when the refs
were literally `eos3a` and `eos4`. Unfortunately, our refs actually look
like `os/eos/<arch>/<branch>`. Expand the matches so they also accept
`/<branch>` at the end of the ref.


pwithnall (approved):
Oooops, sorry

PR #291, opened 2021-08-20
eos-updater: Mock architecture for eos3a → eos4 ref matching tests

dbnicholson:
When the test host is `aarch64`, the expected results of these tests are
effectively ignored unless architecture itself is being mocked. See
17e9b623bac9af8fd34eb593cc238a89327ea9c4 for details.

This caused a problem for the ref matching tests when run on `aarch64`.
These cases set a condition that would cause the checkpoint to be
ignored and then use different refs to check if the condition is
actually used. When the host arch is `aarch64`, the test later decides
that the expected result is an ignored checkpoint regardless of what the
test case said the expected result is.

To workaround this, change the ref matching cases to override the
architecture to `aarch64`. This prepares the checkpoint to be ignored
when the refs match as it was previously using the split disk override
for. However, now it skips past the condition that adjusts the expected
outcome if the host itself is `aarch64`.


dbnicholson:
This time I actually did a scratch test on OBS and saw that it succeeded for aarch64 (https://obs-master.endlessm-sf.com/package/show/home:dbn:branches:eos:3.9:endless/eos-updater).

jprvita (approved):

PR #292, opened 2021-09-24
eos-updater: Fix for upgrading to eos4 and newer's ref

starnight:
Users may upgrading system from booted ref eos3a to both eos4, or even
newer target ref in the future. The newer target ref is defined as
"latest1" in T32550.

This PR allows eos3.9 nexthw upgrading to eos4 or newer ref as well.

https://phabricator.endlessm.com/T32552
https://phabricator.endlessm.com/T32542


eos-updater/poll-common.c
```diff
@@ -514,8 +514,9 @@ should_follow_checkpoint (OstreeSysroot     *sysroot,
   g_autoptr(GHashTable) hw_descriptors = NULL;
   const gchar *sys_vendor, *product_name;
   gboolean is_eos3a_to_eos4 =
```
pwithnall, line 516: Since we’re now allowing transitions which aren’t `eos3a` → `eos4`, and disallowing transitions which are `eos3a` → other things, I think this variable should be renamed to cover its new meaning. Perhaps `is_conditional_upgrade_path`?

pwithnall, line 516: It would also be useful to add a comment above this variable like:
```
/* https://phabricator.endlessm.com/T32542, https://phabricator.endlessm.com/T32552 */
```
to allow the reasoning for the choice of branch names/prefixes to be looked up later easily.

pwithnall:
Once you’ve got the tests passing, I think it would add confidence to add a couple of new test entries in the `tests` array in `test_update_refspec_checkpoint_eos3a_eos4()` which check the `os/eos/arm64/eos3a` → `os/eos/arm64/latest1` transition (for [T32552](https://phabricator.endlessm.com/T32552)) and the `os/eos/nexthw/eos3.9` → `os/eos/amd64/eos4` transition (for [T32542](https://phabricator.endlessm.com/T32542)).

eos-updater/poll-common.c
starnight: sure!

starnight: good point!  Thanks

eos-updater/poll-common.c
```diff
@@ -571,13 +573,23 @@ should_follow_checkpoint (OstreeSysroot     *sysroot,
...
     {
       *out_reason = g_strdup (_("Read-only systems are not supported in EOS 4."));
       return FALSE;
     }
 
+  g_message("booted_ref=%s, target_ref=%s", booted_ref, target_ref);
+  /* https://phabricator.endlessm.com/T32542 */
+  if (is_conditional_upgrade_path &&
+      g_str_has_suffix (booted_ref, "nexthw/eos3.9") &&
+      g_str_has_suffix (target_ref, "eos3a"))
```
wjt, line 587: I don't think this check is necessary. We will not be publishing a redirect from nexthw/eos3.9 to eos3a.

starnight: ok

tests/test-update-refspec-checkpoint.c
```diff
@@ -1614,10 +1617,13 @@ test_update_refspec_checkpoint_eos3a_eos4 (EosUpdaterFixture *fixture,
...
       g_assert_no_error (error);
 
+      test_is_aarch64 = (g_strcmp0 (tests[i].uname_machine, "aarch64") == 0);
+      src_is_eos3a = (g_strcmp0 (tests[i].src_ref, "eos3a") == 0);
       if (tests[i].expect_checkpoint_followed &&
           (!host_is_aarch64 ||
            tests[i].force_follow_checkpoint ||
-           tests[i].uname_machine != NULL))
+           !test_is_aarch64 ||
+           (test_is_aarch64 && !src_is_eos3a)))
```
pwithnall, line 1631: I don’t understand the reasoning for changing the conditions here, or introducing `test_is_aarch64` or `src_is_eos3a`. Could you explain the reasoning please?

starnight: I was confused by the test's expected result for the aarch64 part when I saw and tuned the test plans at first time.  After you ask this question, I realize the expected result is already controlled by **expect_checkpoint_followed** and I I have already modified it in the test plans.
So, changed the check condition back.

pwithnall (approved):

PR #295, opened 2021-12-02
build: remove 'version' parameter from static libraries

abderrahim:
it is invalid and fails to build with meson 0.60


pwithnall (approved):
Thanks :)

PR #296, opened 2021-12-15
Add prune-ref flatpak-installer action

dbnicholson:
The purpose of this PR is to add an action for the flatpak installer to cleanup dangling OSTree refs for uninstalled Flatpaks. This is primarily for cleaning up from bugs here or in other software (the image builder).

The PR could be split up some if preferred. The first 4 commits are entirely independent things I came across when making tests. The last commit could be broken up by component. It also might need a couple more tests.

https://phabricator.endlessm.com/T32791

### DIFF
--- a/eos-updater-flatpak-installer/eos-updater-autoinstall.schema.json
+++ b/eos-updater-flatpak-installer/eos-updater-autoinstall.schema.json
@@ -88,6 +88,38 @@
                 }
             },
             "required": ["ref-kind", "name", "branch"]
+        },
+        {
+            "properties": {
+                "action": {
+                    "const": "prune-ref",
+                    "description": "The machine-readable type of event",
+                    "type": "string"
+                },
+                "serial": {
+                    "$ref": "#/definitions/serial"
+                },
+                "filters": {
+                    "$ref": "#/definitions/filters"
+                },
+                "ref-kind": {
+                    "$ref": "#/definitions/ref-kind"
+                },
+                "name": {
+                    "description": "The app-id to be pruned",
+                    "$ref": "#/definitions/name"
+                },
+                "branch": {
+                    "$ref": "#/definitions/branch"
+                },
+                "collection-id": {
+                    "$ref": "#/definitions/collection-id"
+                },
+                "remote": {
+                    "$ref": "#/definitions/remote"
+                }
+            },
+            "required": ["ref-kind", "name", "branch", "collection-id", "remote"]
         }
     ],
     "required": ["action", "serial"],
--- a/libeos-updater-flatpak-installer/perform-flatpak-actions.c
+++ b/libeos-updater-flatpak-installer/perform-flatpak-actions.c
@@ -32,6 +32,42 @@
 #include <string.h>
 
 static gboolean
+validate_collection_id_remote (FlatpakInstallation  *installation,
+                               const gchar          *collection_id,
+                               const gchar          *expected_remote_name,
+                               GError              **error)
+{
+  g_autofree gchar *candidate_remote_name = NULL;
+  g_assert (collection_id != NULL);
+  g_assert (expected_remote_name != NULL);
+
+  g_message ("Finding remote name for %s", collection_id);
+
+  /* Ignore errors here. We always have the @expected_remote_name to use. */
+  candidate_remote_name = euu_lookup_flatpak_remote_for_collection_id (installation,
+                                                                       collection_id,
+                                                                       NULL);
+
+  if (candidate_remote_name != NULL &&
+      g_strcmp0 (expected_remote_name, candidate_remote_name) != 0)
+    {
+      g_set_error (error,
+                   EOS_UPDATER_ERROR,
+                   EOS_UPDATER_ERROR_FLATPAK_REMOTE_CONFLICT,
+                   "Specified flatpak remote ‘%s’ conflicts with the remote "
+                   "detected for collection ID ‘%s’ (‘%s’), cannot continue.",
+                   expected_remote_name,
+                   collection_id,
+                   candidate_remote_name);
+      return FALSE;
+    }
+
+  g_message ("Remote name for %s is %s", collection_id, expected_remote_name);
+
+  return TRUE;
+}
+
+static gboolean
 try_update_application (FlatpakInstallation       *installation,
                         FlatpakRef                *ref,
                         EosUpdaterInstallerFlags   flags,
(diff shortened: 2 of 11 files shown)


eos-updater-flatpak-installer/eos-updater-autoinstall.schema.json
```diff
@@ -88,6 +88,38 @@
                 }
             },
             "required": ["ref-kind", "name", "branch"]
+        },
+        {
+            "properties": {
+                "action": {
+                    "const": "prune-ref",
```
wjt, line 95: @pwithnall would know better, but I am not sure that you can cleanly add new actions like this in a backwards-compatible way without an OS checkpoint. The actions from the OS update being deployed are parsed and prepared by the **old** eos-updater version.

I copied the example from the manpage to my local system, then did a dry run:

```
wjt@camille:~$ cat /var/lib/eos-application-tools/flatpak-autoinstall.d/99-prune-ref.json 
[
    {
        "action": "prune-ref",
        "branch": "stable",
        "serial": 2017100300,
        "ref-kind": "app",
        "name": "org.example.UninstalledApp",
        "collection-id": "org.example.Apps",
        "remote": "example-apps"
    }
]
wjt@camille:~$ /usr/libexec/eos-updater-flatpak-installer --dry-run --mode check 
eos-updater-flatpak-installer-Message: 09:21:29.098: Running in mode ‘check’

(eos-updater-flatpak-installer:4989): libeos-updater-util-WARNING **: 09:21:29.099: Skipping the following actions while parsing ‘99-prune-ref.json’, due to not supporting their contents. They will not be reapplied later; the system may be in an inconsistent state from this point forward.
{"action":"prune-ref","branch":"stable","serial":2017100300,"ref-kind":"app","name":"org.example.UninstalledApp","collection-id":"org.example.Apps","remote":"example-apps"}
[...]
```

I then tried installing an OS update with this file still in place. To my surprise the OS update was deployed despite this warning:

```
Dec 15 09:23:12 camille eos-updater[1915]: Skipping the following actions while parsing ‘99-prune-ref.json’, due to not supporting their contents. They will not be reapplied later; the system may be in an inconsistent state from this point forward.
                                           {"action":"prune-ref","branch":"stable","serial":2017100300,"ref-kind":"app","name":"org.example.UninstalledApp","collection-id":"org.example.Apps","remote":"example-apps"}
```

So it looks like I am wrong and in fact the action would be applied after booting into the new OS version which supports this feature.

dbnicholson: You might still be right (there's a big comment in the code about it), but this action isn't used in the pre-upgrade part by eos-updater since there's no fetching involved.

wjt: I think we would need to thoroughly test whether an unrecognised action at the prepare phase would prevent:

- subsequent actions in the same file being prepared (I think yes)
- actions in subsequently-named files being prepared (?)

libeos-updater-flatpak-installer/perform-flatpak-actions.c
```diff
@@ -69,44 +105,23 @@ try_update_application (FlatpakInstallation       *installation,
...
-                       collection_id,
-                       candidate_remote_name);
-          return FALSE;
-        }
+  g_assert (remote_name != NULL);
 
-      g_message ("Remote name for %s is %s", collection_id, remote_name);
-    }
+  if (collection_id != NULL &&
+      !validate_collection_id_remote (installation,
```
pwithnall, line 120: Nitpick: The branch would be a bit clearer if the `validate_collection_id_remote()` function was split out in a preparatory commit. No point doing that now, though, since the review is underway :)

libeos-updater-flatpak-installer/perform-flatpak-actions.c
```diff
@@ -397,10 +452,60 @@ check_if_flatpak_is_installed (FlatpakInstallation        *installation,
...
+  g_autofree gchar *formatted_ref = NULL;
+  g_autofree gchar *refspec = NULL;
+  g_autofree gchar *rev = NULL;
+
+  if (!ostree_repo_open (repo, NULL, error))
+    return FALSE;
+
+  formatted_ref = flatpak_ref_format_ref (action->ref->ref);
+  refspec = g_strdup_printf ("%s:%s", action->ref->remote, formatted_ref);
+  if (!ostree_repo_resolve_rev_ext (repo,
```
pwithnall, line 497: Should this be taking the collection ID into account too? Perhaps with `ostree_repo_resolve_collection_ref()`?

libeos-updater-util/flatpak-util.c
```diff
@@ -1356,26 +1430,12 @@ squash_ref_actions_ptr_array (GPtrArray *ref_actions  /* (element-type EuuFlatpa
...
+      EuuFlatpakRemoteRefAction *existing_action_for_ref =
+        g_hash_table_lookup (hash_table, action->ref->ref);
 
-      if (action->type == EUU_FLATPAK_REMOTE_REF_ACTION_INSTALL ||
-          action->type == EUU_FLATPAK_REMOTE_REF_ACTION_UNINSTALL ||
-          existing_action_for_ref == NULL ||
-          existing_action_for_ref->type == EUU_FLATPAK_REMOTE_REF_ACTION_UPDATE)
+      if (existing_action_for_ref == NULL ||
+          (ref_action_type_priority (action->type) >=
+           ref_action_type_priority (existing_action_for_ref->type)))
```
pwithnall, line 1438: aiui this logic will squash `[uninstall org.gnome.Foobar, prune-ref org.gnome.Foobar]` into just `[uninstall org.gnome.Foobar]`, dropping the prune-ref. Is that what we want? Are we sure that uninstalls will now always prune?

libeos-updater-util/flatpak-util.c
```diff
@@ -1844,6 +1906,8 @@ find_related_refs_for_action (FlatpakInstallation       *installation,
         if (!flatpak_transaction_add_update (transaction, ref_action_ref, NULL, NULL, error))
           return FALSE;
         break;
+      case EUU_FLATPAK_REMOTE_REF_ACTION_PRUNE_REF:
+        break;
```
pwithnall, line 1910: Suggest adding a short comment in here to make it clear this case isn’t accidentally unimplemented.

libeos-updater-util/flatpak-util.c
```diff
@@ -2671,3 +2744,64 @@ euu_flatpak_transaction_uninstall (FlatpakInstallation *installation,
 
   return transaction_run_single_op (transaction, cancellable, error);
 }
+
+
```
pwithnall, line 2748: Nitpick: Double empty line should just be a single empty line

libeos-updater-util/flatpak-util.c
```diff
@@ -2671,3 +2744,64 @@ euu_flatpak_transaction_uninstall (FlatpakInstallation *installation,
...
+gboolean
+euu_flatpak_transaction_prune_ref (FlatpakInstallation  *installation,
+                                   const gchar          *remote,
+                                   const gchar          *formatted_ref,
+                                   gboolean              no_prune,
```
pwithnall, line 2753: Could do with a brief comment (maybe formatted as a gtk-doc comment for the function) which explains what `no_prune` means in the context of a function whose name suggests that the only thing it does is to prune.

libeos-updater-util/tests/flatpak-util.c
```diff
@@ -344,6 +344,49 @@ test_no_compress_install_install_different_branches (void)
...
+  FlatpakToInstallDirectory directory = { files, G_N_ELEMENTS (files) };
+  g_autoptr(GHashTable) uncompressed_ref_actions_table = flatpak_to_install_directory_to_hash_table (&directory);
+  g_autoptr(GPtrArray) flattened_actions_list = euu_flatten_flatpak_ref_actions_table (uncompressed_ref_actions_table);
+
+  g_assert_cmpuint (flattened_actions_list->len, ==, 1);
+  g_assert_cmpint (((EuuFlatpakRemoteRefAction *) g_ptr_array_index (flattened_actions_list, 0))->type, ==,
+                   EUU_FLATPAK_REMOTE_REF_ACTION_INSTALL);
+}
+
+/* Test that no compresson occurs if 'install' and 'prune-ref' are on
```
pwithnall, line 367: Typo: ‘compression’

eos-updater-flatpak-installer/eos-updater-autoinstall.schema.json
pwithnall: I can’t remember what the code was meant to do (sorry), so defer to the big comment and (more importantly) testing of whether this actually works.

PR #299, opened 2022-02-08
Don't use volatile qualifier for quark

mwleeds:
It is not needed per the g_dbus_error_register_error_domain() docs.


pwithnall:
ok to test

PR #300, opened 2022-02-16
Revert "apply: Use staged deployment on booted systems"

dbnicholson:
This reverts commit 12d263bfd3ee9e1e44a8d3704088edd038fd8883. On systems
such as PAYG where `/boot` is an automount,
`ostree-finalize-staged.service` fails to work correctly if the
automount expires before shutdown. Until a solution to that issue is
found, go back to the non-staged deployments we've used for years.

https://phabricator.endlessm.com/T5658
https://github.com/systemd/systemd/issues/22528
https://github.com/ostreedev/ostree/issues/2543

### DIFF
--- a/eos-updater/apply.c
+++ b/eos-updater/apply.c
@@ -241,7 +241,6 @@ apply_internal (ApplyData     *apply_data,
   g_autoptr(GKeyFile) origin = NULL;
   g_autoptr(OstreeSysroot) sysroot = NULL;
   const gchar *osname = get_test_osname ();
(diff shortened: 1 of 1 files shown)


dbnicholson:
I'm proposing that we only revert on eos4.0 and not on master. Also, I left the sysroot autocleanup change in since it has no effect if staged deployments aren't in use.

jprvita (approved):

wjt (approved):

PR #301, opened 2022-03-03
apply: Don't use staged deployments when /boot is automounted

dbnicholson:
The ostree staged deployment process works by waiting until shutdown to
swap the `/boot` symlinks to make the new deployment the default.
However, when `/boot` is the EFI System Partition and there's no `fstab`
entry, `systemd-gpt-auto-generator` sets up an automount so that the
VFAT filesystem is only exposed when needed.

Unfortunately, there are 2 bugs that make this process very fragile:

* Once a systemd automount unit is scheduled to be stopped, it ignores
  notifications from autofs that the target filesystem should be
  mounted. Therefore, if `/boot` isn't mounted when shutdown begins,
  `ostree admin finalize-staged` will fail. See
  https://github.com/systemd/systemd/issues/22528.

* autofs is not mount namespace aware, so it will begin the expiration
  timer for a mount unit unless a process in the root namespace is
  keeping it active. Since `ostree admin finalize-staged` is run from a
  mount namespace (either via systemd or its own to ensure `/sysroot`
  and `/boot` are mounted read-write), the automount daemon (systemd)
  will try to unmount the filesystem if it expires during this process.
  See https://bugzilla.redhat.com/show_bug.cgi?id=2056090.

Therefore, if `/boot` is an autofs filesystem, use a full deployment
instead of a staged deployment. Since systems with an automounted
`/boot` are not common, we want to retain the benefit of staged
deployments for more normal systems. See
https://github.com/ostreedev/ostree/issues/2543 for potential future
fixes in ostree.

https://phabricator.endlessm.com/T33136

### DIFF
--- a/eos-updater/apply.c
+++ b/eos-updater/apply.c
@@ -270,11 +270,13 @@ apply_internal (ApplyData     *apply_data,
 
   origin = ostree_sysroot_origin_new_from_refspec (sysroot, update_refspec);
 
-  /* When booted into an OSTree system, stage the deployment so that the
-   * /etc merge happens during shutdown. Otherwise (primarily the test
-   * suite), deploy the finalized tree immediately.
+  /* When booted into an OSTree system without an automount /boot, stage the
+   * deployment so that the /etc merge happens during shutdown. Otherwise
+   * (primarily sd-boot and the test suite), deploy the finalized tree
+   * immediately.
    */
-  staged_deploy = ostree_sysroot_is_booted (sysroot);
+  staged_deploy = ostree_sysroot_is_booted (sysroot) &&
+    !eos_updater_sysroot_boot_is_automount (sysroot, NULL);
   if (staged_deploy)
     {
       g_message ("Creating staged deployment for revision %s", update_id);


dbnicholson:
I haven't tested this yet but wanted to get it out there. I have another version that instead checks if `/boot` is VFAT, but I think this is better since automounting is the real problem, not VFAT.

eos-updater/apply.c
```diff
@@ -227,6 +228,65 @@ update_remote_branches (OstreeRepo   *repo,
...
+  if (cache == NULL)
+    {
+      g_warning ("Failed to create libmount cache: %s", g_strerror (errno));
+      goto out;
+    }
+  mnt_table_set_cache (tb, cache);
+  mnt_unref_cache (cache);
+
+  /* Find the /boot mountpoint iterating forward so that the autofs mount is
+   * found and not the automount target filesystem.
```
jprvita, line 270: It took me a little bit to understand how this works -- perhaps you could explain it a bit more in the comment? Maybe something like:
```suggestion
  /* Find the /boot mountpoint iterating forward so that the autofs mount is
   * found and not the automount target filesystem, as the autofs mount is
   * listed first in mountinfo.
```

jprvita (approved):
I am not familiar with libmount's API, but reading through these changes and the API docs, this PR makes sense and looks correct.

I left a small suggestion to improve the comment in the code, feel free to take it or not.

eos-updater/apply.c
dbnicholson: :+1:

dbnicholson:
I tested this on a PAYG system and could see that it did a finalized deployment instead of a staged deployment. I ended up splitting out the method to the util library so I could unit test it.

jprvita:
> I tested this on a PAYG system and could see that it did a finalized deployment instead of a staged deployment. I ended up splitting out the method to the util library so I could unit test it.

The new changes look good to me -- not sure if you want an extra pair of eyes on this, since this is critical code. In any case, master PAYG systems are currently failing to boot, so we won't be able to test it on ostrees until that is fixed.

dbnicholson:
I think @pwithnall is too deep in gnome-software land to take a look, so I'm going to go ahead and merge.

pwithnall (commented):
I just took a quick look and can’t see anything wrong with this, but I haven’t thought deeply about the consequences of the changes. The tests are very pleasing

PR #304, opened 2022-08-04
eos-updater: Block Weibu F3C updates to EOS 5 from EOS 4

starnight:
Block Weibu F3C (Endless EE-200) updates to EOS 5 from EOS 4. It is not
on the EOS 5 support list. Add corresponding tests as well.

https://phabricator.endlessm.com/T33311

### DIFF
--- a/eos-updater/poll-common.c
+++ b/eos-updater/poll-common.c
@@ -450,7 +450,7 @@ booted_system_has_i8565u_cpu (void)
 }
 
 /* Check @sys_vendor/@product_name against a list of systems which are no longer
- * supported. */
+ * supported since EOS 4. */
 static gboolean
 booted_system_is_unsupported_by_eos4_kernel (const gchar *sys_vendor,
                                              const gchar *product_name)
@@ -469,6 +469,33 @@ booted_system_is_unsupported_by_eos4_kernel (const gchar *sys_vendor,
       { "Acer", "Veriton Z6860G" },
       { "ASUSTeK COMPUTER INC.", "Z550MA" },
       { "Endless", "ELT-JWM" },
+      { "Endless", "EE-200" },
+    };
+
+  for (gsize i = 0; i < G_N_ELEMENTS (no_upgrade_systems); i++)
+    {
+      if (g_str_equal (sys_vendor, no_upgrade_systems[i].sys_vendor) &&
+          g_str_equal (product_name, no_upgrade_systems[i].product_name))
+        return TRUE;
+    }
+
+  return FALSE;
+}
+
+/* Check @sys_vendor/@product_name against a list of systems which are no longer
+ * supported since EOS 5. */
+static gboolean
+booted_system_is_unsupported_by_eos5_kernel (const gchar *sys_vendor,
+                                             const gchar *product_name)
+{
+  const struct
+    {
+      const gchar *sys_vendor;
+      const gchar *product_name;
+    }
+  no_upgrade_systems[] =
+    {
+      { "Endless", "EE-200" },
     };
 
   for (gsize i = 0; i < G_N_ELEMENTS (no_upgrade_systems); i++)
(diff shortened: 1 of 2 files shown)


eos-updater/poll-common.c
```diff
@@ -513,11 +514,15 @@ should_follow_checkpoint (OstreeSysroot     *sysroot,
...
-  /* https://phabricator.endlessm.com/T32542, https://phabricator.endlessm.com/T32552 */
+  /* https://phabricator.endlessm.com/T32542,
+   * https://phabricator.endlessm.com/T32552,
+   * https://phabricator.endlessm.com/T33311 */
   gboolean is_conditional_upgrade_path =
     (g_str_equal (booted_ref, "eos3a") ||
      g_str_has_suffix (booted_ref, "/eos3a") ||
-     g_str_has_suffix (booted_ref, "nexthw/eos3.9"));
+     g_str_has_suffix (booted_ref, "nexthw/eos3.9") ||
+     g_str_has_suffix (booted_ref, "latest1") ||
```
dbnicholson, line 524: I think this should be `/latest1`. We wouldn't want it to match something like `/foolatest1`.

eos-updater/poll-common.c
```diff
@@ -513,11 +514,15 @@ should_follow_checkpoint (OstreeSysroot     *sysroot,
...
+     g_str_has_suffix (booted_ref, "eos4"));
```
dbnicholson, line 525: Likewise, I think this should be `/eos4`. However, I wonder if we should add it at all. It would make the 2nd commit unneeded since the `eos4` branch wouldn't be considered for upgrade at all. On the other hand, if we later do add an `eos4` to `eos4a` upgrade, we would want the `eos4` to `latest1` block you've added here.

eos-updater/poll-common.c
```diff
@@ -580,6 +585,24 @@ should_follow_checkpoint (OstreeSysroot     *sysroot,
...
     }
 
+  /* https://phabricator.endlessm.com/T33311 */
+  if (is_conditional_upgrade_path &&
+      sys_vendor != NULL && product_name != NULL &&
+      booted_system_is_unsupported_by_eos4_kernel (sys_vendor, product_name))
+    {
+      *out_reason = g_strdup_printf (_("%s %s systems are not supported in EOS 5."), sys_vendor, product_name);
+      return FALSE;
+    }
```
dbnicholson, line 595: I think this is a no-op since it's exactly the same code as the T31772 block above. I think either you don't need this block or you should add a new `booted_system_is_unsupported_by_eos5_kernel` function and add that to the above conditional. Basically that block would become a "you can no longer upgrade to newer Endless releases for this hardware" conditional.

dbnicholson (changes requested):

eos-updater/poll-common.c
starnight: Good point!

eos-updater/poll-common.c
starnight: Oops!  Thanks for the catch up!

eos-updater/poll-common.c
starnight: If we are not going to have `eos4a` and `eos4` branch will not upgrade right now, then we can drop the 2nd commit.

tests/test-update-refspec-checkpoint.c
```diff
@@ -1491,6 +1491,8 @@ test_update_refspec_checkpoint_eos3a_eos4 (EosUpdaterFixture *fixture,
       { NULL, NULL, "ASUSTeK COMPUTER INC.", "Z550MA", FALSE, NULL, NULL, NULL, FALSE, FALSE },
       { NULL, NULL, "Endless", "ELT-JWM", FALSE, NULL, NULL, NULL, FALSE, FALSE },
       { NULL, NULL, "Endless", "ELT-JWM", FALSE, NULL, NULL, NULL, TRUE, TRUE },
+      { "latest1", NULL, "Endless", "EE-200", FALSE, NULL, NULL, NULL, FALSE, FALSE },
+      { "latest1", NULL, "Endless", "EE-200", FALSE, NULL, NULL, NULL, TRUE, TRUE },
```
dbnicholson, line 1495: Unfortunately, my request to make you change the suffix match to `/latest1` broke these tests. A ref starting with `/` is illegal, so I think you'll have to use something more specific like `os/eos/amd64/latest1`.

dbnicholson (changes requested):

tests/test-update-refspec-checkpoint.c
starnight: Oops!  My fault.

dbnicholson (approved):
Looks good!

PR #306, opened 2022-09-08
Revert "apply: Don't use staged deployments when /boot is automounted"

dbnicholson:
This reverts commit a19821a2b419850e188da3eee4fe3465528ca920. OSTree has been fixed to support this use case by keeping `/boot` open in the root namespace until the staged deployment completes finalization. See https://github.com/ostreedev/ostree/pull/2544 for details.

https://phabricator.endlessm.com/T33775

### DIFF
--- a/eos-updater/apply.c
+++ b/eos-updater/apply.c
@@ -270,13 +270,11 @@ apply_internal (ApplyData     *apply_data,
 
   origin = ostree_sysroot_origin_new_from_refspec (sysroot, update_refspec);
 
-  /* When booted into an OSTree system without an automount /boot, stage the
-   * deployment so that the /etc merge happens during shutdown. Otherwise
-   * (primarily sd-boot and the test suite), deploy the finalized tree
-   * immediately.
+  /* When booted into an OSTree system, stage the deployment so that the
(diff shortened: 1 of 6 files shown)


eos-updater/apply.c
```diff
@@ -270,13 +270,11 @@ apply_internal (ApplyData     *apply_data,
...
-  /* When booted into an OSTree system without an automount /boot, stage the
-   * deployment so that the /etc merge happens during shutdown. Otherwise
-   * (primarily sd-boot and the test suite), deploy the finalized tree
-   * immediately.
+  /* When booted into an OSTree system, stage the deployment so that the
+   * /etc merge happens during shutdown. Otherwise (primarily the test
+   * suite), deploy the finalized tree immediately.
    */
-  staged_deploy = ostree_sysroot_is_booted (sysroot) &&
-    !eos_updater_sysroot_boot_is_automount (sysroot, NULL);
```
wjt, line 279: Is the function we're no longer calling still needed elsewhere?

dbnicholson: No. Shall I revert it as well?

dbnicholson: Updated with the revert of the support code.

PR #309, opened 2022-11-17
Make licensing REUSE-compliant

pwithnall:
An interesting side diversion.

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -1,0 +1,38 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: eos-updater
+Upstream-Contact: Philip Withnall <withnall@endlessm.com>
+Source: github.com/endlessm/eos-updater
+
+# These test files are too simple to be copyrightable, but adding
+# copyright/licensing headers to them (or beside them in a .license file) may
+# affect the tests
+Files:
+ libeos-updater-util/tests/config-test-invalid.conf
+ libeos-updater-util/tests/config-test.conf
+ tests/gpghome/C1EB8F4E.asc
+ tests/gpghome/keyid
+ tests/gpghome/pubring.gpg
+ tests/gpghome/random_seed
+ tests/gpghome/secring.gpg
+Copyright: See individual files
+License: CC0-1.0
+
+# Debian packaging is under the same license as the package but the file formats
+# don’t easily allow copyright/license data to be added
+Files:
+ debian/**
+Copyright: See individual files
+License: LGPL-2.1-or-later
+
+# The JSON Schema format doesn’t easily allow copyright/license data to be added
+Files:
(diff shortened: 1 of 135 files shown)


.reuse/dep5
```diff
@@ -0,0 +1,38 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
```
dbnicholson, line 1: Does this essentially duplicate `debian/copyright`? Is there any way to use one as the other like a symlink? Not critical, but this seems like the type of information that would only get updated in one place or the other but not both.

dbnicholson (commented):
This looks great to me. Bravo for going through such a task. I don't think I can approve the license change on the man pages, but it seems totally reasonable to me.

eos-autoupdater/docs/eos-autoupdater.8
```diff
@@ -1,5 +1,6 @@
 .\" Manpage for eos-autoupdater.
-.\" Documentation is under the same licence as the eos-updater package.
+.\" SPDX-License-Identifier: CC-BY-SA-4.0
```
wjt, line 2:
```
git shortlog -se [all manpages touched by this patch]
```

reveals that all contributions have been made by Endless employees, except 2b732527fb316c3e9c386ccdad763421f6be2cae which was done under contract. So EOSF owns these files.

@ramcq Please approve this relicensing!

.reuse/dep5
pwithnall: tl;dr: Agreed, it’s awkward to have two sort-of-duplicate files, but I can’t think of a better approach. Suggestions welcome!

---

It is the same format as `debian/copyright`, but I’m not sure they can be merged as they have slightly conflicting purposes. `debian/copyright` aims to give the copyright/licensing information for every file in the repository, whereas `.reuse/dep5` is a way to give copyright/licensing information for files where it can’t be included inline. Its use is discouraged in favour of putting the licensing/copyright inside the files themselves.

If `.reuse/dep5` were to be merged/linked with `debian/copyright`, it would then contain licensing/copyright information for every file in the repository, which would go against the intent of using it only for difficult files.

One [alternative to using `.reuse/dep5`](https://reuse.software/spec/#comment-headers) is to instead add a `.license` file for each file listed in it (so `tests/gpghome/keyid.license` to accompany `tests/gpghome/keyid`, etc.). I shied away from that as it would pollute the repository a bit (and wildcards can’t be used, so we’d need O(N) files for `debian/**` rather than one entry in `.reuse/dep5`). I can give that a go if you think it’d be better.

eos-autoupdater/docs/eos-autoupdater.8
ramcq: Approved!

.reuse/dep5
wjt: I think this kind of duplication is better than littering the source tree with `.license` files.

PR #310, opened 2022-11-17
Polish flake8 setup

dbnicholson:
I went to run the test suite and my more recent version of `flake8` didn't like `eos-updater-prepare-volume`. Then I decided to get the whole project spruced up.


pwithnall (approved):
A++

PR #311, opened 2022-11-21
More eos4 → eos5 checkpoints

dbnicholson:
This adds checkpoints to:

* Confirm `/var/lib/flatpak/repo` has been split from `/ostree/repo`
* Confirm that the `sysroot.bootloader` option is set in the `/ostree/repo` config

before allowing upgrade from eos4 → eos5.

Both of the actual checks were pretty simple. I ended up spending much more time mucking with the test suite. Probably some of that can be split out to a separate PR if desired.

https://phabricator.endlessm.com/T34110

### DIFF
--- a/tests/test-update-refspec-checkpoint.c
+++ b/tests/test-update-refspec-checkpoint.c
@@ -1573,17 +1573,39 @@ test_update_refspec_checkpoint_eos3a_eos4 (EosUpdaterFixture *fixture,
   CheckpointTestData tests[] =
     {
       /* Normal system */
-      { NULL, NULL, NULL, NULL, FALSE, NULL, NULL, NULL, FALSE, TRUE },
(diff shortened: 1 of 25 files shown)


tests/test-update-refspec-checkpoint.c
```diff
@@ -1413,10 +1415,10 @@ do_update_refspec_checkpoint (EosUpdaterFixture  *fixture,
                               CheckpointTestData *test_machine,
                               gboolean            host_is_aarch64)
 {
-  const gchar *src_ref = test_machine->src_ref;
+  const gchar *src_ref = test_machine->src_ref ? test_machine->src_ref : default_src_ref;
```
starnight, line 1466: Good point!

pwithnall (approved):
Nice!

PR #312, opened 2022-11-21
eos-updater: Add a new UpdateIsUserVisible D-Bus property and implicitly set it if the major version in an update differs from what’s booted

pwithnall:
See the commit messages for explanations.

https://phabricator.endlessm.com/T34079

### DIFF
--- a/eos-updater/com.endlessm.Updater.xml
+++ b/eos-updater/com.endlessm.Updater.xml
@@ -186,6 +186,15 @@
     <property name="Version" type="s" access="read"/>
 
     <!--
+      UpdateIsUserVisible:
+
+      If the update contains significant user visible changes which should be
+      notified to the user in advance of the update being applied, this is
+      `True`. Otherwise, or if no update is available, this is `False`.
+    -->
+    <property name="UpdateIsUserVisible" type="b" access="read"/>
+
+    <!--
       DownloadSize:
 
       Size (in bytes) of the update when downloaded, or `-1` if an update is
--- a/eos-updater/poll-common.c
+++ b/eos-updater/poll-common.c
@@ -84,12 +84,33 @@ async_result_cb (GObject      *source_object,
   *result_out = g_object_ref (result);
 }
 
+/**
+ * is_checksum_an_update:
+ * @repo: the #OstreeRepo
+ * @update_checksum: checksum of the commit to potentially update to
+ * @booted_ref: ref which is currently booted
+ * @update_ref: ref of the branch to potentially update to
+ * @out_commit: (not optional) (nullable) (out) (transfer full): return location
+ *   for the #GVariant containing the commit identified by @update_checksum *if*
+ *   it is an update compared to @booted_ref; %NULL is returned otherwise
+ * @error: return location for a #GError, or %NULL
+ *
+ * Checks whether an update from @booted_ref to @update_ref would actually be
+ * an update, or would end up switching to an older release. See the block
+ * comment below for the details.
+ *
+ * The return value indicates whether there was an error in the check. It does
+ * not indicate whether the checksum is an update. If the checksum is an update,
+ * @out_commit returns a non-%NULL value.
+ *
+ * Returns: %TRUE if @out_commit is defined, %FALSE on error
+ */
 gboolean
 is_checksum_an_update (OstreeRepo *repo,
-                       const gchar *checksum,
+                       const gchar *update_checksum,
                        const gchar *booted_ref,
-                       const gchar *upgrade_ref,
-                       GVariant **commit,
+                       const gchar *update_ref,
+                       GVariant **out_commit,
                        GError **error)
 {
   g_autoptr(GVariant) current_commit = NULL;
@@ -100,8 +121,8 @@ is_checksum_an_update (OstreeRepo *repo,
   g_autoptr(GError) local_error = NULL;
 
   g_return_val_if_fail (OSTREE_IS_REPO (repo), FALSE);
-  g_return_val_if_fail (checksum != NULL, FALSE);
-  g_return_val_if_fail (commit != NULL, FALSE);
+  g_return_val_if_fail (update_checksum != NULL, FALSE);
+  g_return_val_if_fail (out_commit != NULL, FALSE);
   g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
 
   booted_checksum = eos_updater_get_booted_checksum (error);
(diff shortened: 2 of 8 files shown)


wjt:
Tests failing:

```
eos-updater-tests:ERROR:../tests/test-update-direct.c:780:test_update_sizes: assertion failed (eos_updater_get_download_size (updater) == expected_download): (11635 == 11855)
```

eos-updater/poll.c
```diff
@@ -494,6 +498,22 @@ check_for_update_following_checkpoint_commits (OstreeRepo     *repo,
                               error))
     return FALSE;
 
+  /* FIXME: Once we’ve worked out which commit to update to, we really should
+   * then iterate back through the commits on @new_refspec until we get to:
```
wjt, line 502: I'm a bit sceptical of the idea that determining whether updating from A to B should be treated as a major upgrade rather than a minor update should involve iterating over the complete commit history:

1. The server doesn't necessarily have the complete history for a ref. I checked on the server and the `os/eos/amd64/eos3a` only has versions `3.9.0` through `3.9.7`. I think this is an artefact of how our branching works, basically the `os/eos/amd64/eos3a` is an alias for the latest 3.x branch, `os/eos/amd64/eos3.9`. In the dev repo, `os/eos/amd64/master` only goes back 5 days. (For releases we do have the commits so we could in principle reconstruct this history. For nightlies we don't keep old commits.)
2. Each `.commit` object is like 4.3MB. If you have to fetch dozens of these it could end up comparable to the size of the update itself. And you either have to store them, or re-fetch them on each poll (in the case where the update is not fetched).
3. The client may not be able to fetch the history. Currently, if you have a random OS image it can be used as a USB updates source without further work. But our OS images don't include the complete ostree ref history for the commit within.

Here's another idea. Our version numbers take the form X.Y.Z, where X changes when we want it to; Y changes when we rebranch the OS (which we might do for reasons uninteresting to the typical user, such as branching 5.1 soon after 5.0 to introduce a newer kernel); and Z changes on point releases. Can we simply declare that we will give any new OS release that warrants a big banner a new X, as we did with EOS 4 (removed a bunch of features) and will do with EOS 5? Then we could determine whether updating from A to B is a major upgrade or minor update by comparing the version numbers, which is already a property of both commits. This would be consistent with how GNOME is versioned; and how (say) Fedora treats upgrades from a user's perspective, where 32 → 33 is noteworthy but miscellaneous package updates within a series is not.

pwithnall: Those are all good arguments against this design. I was shying away from comparing version numbers (or branch names) because of the fear of overloading their semantics, but I guess the version number semantics are an exact match for this, on EOS at least.

It won’t necessarily be such a great match for other distros who use/want to use eos-updater, but perhaps let’s cross that bridge when we come to it.

Comparing version numbers also isn’t a great match for the other potential uses I had in mind for `UpdateFlags` — indicating update urgency and severity. But since nobody’s asked for those features, they shouldn’t prejudice the implementation of this one. We can always implement them a different way in future, if anyone ever asks for them.

I’m looking at reworking this to compare version numbers. The `UpdateFlags` D-Bus property can stay (although perhaps I’ll change it to a stand-alone `UpdateIsUserVisible` property) but the `eos-updater.update-flags` commit metadata key will be removed. That means the corresponding [eos-build PR](https://github.com/endlessm/eos-build/pull/1398) will probably get closed unmerged.

wjt:
On Tue, 22 Nov 2022 at 16:57, Philip Withnall ***@***.***>
wrote:

> It won’t necessarily be such a great match for other distros who use/want
> to use eos-updater, but perhaps let’s cross that bridge when we come to it.
>

A fair point! But TTBOMK the one distro that does so is GNOME OS, which I
think would want the big banner for each 6-monthly major version bump.
(There was some past complaint that it is not only used for this.)

eos-updater/poll.c
pwithnall: Significantly reworked it to compare version numbers instead.

eos-updater/com.endlessm.Updater.xml
```diff
@@ -185,6 +185,15 @@
     -->
     <property name="Version" type="s" access="read"/>
 
+    <!--
+      UpdateIsUserVisible:
+
+      If the update contains significant user visible changes which should be
```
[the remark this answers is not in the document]
pwithnall: I can see where you’re coming from, but I disagree that it’s an issue. I’m thinking of the property name as a contraction of the (too long) `UpdateIsSignificantlyUserVisible`, and I think it’s OK for the docstring to clarify that.

I’m hesitant to use names like `MajorUpdate` as that encodes the semantics of “major version number change bi-implies significant user-visible changes” in API. The current naming means we can change the decision logic for what forms a significant user-visible update in future without changing API.

`SignificantChange` doesn’t quite encode what we’re looking for: for example, un-merging the flatpak and OS OSTree repos was a significant change, but it was almost entirely (hopefully) not-user-visible.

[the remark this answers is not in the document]
wjt: Alright! `UpdateIsNotableEnoughThatWeShouldTellTheUserAboutIt` is too long anyawy. :)

eos-updater/poll.c
dbnicholson: Definitely for EOS the release versions follow semantic versioning. I may be wrong, but I think I even added some code in the release tooling that ensures the version isn't a downgrade. So, I'd say for EOS comparing version numbers should work fine. Certainly there's code in eos-build/ostree that decides on major vs minor upgrade based entirely on version numbers.

(There's a bit of minutiae about how pre-releases are handled. The commit metadata version field uses `~` like in Debian version handling, but `~` isn't allowed in OSTree refs, so it's converted to `-` for the `release/*` refs.)

eos-updater/poll-common.c
```diff
@@ -84,6 +84,33 @@ async_result_cb (GObject      *source_object,
...
+
+  if (version_a == NULL || version_b == NULL)
+    return 0;
+
+  /* Take the first whole integer off each string, and assume it’s the major
+   * version number. This should work regardless of whether the strings are in
+   * `X.Y.Z` form or `X.Y` or `X`. Note that this parsing is locale
+   * independent. */
+  version_a_major = g_ascii_strtoull (version_a, NULL, 10);
+  version_b_major = g_ascii_strtoull (version_b, NULL, 10);
```
dbnicholson, line 104: Clever way to handle it! That seems pretty robust.

PR #317, opened 2022-11-29
build: Enable fatal warnings from g-ir-scanner

pwithnall:
We can do this unconditionally since the behaviour of `g-ir-scanner` is unlikely to change significantly, and warnings generally indicate binding problems which should probably count as a release blocker.

Currently, with `g-ir-scanner` 1.72.0, there are no warnings or errors.

https://phabricator.endlessm.com/T22636


pwithnall:
The use of `fatal_warnings` doesn’t need a Meson version bump because it’s [supported since 0.55](https://mesonbuild.com/Gnome-module.html#gnomegenerate_gir) and eos-updater depends on 0.58.

PR #319, opened 2022-12-09
eos-updater: Add a ReleaseNotesUri property

pwithnall:
This adds a `ReleaseNotesUri` property to the D-Bus interface. This is populated from an optional `eos-updater.release-notes-uri` metadata key in OSTree commits, which has a couple of optional string substitutions applied to it to allow a single URI template to be refined for multiple possible update paths.

This allows different release notes pages to be shown to users depending on which version they are updating from.

Includes a unit test.

https://phabricator.endlessm.com/T34082

### DIFF
--- a/eos-updater/docs/eos-updater.8
+++ b/eos-updater/docs/eos-updater.8
@@ -56,6 +56,43 @@ unit. See \fBsystemctl\fP(1).
 If the computer has been converted to not use OSTree, automatic updates are
 permanently disabled.
 .\"
+.SH COMMIT METADATA
+.IX Header "COMMIT METADATA"
+.\"
+\fBeos\-updater\fP understands various well\-known keys in the metadata of the
+OSTree commits it parses, in addition to the ones used by OSTree itself.
+.\"
+.IP "\fIeos.checkpoint\-target\fP (type \fBs\fP)" 4
+.IX Item "eos.checkpoint\-target"
+If this is specified, it contains the name of a new ref to upgrade the
+deployment to, but only if the booted OS is the commit containing the
+\fIeos.checkpoint\-target\fP key. Effectively, this creates a checkpoint commit
+which a computer must boot into before it can upgrade to any subsequent
+releases.
+.\"
+.IP "\fIeos\-updater.release\-notes\-uri\fP (type \fBs\fP)" 4
+.IX Item "eos\-updater.release\-notes\-uri"
+Optional URI pointing to release notes for the OS release contained in the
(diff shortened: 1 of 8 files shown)


eos-updater/poll-common.c
```diff
@@ -1190,6 +1217,13 @@ parse_latest_commit (OstreeRepo           *repo,
   else
     *out_version = g_variant_dup_string (version, NULL);
 
+  if (metadata != NULL)
+    release_notes_uri_template = g_variant_lookup_value (metadata, "eos-updater.release-notes-uri", G_VARIANT_TYPE_STRING);
+  if (release_notes_uri_template == NULL)
+    *out_release_notes_uri_template = NULL;
+  else
+    *out_release_notes_uri_template = g_variant_dup_string (release_notes_uri_template, NULL);
```
dbnicholson, line 1225: I was just writing some similar code. Doesn't this do the same thing without the intermediate variant?

```
if (metadata == NULL ||
    !g_variant_lookup (metadata, "eos-updater.release-notes-uri", "s", *out_release_notes_uri_template))
  *out_release_notes_uri_template = NULL;
```

I guess you're following the same pattern as above, though.

dbnicholson (changes requested):
Looks reasonable. It would be nice if there was some documentation of the `eos-updater.release-notes-uri` commit metadata value, particularly since it's actually a template. I'm not sure where that documentation would go, though. eos-updater(8)?

eos-updater/poll-common.c
pwithnall: You’re right. I’ve tweaked the commit and added a second commit which cleans up the code above.

eos-updater/docs/eos-updater.8
```diff
@@ -56,6 +56,43 @@ unit. See \fBsystemctl\fP(1).
...
+to updating from, as the current one has now reached end\-of\-life. The new ref
+must be on the same remote.
+.\"
+.IP "\fIversion\fP (type \fBs\fP)" 4
+.IX Item "version"
+Version number of the OS release contained in the commit. This is a string, but
+it is assumed to be in \fImajor.minor.micro\fP format. \fBeos\-updater\fP
+assumes that there are user-visible changes between different major versions of
+the OS, and notifies that in its D\-Bus interface.
+.\"
```
dbnicholson, line 95: :100: much appreciated. By now I have the metadata fields mostly internalized, but it still helps a lot to have a documented place to read about them.

dbnicholson (approved):

PR #320, opened 2022-12-15
tests: Reduce EOS_UPDATER_FORCE_FOLLOW_CHECKPOINT tests

dbnicholson:
We only need one of these tests with one or more known negative conditions to check that eos-updater honors the
`EOS_UPDATER_FORCE_FOLLOW_CHECKPOINT` environment variable. This is a slow test, so any redundant cases should be dropped.


pwithnall:
OK, although this will be reducing our branch coverage because only the first one of those negative paths is ever going to get taken by the test.

I suggest that we’re going to keep wanting to add more checkpoint tests, and if there’s a constant tension between speed and coverage, coverage might end up losing out. Perhaps we should be looking at speeding each test up instead. I realise that’s harder though, so let’s ship this for now. Thanks Dan :)

PR #321, opened 2022-12-19
Delay user visible autoupdate

dbnicholson:
The headline feature here is to delay applying user visible updates as controlled by the new `UserVisibleUpdateDelayDays` configuration option. To support this, eos-autoupdater writes the poll results to disk at `/var/lib/eos-updater/autoupdater-poll-results`. This allows detecting how long it's been since the update was seen.

Those are the last 2 commits in the series and both have tests. The rest of the series is tangential fixes to the test suite. Since there was previously no standalone autoupdater tests, the helper code was only allowing it to be run in forced mode. That's fine if you're just using the autoupdater to drive the eos-updater states, but it's not that helpful when trying to judge the autoupdater's behavior. I can split that part of the series out if it helps.

https://phabricator.endlessm.com/T34083


eos-autoupdater/main.c
```diff
@@ -193,6 +224,24 @@ info (const gchar *msgid,
   va_end (args);
 }
 
+static gint
+mkdir_print_errors (const gchar *path,
+                    gint         mode)
+{
+  gint ret = g_mkdir_with_parents (path, mode);
+
+  if (ret != 0) {
```
pwithnall, line 239: Nitpick: Coding style is to have braces 2-spaces indented on the following line. I think there are a few places where the autoupdater doesn’t comply with this style (which is confusing), but we might as well tidy them up when touching them.

Same in a few places in this commit.

eos-autoupdater/main.c
```diff
@@ -274,6 +316,153 @@ update_stamp_file (guint64 last_successful_update_secs,
...
+
+      return NULL;
+    }
+
+  results_variant = g_variant_new_from_bytes (G_VARIANT_TYPE ("a{sv}"),
+                                              results_contents,
+                                              FALSE);
+  g_variant_lookup (results_variant, "LastChangedUsecs", "t", &last_changed_usecs);
+  g_variant_lookup (results_variant, "UpdateRefspec", "&s", &update_refspec);
+  g_variant_lookup (results_variant, "UpdateID", "&s", &update_id);
```
pwithnall, line 366: Any/All of these lookups may fail (and return `FALSE`) if the file is corrupt. Is that OK, or do you want to return `NULL` from the function? In any case, perhaps a comment about how invalid files are handled would clarify things.

eos-autoupdater/main.c
```diff
@@ -274,6 +316,153 @@ update_stamp_file (guint64 last_successful_update_secs,
...
+  g_autoptr(GFile) results_file = NULL;
+  g_auto(GVariantBuilder) builder = G_VARIANT_BUILDER_INIT (G_VARIANT_TYPE ("a{sv}"));
+  g_autoptr(GVariant) results_variant = NULL;
+  g_autoptr(GBytes) results_contents = NULL;
+  g_autoptr(GError) error = NULL;
+
+  g_assert (results != NULL);
+
+  g_variant_builder_add (&builder, "{s@v}", "LastChangedUsecs",
+                         g_variant_new_variant (g_variant_new_uint64 (results->last_changed_usecs)));
```
pwithnall, line 397: Is the `g_variant_new_variant()` call needed here? I thought the `@v` format string takes a value (as a `GVariant*`) and wraps it in a variant wrapper.

i.e. I think (but I could easily be wrong, because this is confusing) that the current code will create the following dict (in text format): `{ 'LastChangedUsecs': <<@t 1234>>, 'UpdateRefspec': <<'some-refspec'>>, 'UpdateID': <<'some-id'>>}`

and what we want is:
`{ 'LastChangedUsecs': <@t 1234>, 'UpdateRefspec': <'some-refspec'>, 'UpdateID': <'some-id'>}`

eos-autoupdater/main.c
```diff
@@ -274,6 +316,153 @@ update_stamp_file (guint64 last_successful_update_secs,
...
+    g_variant_builder_add (&builder, "{s@v}", "UpdateID",
+                           g_variant_new_variant (g_variant_new_string (results->update_id)));
+
+  if (mkdir_print_errors (state_dir, 0755) != 0)
+    return;
+
+  results_variant = g_variant_builder_end (&builder);
+  results_contents = g_variant_get_data_as_bytes (results_variant);
+  results_path = g_build_filename (state_dir, POLL_RESULTS_NAME, NULL);
+  results_file = g_file_new_for_path (results_path);
```
pwithnall, line 411: Since we depend on GLib ≥ 2.56, I think this can be condensed to:
```suggestion
  results_file = g_file_new_build_filename (state_dir, POLL_RESULTS_NAME, NULL);
```

Same in a few other places too

tests/test-autoupdater.c
```diff
@@ -0,0 +1,242 @@
...
+
+  results_bytes = g_file_load_bytes (results_file, NULL, NULL, &error);
+  g_assert_no_error (error);
+  g_assert_nonnull (results_bytes);
+
+  results_variant = g_variant_new_from_bytes (G_VARIANT_TYPE ("a{sv}"),
+                                              results_bytes,
+                                              FALSE);
+  if (last_changed_usecs)
+    if (!g_variant_lookup (results_variant, "LastChangedUsecs", "t", last_changed_usecs))
```
pwithnall, line 105: Nitpick: these `if`s could be combined:
```suggestion
  if (last_changed_usecs &&
      !g_variant_lookup (results_variant, "LastChangedUsecs", "t", last_changed_usecs))
```

tests/test-autoupdater.c
```diff
@@ -237,6 +325,12 @@ main (int    argc,
   g_test_init (&argc, &argv, G_TEST_OPTION_ISOLATE_DIRS, NULL);
 
   eos_test_add ("/autoupdater/poll-results", NULL, test_poll_results);
+  UserVisibleTestData uvu_delay_data = { 1, FALSE, FALSE };
+  eos_test_add ("/autoupdater/user-visible-update/delay", &uvu_delay_data, test_user_visible_update_delay);
+  UserVisibleTestData uvu_nodelay_data = { 0, FALSE, TRUE };
+  eos_test_add ("/autoupdater/user-visible-update/nodelay", &uvu_nodelay_data, test_user_visible_update_delay);
+  UserVisibleTestData uvu_force_data = { 1, TRUE, TRUE };
+  eos_test_add ("/autoupdater/user-visible-update/force", &uvu_force_data, test_user_visible_update_delay);
```
pwithnall, line 333: Nitpick: It might be clearer to use compound literals, since we’re always targeting a compiler which can support them:
```suggestion
  eos_test_add ("/autoupdater/user-visible-update/delay", (&(UserVisibleTestData) {
                  .update_delay = 1,
                  .force_update = FALSE,
                  .expected_has_commit = FALSE,
                }), test_user_visible_update_delay);
  eos_test_add ("/autoupdater/user-visible-update/nodelay", (&(UserVisibleTestData) {
                  .update_delay = 0,
                  .force_update = FALSE,
                  .expected_has_commit = TRUE,
                }), test_user_visible_update_delay);
  eos_test_add ("/autoupdater/user-visible-update/force", (&(UserVisibleTestData) {
                  .update_delay = 1,
                  .force_update = TRUE,
                  .expected_has_commit = TRUE,
                }), test_user_visible_update_delay);
```

eos-autoupdater/docs/eos-autoupdater.conf.5
```diff
@@ -63,6 +63,14 @@ random interval, uniformly distributed between 0 and the number of days set in
 this option. It must be \fI0\fP or greater. An interval of \fI0\fP days means
 that the wait is always exactly \fIIntervalDays=\fP long.
 .\"
+.IP "\fIUserVisibleUpdateDelayDays=\fP"
+.IX Item "UserVisibleUpdateDelayDays="
+Number of days to wait until applying a user visible update. If
```
pwithnall, line 68:
```suggestion
Number of days to wait until applying a ‘user visible’ update. If
```

Just to emphasise the ‘user visible’ bit of it a bit

eos-autoupdater/docs/eos-autoupdater.conf.5
```diff
@@ -63,6 +63,14 @@ random interval, uniformly distributed between 0 and the number of days set in
...
+\fBeos\-autoupdater\fP(8) is triggered earlier than this number of days
+since a user visible update was seen, it will exit. It must be \fI0\fP
+or greater. An interval of \fI0\fP days means that user visible updates
+will be applied immediately.
```
pwithnall, line 72: Might also be worthwhile mentioning that `--force-update` will cause this setting to be ignored

pwithnall (approved):
Looks generally great! I only have a few little nitpicks/suggestions. Looks good to merge once those are addressed

eos-autoupdater/main.c
```diff
@@ -343,7 +343,9 @@ do_update_step (UpdateStep step, EosUpdater *proxy)
     case UPDATE_STEP_FETCH:
       {
         g_auto(GVariantDict) options_dict = G_VARIANT_DICT_INIT (NULL);
-        g_variant_dict_insert (&options_dict, "force", "b", force_update);
+        const gchar *force_fetch_env = g_getenv ("EOS_UPDATER_TEST_AUTOUPDATER_FORCE_FETCH");
```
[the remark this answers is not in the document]
dbnicholson: Fair. I'll add a CLI option.

eos-autoupdater/main.c
dbnicholson: Good catch. I agree with making the style consistent when in the area.

eos-autoupdater/main.c
dbnicholson: Yes, a further comment would help. Originally I had all of these catching and handling the `NULL` results, but then I added the `poll_results_new` helper so they could all be handled in one place.

I do believe this should try to handle a corrupted data file. Even though it might hide problems in the code, I think it would be better than having the autoupdater get wedged. I'll add some warnings here so maybe someone will notice them in the logs if this code has persistent issues.

tests/test-autoupdater.c
dbnicholson: :+1:

tests/test-autoupdater.c
dbnicholson: Wow, I was not aware you could do such a thing in C.

eos-autoupdater/docs/eos-autoupdater.conf.5
dbnicholson: Good call.

eos-autoupdater/main.c
dbnicholson: Sorry, now I see what you're saying. The helper is already returning `NULL` if the file doesn't exist, so should it do the same if there are fields missing? I'm worried about the autoupdater never applying a user visible update if there are issues in serializing or deserializing the state. However, I don't know if it will do the right thing with partial results, so I think maybe it is best to return `NULL` with a warning when that happens.

eos-autoupdater/main.c
dbnicholson: After reading the documentation again and writing a little test program, using `@v` results in the variant being used directly instead of having `g_variant_new` create a new variant automatically. This is how it's used in other parts of eos-updater. I don't know why you wouldn't let the variant get created automatically, though. It's also what the [dictionaries format string](https://docs.gtk.org/glib/gvariant-format-strings.html#dictionaries) documentation tells you to do.

Here's my test program and output:
```
#include <glib.h>

int main(int argc, char *argv[])
{
  GVariantBuilder *a, *b;
  GVariant *a_dict, *b_dict;
  gchar *a_str, *b_str;

  a = g_variant_builder_new (G_VARIANT_TYPE ("a{sv}"));
  g_variant_builder_add (a, "{s@v}", "key",
                         g_variant_new_variant (g_variant_new_string ("value")));
  a_dict = g_variant_builder_end (a);
  a_str = g_variant_print (a_dict, FALSE);

  b = g_variant_builder_new (G_VARIANT_TYPE ("a{sv}"));
  g_variant_builder_add (b, "{sv}", "key",
                         g_variant_new_string ("value"));
  b_dict = g_variant_builder_end (b);
  b_str = g_variant_print (b_dict, FALSE);

  g_print ("a: %s\nb: %s\n", a_str, b_str);

  return 0;
}
$ ./vardict 
a: {'key': <'value'>}
b: {'key': <'value'>}
```

eos-autoupdater/main.c
dbnicholson: I ended up wanting to print the path on errors, so I decided to keep what I had rather than having to free the results of `g_file_get_path` in the error cases.

tests/test-autoupdater.c
dbnicholson: Unfortunately, this gets miscompiled. I believe because `g_test_add` is a macro whose body is a block, the lifetime of the compound literal ends after the call to `eos_test_add`. By the time `g_test_run` iterates the table, the addresses are no longer valid. What's happening for me is that the last compound literal gets used for all the cases since the compiler is reusing the same address.

I'll play around with it a bit more, but I think I'll likely have to keep what I have.

dbnicholson: If I change `eos_test_add` to be a function instead of macro, it works.

dbnicholson: Oh, this has revealed a bug in https://github.com/endlessm/eos-updater/commit/11a34ee1625129b02909069db2740189a2bd0145 since the first 2 cases weren't actually being run. I'll split this part out.

dbnicholson: #322

dbnicholson:
I've addressed all the comments. I already squashed all the changes into the original commits, so [this diff](https://github.com/endlessm/eos-updater/compare/fddc15ae0fb9e7901e7bec84521f065061dc895d..00d5db010e933f18545e74654d3a54ab6c6ff5c9) will have to suffice if you want to see what changed.

Since you're out a couple more days, I'm going to merge this now.

eos-autoupdater/main.c
pwithnall: You can use `g_file_peek_path()` instead of `g_file_get_path()` to avoid freeing the results. For local files it returns the internally-used path. For remote files, it builds the path (as with `g_file_get_path()`) but caches it internally.

But since you’ve already reworked this once, don’t feel obliged to rework it again! Just mentioning `g_file_peek_path()` for future reference really.

eos-autoupdater/main.c
pwithnall: My comment was (not worded very well and) about whether it’s OK for the function to ignore partial errors (some but not all expected fields missing/incorrectly typed), or whether the function should return `NULL` on any error and only return success if every field is found and is valid.

The question of whether the update should be applied if this function returns an error is a separate one. I agree that we should avoid a corrupt file here blocking an update, so printing a warning and carrying on anyway seems appropriate.

pwithnall:
Looks good, thanks for the detailed replies and for going ahead and merging :+1:

PR #322, opened 2023-01-03
Fix tests using compound literals

dbnicholson:
Passing compound literals to `eos_test_add` was subtly broken due the implementation of the `g_test_add` macro. Workaround it by converting `eos_test_add` to a function and fixing a test that was silently being skipped.


dbnicholson:
I just remembered you're out for a couple more days, so I'm going to merge this now. Feel free to add any comments if you think my analysis is wrong or there should be another approach.

pwithnall (commented):
LGTM, retrospectively :)

PR #324, opened 2023-01-05
eos-updater: Only allow latest2 checkpoint from latest1

dbnicholson:
Only EOS 4 systems on the latest ref (latest1) should upgrade to EOS 5 (latest2). Systems on the LTS ref (eos4) or any other ref should not. In order to ensure that the validation applies to any ref trying to follow the latest2 checkpoint, switch the initial check so that it applies to the target ref instead of the booted ref.

https://phabricator.endlessm.com/T34298


starnight (approved):
Thanks for the catch up!

PR #328, opened 2023-01-16
eos-updater: Install python-dbusmock template

pwithnall:
Since https://github.com/martinpitt/python-dbusmock/pull/169 (not yet released!), python-dbusmock supports loading them from `$XDG_DATA_DIRS`.

If we install the template, other projects (like gnome-software) can eventually use the installed version rather than shipping their own copy.

### DIFF
--- a/debian/eos-updater-dev.install
+++ b/debian/eos-updater-dev.install
@@ -2,3 +2,4 @@ usr/lib/*/libeos-updater-util-0.so
 usr/lib/*/libeos-updater-flatpak-installer-0.so
 usr/share/gir-1.0/EosUpdaterUtil-0.gir
 usr/share/gir-1.0/EosUpdaterFlatpakInstaller-0.gir
+usr/share/python-dbusmock-templates/eos_updater.py
--- a/eos-updater/tests/meson.build
+++ b/eos-updater/tests/meson.build
@@ -1,7 +1,10 @@
 # Copyright 2022 Endless OS Foundation, LLC
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
-# python-dbusmock doesn’t currently provide a way to install out-of-tree templates
+install_data('eos_updater.py',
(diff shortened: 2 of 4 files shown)


pwithnall:
As I though, it’s giving a meson warning about using `install_tag` without requiring meson 0.60. I’ll fix that up.

pwithnall:
Pushed an additional commit to bump the meson dependency to 0.60. That should be fine because we ship 0.61 in EOS 5.0, and this PR isn’t going to get backported.

debian/eos-updater-dev.install
```diff
@@ -2,3 +2,4 @@ usr/lib/*/libeos-updater-util-0.so
 usr/lib/*/libeos-updater-flatpak-installer-0.so
 usr/share/gir-1.0/EosUpdaterUtil-0.gir
 usr/share/gir-1.0/EosUpdaterFlatpakInstaller-0.gir
+usr/share/python-dbusmock-templates/eos_updater.py
```
dbnicholson, line 5: No newline at the end of the file.

eos-updater/tests/meson.build
```diff
@@ -2,6 +2,10 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
 # python-dbusmock doesn’t currently provide a way to install out-of-tree templates
```
dbnicholson, line 4: This comment is outdated now.

dbnicholson (approved):
Looks good modulo a couple nitpicks.

debian/eos-updater-dev.install
pwithnall: Fixed

eos-updater/tests/meson.build
pwithnall: Fixed, not sure how I missed that :facepalm:

pwithnall:
Merging after fixing the nitpicks, thanks for the review!

PR #331, opened 2023-08-22
Support building with no tests

nanonyme:
With freedesktop-sdk we build eos-updater but without tests. As part of https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1564 this submission is to allow said build mode without patches

### DIFF
--- a/libeos-updater-flatpak-installer/meson.build
+++ b/libeos-updater-flatpak-installer/meson.build
@@ -61,4 +61,6 @@ libeos_updater_flatpak_installer_gir = gnome.generate_gir(libeos_updater_flatpak
   fatal_warnings: true,
 )
 
-subdir('tests')
+if get_option('tests')
+  subdir('tests')
+endif
--- a/meson.build
+++ b/meson.build
@@ -149,4 +149,6 @@ subdir('eos-updater-avahi')
 subdir('eos-updater-ctl')
 subdir('eos-updater-flatpak-installer')
 subdir('eos-updater-prepare-volume')
-subdir('tests')
+if get_option('tests')
+  subdir('tests')
+endif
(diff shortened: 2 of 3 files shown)


nanonyme:
Ping @ramcq

pwithnall:
Why would you want to support building with no tests?

nanonyme:
Changelog says they were patched out because they required building ostree against libsoup2

nanonyme:
We build ostree explicitly with --without-soup. We use curl instead.

nanonyme:
I think it's indeed better to get rid of soup dependency for tests.

dsd:
Thanks @pwithnall . I had assumed this was rather innocent, but you have good points, so I have reverted the commit.
I tried to look for any further info about the actual failure (test failure with ostree-with-curl?) but didn't see anything. In fact the [merge req mentioned above](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/12157) only seems to [disable tests because of flake8 being a dependency](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/12157/diffs?commit_id=dc0aa4f5dcd6eb48154ff3b52b1e1271f8486bc2)?

Happy to reconsider approaches for this once we have a better understanding of the problem at hand.

nanonyme:
@dsd building the tests when eos-updater is built against ostree built against curl used to fail but this seems not to be the case anymore. IMHO it's perfectly fine for tests to require flake8 and this simply needs to be fixed downstream if there's a problem with that.

nanonyme:
@pwithnall so the correct fix for eos-updater would be "drop dependency on ostree-trivial-httpd". I would file it as an issue but apparently this repo has issue tracker disabled.

nanonyme:
This was already dropped as part of https://github.com/endlessm/eos-updater/pull/331

dsd:
Great, sounds like we are done here then?

PR #332, opened 2023-12-08
Port to Soup 3

dbnicholson:
I think this is not urgent since there's no longer a potential conflict with ostree, but I went ahead and did it. The soup usage is actually pretty minimal:

1. `SoupServer` in `eos-update-server` and `libeos-update-server`
2. `SoupServer` used for tests
3. A stray `SoupURI` parsing that can just be converted to use `GUri`

The `SoupServer` API is mostly the same with some shuffled symbol names and requiring use of accessors for objects. While here I did some dependency cleanup. The actual Soup 3 porting is in one big commit at the end.

https://phabricator.endlessm.com/T33677


starnight:
I don't know why I cannot click approve in the review ...  GitHub is weird today.

But, thanks!

PR #333, opened 2023-12-11
Use g_autoslist to clean up GSLists

wjt:
Reasonable people may disagree about whether this is cleaner or not but I saw https://github.com/endlessm/eos-updater/commit/810a04b0fb126dc82c30a4e31c5efe33d294dd35 on #332  and thought “Hey, I think there's a way to do that”.

https://phabricator.endlessm.com/T33677


dbnicholson (approved):
Wow, I never noticed that before. It's not shown anywhere in the gi-docgen docs, but even in the [old documentation](https://developer-old.gnome.org/glib/stable/glib-Miscellaneous-Macros.html#g-autoslist) it's not in a place I would have looked.

PR #334, opened 2023-12-13
Bump version to 1.4.0

wjt:
Our package has been building as 1.3.0~rc3 for, as best I can tell, 10 years.

Update the version number in meson.build. I also pushed a `Version_1.4.0` tag pointing at this commit.

This is for the benefit of GNOME OS, who would like to remove the soup-2.4 dependency, ideally sooner than the release of Endless OS 5.2 at some point in the future, but do not want to build from a random Git commit.

https://gitlab.gnome.org/GNOME/gnome-build-meta/-/merge_requests/2548


wjt:
Added a patch to make the flake8 dependency optional, which is helpful in gnome-build-meta because there is no flake8 readily available, and the build process doesn't run tests anyway.

wjt:
retest this please

wjt:
I guess the flake8 thing was discussed here:

https://github.com/endlessm/eos-updater/pull/331#issuecomment-1837435433

When I found myself adding flake8 and all its build dependencies to gnome-build-meta, I have to say I felt differently about it. :)

PR #335, opened 2024-03-08
Miscellaneous cleanup

wjt:
Prep work for https://phabricator.endlessm.com/T35247.

PR #338, opened 2024-03-15
Prevent systems using nvme-remap from crossing checkpoints

wjt:
- Remove all previous checkpoint checks
- Factor out the should_follow_checkpoint() function so it can be tested in isolation
- Add a new check that prevents upgrades if nvme-remap driver is in use

~~I don't actually have this hardware so this is only tested by the test case; and it's a Friday evening; hence draft.~~

I targeted `eos5.1` with this PR because that's where the code needs to land to block upgrading to `eos6.0`; we'd merge to master after landing on `eos5.1`, so the same commits (not cherry-picks) will be on both.

https://phabricator.endlessm.com/T35247
https://phabricator.endlessm.com/T35250


libeos-updater-util/checkpoint.c
```diff
@@ -0,0 +1,145 @@
...
+{
+  g_autoptr(GEnumClass) enum_class = G_ENUM_CLASS (g_type_class_ref (EUU_TYPE_CHECKPOINT_BLOCK));
+  GEnumValue *value = g_enum_get_value (enum_class, (gint) reason);
+
+  g_return_val_if_fail (value != NULL, NULL);
+  return value->value_nick;
+}
+
+static gboolean
+is_nvme_remap_in_use (OstreeSysroot *sysroot)
```
wjt, line 48: I don't actually have the relevant hardware. I will build some packages and ask those in Taipei to check them, if this doesn't land before that.

libeos-updater-util/tests/checkpoint.c
```diff
@@ -60,6 +59,7 @@ teardown (Fixture       *fixture,
 {
   g_clear_object (&fixture->sysroot);
   g_clear_object (&fixture->repo);
+  g_clear_object (&fixture->root_dir);
```
dbnicholson, line 62: I think this and the `g_unsetenv` below should be rolled into the previous commit.

libeos-updater-util/tests/checkpoint.c
```diff
@@ -89,6 +89,69 @@ test_force_no_follow (Fixture                  *fixture,
 
   g_assert_false (euu_should_follow_checkpoint (fixture->sysroot, "os/eos/amd64/latest2", "os/eos/amd64/latest3", &reason));
   g_assert_cmpstr (reason, !=, NULL);
+
+  g_unsetenv ("EOS_UPDATER_FORCE_FOLLOW_CHECKPOINT");
```
dbnicholson, line 99: Probably wouldn't be a bad idea to run this also in `setup()` to ensure it's not inherited from the environment for tests that don't set it.

eos-updater/poll-common.c
```diff
@@ -407,20 +380,77 @@ cleanstr (gchar *s)
...
+                        product,
+                        booted_ref,
+                        target_ref,
+                        reason_nick ?: "");
+}
+
+static gboolean
+stop_aggregate_timer (gpointer data)
+{
+  g_return_val_if_fail (EMTR_IS_AGGREGATE_TIMER (data), G_SOURCE_REMOVE);
```
dbnicholson, line 410: Maybe the compiler will elide this function, but I'd assume this function needs to be under `
#ifdef HAS_EOSMETRICS_0` for the `emtr` symbols and macros.

eos-updater/poll-common.c
```diff
@@ -407,20 +380,77 @@ cleanstr (gchar *s)
...
+                                  1,
+                                  stop_aggregate_timer,
+                                  timer,
+                                  g_object_unref);
+    }
+  else
+    {
+      g_message ("Skipping metric event %s: %s (metrics disabled)",
+               EOS_UPDATER_METRIC_CHECKPOINT_BLOCKED,
+               payload_str);
```
dbnicholson, line 451: I presume this was `g_debug` before and now the alignment is off.

dbnicholson (changes requested):
Looks good. I left some minor issues, but overall I like it. I think the pseudo test in the last commit leaves a bit to be desired:

* We don't actually install the `libeos-updater-util-0-tests` package. It would be nice to be able to do the check without having to alter the system.
* The refs are hardcoded, so it would make it hard to test if we had checkpoints that only needed to be crossed on certain upgrade paths.

That said, it's good enough.

libeos-updater-util/checkpoint.c
```diff
@@ -0,0 +1,70 @@
...
+euu_should_follow_checkpoint (OstreeSysroot     *sysroot,
+                              const gchar       *booted_ref,
+                              const gchar       *target_ref,
+                              gchar            **out_reason)
+{
+  g_return_val_if_fail (OSTREE_IS_SYSROOT (sysroot), FALSE);
+  g_return_val_if_fail (booted_ref != NULL, FALSE);
+  g_return_val_if_fail (target_ref != NULL, FALSE);
+  g_return_val_if_fail (out_reason != NULL, FALSE);
+  g_return_val_if_fail (*out_reason == NULL, FALSE);
```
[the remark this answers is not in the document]
wjt: Good point. This does get deleted in a subsequent patch but I will amend it.

eos-updater/poll-common.c
wjt: Good point!

The other option is to not wait 1 second and just stop it immediately after it starts. The aggregate will still get reported even if its count is 0…

eos-updater/poll-common.c
wjt: Yes, I changed this for debugging purposes and didn't mean to commit it.

wjt:
The pseudo-test was done like that for convenience. I guess you're right and I should do the small amount of buildsystem work to install `/usr/libexec/eos-updater-check-checkpoint-checks` or similar. Honestly one reason I didn't is that I can't decide what to call it.

libeos-updater-util/checkpoint.c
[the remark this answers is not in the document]
wjt: Actually my original code is correct. `out_reason` has type `char **`. The first assertion tests that the caller has supplied a non-NULL pointer as the location to store a pointer to the reason. The second assertion is that that location currently holds a NULL pointer, so that assigning to `*out_reason` will not leak whatever string was previously there.

eos-updater/poll-common.c
wjt: I moved the `#ifdef` out to encompass both functions, and added an `#else` branch that defines a stub `maybe_send_checkpoint_blocked_event()`.

libeos-updater-util/checkpoint.c
wjt: @starnight confirmed that this works correctly on the device he has that uses `nvme-remap`. Thanks!

wjt:
I moved the dodgy fake test to a standalone `eos-updater-test-checkpoint` executable, installed as part of the `eos-updater` package. I am not that keen on the name – it doesn't test the checkpoint, it tests the checks that are applied at the checkpoint – but, good enough? For good measure I also added a manpage, and hope to never write groff source again in my lifetime.

libeos-updater-util/checkpoint.c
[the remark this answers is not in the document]
dbnicholson: Ah, correct on both accounts.

eos-updater-test-checkpoint/meson.build
```diff
@@ -0,0 +1,24 @@
...
+  c_args: [
+    '-DG_LOG_DOMAIN="eos-updater-test-checkpoint"'
+  ],
+  dependencies: eos_updater_test_checkpoint_deps,
+  include_directories: root_inc,
+  install: true,
+  install_dir: get_option('libexecdir'),
+)
+
+install_man('docs/eos-updater-test-checkpoint.8')
```
dbnicholson, line 24: Missing newline here.

eos-updater-test-checkpoint/meson.build
```diff
@@ -0,0 +1,24 @@
+# Copyright 2024 Endless OS Foundation, LLC
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+eos_updater_test_checkpoint_sources = [
+  'main.c'
+]
+
+eos_updater_test_checkpoint_deps = [
+  dependency('ostree-1', version: '>= 2019.2'),
```
dbnicholson, line 9: Can you change this to `ostree_dep` per 0104703e68bcb2442b386aa43beb27b0ec1ee6c2?

dbnicholson (approved):
Looks great. Way to roff, by the way. I can't think of a better name for the program, so this all looks fine to me modulo the last couple nitpicks.

eos-updater-test-checkpoint/meson.build
wjt: This bad default is one of the reasons I added an `.editorconfig` in #335 :)

eos-updater-test-checkpoint/meson.build
wjt: I can't, because that patch isn't on `eos5.1`. Oh, but it's the very next commit after the current `eos5.1` so I'll rebase this branch onto it.

dbnicholson: Ah ha. I forgot about the eos5.1 base.

eos-updater/poll-common.c
```diff
@@ -406,21 +380,83 @@ cleanstr (gchar *s)
...
-EosMetricsInfo *
-eos_metrics_info_new (const gchar *booted_ref)
+static GVariant *
+checkpoint_blocked_event_payload_new (const char         *booted_ref,
+                                      const char         *target_ref,
+                                      EuuCheckpointBlock  reason)
 {
   g_autoptr(GHashTable) hw_descriptors = NULL;
-  g_autoptr(EosMetricsInfo) info = NULL;
+  g_autofree gchar *vendor, *product;
```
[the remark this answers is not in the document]
wjt: I think they're assigned to on all paths so it's safe not to, but you're right, it would be clearer to do thay

PR #340, opened 2024-03-21
Initialise g_auto* variables

wjt:
The `vendor` & `product` case was noticed by @dbnicholson on another PR. As described in the commit message and in that comment that I have now lost, I don't think this is a correctness issue, but it is a good-practice and readability issue.

I then searched the source tree for other lines containing the string `g_auto` and not the string `=`, and found a couple more cases. One of them is actually wrong, but only in an error path that will never be reached in practice.

https://phabricator.endlessm.com/T35247


dbnicholson (approved):
I agree it's probably nothing. I'll merge when the CI passes.

PR #341, opened 2024-04-18
Update README with testing information

manuq:


README.md
```diff
@@ -44,6 +44,15 @@ using `meson test`; and some installed-only tests which can be run using
 `sudo gnome-desktop-testing-runner eos-updater-0` in your installation prefix
 _after_ installing eos-updater.
 
+For the post-install tests you need `gnome-desktop-testing-runner`
+installed (eg. `sudo apt-get install gnome-desktop-testing`) and
+to configure the build like this:
+
+```
+meson setup builddir
+meson configure -Dinstalled_tests=true builddir/
```
dbnicholson, line 53: You can actually pass the `-D` options directly to `setup`. The only reason to use `configure` is to change the configured options.

dbnicholson (changes requested):
That's helpful, just one minor nitpick.

README.md
manuq: Ah, that's much better, thanks!

PR #342, opened 2024-11-15
ctl: Don't show Python backtrace if no update is available

wjt:
https://phabricator.endlessm.com/T26906


dbnicholson (approved):
Nice work! This is indeed a long embedded thorn.

PR #343, opened 2026-03-31
apply: Check the partition's free size of /boot before apply

starnight:
The systems which start before EOS 5.0.0 have only 250 MB ESP partition. However, the kernel & initramfs composed EFI image has grown to almost 100 MB. The PAYG system mounts ESP partition to /boot. The space is not enough to hold the 3rd EFI image for PAYG provisioned system in the ESP during OSTree deploying the new updated OSTree. So, here introduces `check_boot_free_space()` function to check the partition's free size of /boot in the early apply.

Delete one unused ostree deployment if the the partition's free size of /boot is not enough. The used ostree deployment could be **pending**, or **rollback** ostree deployment. But, prefer removing the **rollback** (older) over the **pending** deployment. Then, there should be enough free space for the new updated OSTree deployment.

https://phabricator.endlessm.com/T33799

### DIFF
--- a/eos-updater/apply.c
+++ b/eos-updater/apply.c
@@ -22,6 +22,7 @@
  *  - Vivek Dasmohapatra <vivek@etla.org>
  */
 
+#include <sys/statvfs.h>
 #include <eos-updater/apply.h>
 #include <eos-updater/data.h>
 #include <eos-updater/object.h>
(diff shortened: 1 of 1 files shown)


starnight:
Should prepare this to eos6.0.  Then, cherry pick to master.

PR #344, opened 2026-03-31
apply: Check the partition's free size of /boot before apply

starnight:
The systems which start before EOS 5.0.0 have only 250 MB ESP partition. However, the kernel & initramfs composed EFI image has grown to almost 100 MB. The PAYG system mounts ESP partition to /boot. The space is not enough to hold the 3rd EFI image for PAYG provisioned system in the ESP during OSTree deploying the new updated OSTree. So, here introduces `check_boot_free_space()` function to check the partition's free size of /boot in the early apply.

Delete one unused ostree deployment if the the partition's free size of /boot is not enough. The used ostree deployment could be **pending**, or **rollback** ostree deployment. But, prefer removing the **rollback** (older) over the **pending** deployment. Then, there should be enough free space for the new updated OSTree deployment.

https://phabricator.endlessm.com/T33799

### DIFF
--- a/eos-updater/apply.c
+++ b/eos-updater/apply.c
@@ -22,6 +22,9 @@
  *  - Vivek Dasmohapatra <vivek@etla.org>
  */
 
+#include <sys/stat.h>
+#include <sys/statvfs.h>
+#include <sys/utsname.h>
 #include <eos-updater/apply.h>
 #include <eos-updater/data.h>
 #include <eos-updater/object.h>
@@ -230,6 +233,69 @@ update_remote_branches (OstreeRepo   *repo,
 }
 
 static gboolean
+check_boot_free_space (gboolean *enough_space, GError **error)
+{
+  g_autofree gchar *cmdline = NULL;
+  struct statvfs boot_stat;
+  guint64 boot_free_bytes;
+  /* Default requirement: at least 100 MiB free on /boot */
+  guint64 boot_min_free_bytes = 100 * 1024 * 1024;
+  struct utsname uts;
+  struct stat image_stat;
+
+  *enough_space = TRUE;
+
+  /* Only check free space on provisioned PAYG systems, identified by the
+   * presence of "eospayg" in the kernel command line.
+   */
+  if (!g_file_get_contents ("/proc/cmdline", &cmdline, NULL, error))
+    return FALSE;
+
+  if (strstr (cmdline, "eospayg") == NULL)
+    return TRUE;
+
+  if (uname (&uts) == 0)
+    {
+      g_autofree gchar *image_path = g_strdup_printf ("/lib/modules/%s/payg-image.efi",
+                                                      uts.release);
+      if (stat (image_path, &image_stat) == 0)
+        {
+          /* Requirement is 1.1 times the current UKI image size */
+          boot_min_free_bytes = (guint64) (image_stat.st_size * 1.1);
+          g_debug ("PAYG image %s size is %" G_GUINT64_FORMAT " bytes, "
+                   "setting requirement to %" G_GUINT64_FORMAT " bytes",
+                   image_path, (guint64) image_stat.st_size, boot_min_free_bytes);
+        }
+      else
+        {
+          g_warning ("Could not stat %s: %s. Using default 100 MiB.",
+                     image_path, g_strerror (errno));
+        }
+    }
+
+  if (statvfs ("/boot", &boot_stat) != 0)
+    {
+      g_set_error (error, G_IO_ERROR, g_io_error_from_errno (errno),
+                   "Failed to check free space on /boot: %s",
+                   g_strerror (errno));
+      return FALSE;
+    }
+
+  boot_free_bytes = (guint64) boot_stat.f_bavail * (guint64) boot_stat.f_frsize;
+  g_message ("Free space on /boot: %" G_GUINT64_FORMAT " bytes", boot_free_bytes);
+
+  if (boot_free_bytes < boot_min_free_bytes)
+    {
+      *enough_space = FALSE;
+      g_warning ("Not enough free space on /boot: have %" G_GUINT64_FORMAT " bytes, "
+                 "need at least %" G_GUINT64_FORMAT " bytes",
+                 boot_free_bytes, boot_min_free_bytes);
+    }
+
+  return TRUE;
+}
+
+static gboolean
 apply_internal (ApplyData     *apply_data,
                 GCancellable  *cancellable,
                 GError       **error)
@@ -243,6 +309,7 @@ apply_internal (ApplyData     *apply_data,
   g_autoptr(GKeyFile) origin = NULL;
   g_autoptr(OstreeSysroot) sysroot = NULL;
   const gchar *osname = get_test_osname ();
+  gboolean enough_boot_space = TRUE;
   gboolean staged_deploy;
   g_autoptr(GError) local_error = NULL;
 
@@ -255,6 +322,13 @@ apply_internal (ApplyData     *apply_data,
     return FALSE;
   if (!ostree_sysroot_load (sysroot, cancellable, error))
     return FALSE;
+
+  /* Check that /boot has enough free space before deploying. */
+  if (!check_boot_free_space (&enough_boot_space, &local_error))
+    {
+      g_warning ("Failed to check free space on /boot: %s", local_error->message);
+      g_clear_error (&local_error);
+    }
 
   booted_deployment = eos_updater_get_booted_deployment_from_loaded_sysroot (sysroot,
                                                                              error);
--- a/eos-updater/apply.c
+++ b/eos-updater/apply.c
@@ -296,6 +296,48 @@ check_boot_free_space (gboolean *enough_space, GError **error)
 }
 
 static gboolean
+remove_one_unused_deployment (OstreeSysroot  *sysroot,
+                              GCancellable   *cancellable,
+                              GError        **error)
+{
+  g_autoptr(OstreeDeployment) pending = NULL;
+  g_autoptr(OstreeDeployment) rollback = NULL;
+  g_autoptr(GPtrArray) deployments = NULL;
+  OstreeDeployment *to_remove = NULL;
+
+  ostree_sysroot_query_deployments_for (sysroot, NULL, &pending, &rollback);
+
+  /* Prefer removing the rollback (older) over the pending deployment */
+  if (rollback != NULL)
+    to_remove = rollback;
+  else if (pending != NULL)
+    to_remove = pending;
+  else
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_NO_SPACE,
+                   "Not enough free space on /boot and no unused deployment to remove");
+      return FALSE;
+    }
+
+  g_message ("Removing %s deployment to free space on /boot: "
+             "OS name: %s, checksum: %s",
+             to_remove == rollback ? "rollback" : "pending",
+             ostree_deployment_get_osname (to_remove),
+             ostree_deployment_get_csum (to_remove));
+
+  deployments = ostree_sysroot_get_deployments (sysroot);
+  g_ptr_array_remove (deployments, to_remove);
+
+  if (!ostree_sysroot_write_deployments (sysroot, deployments, cancellable, error))
+    return FALSE;
+
+  if (!ostree_sysroot_cleanup (sysroot, cancellable, error))
+    return FALSE;
+
+  return TRUE;
+}
+
+static gboolean
 apply_internal (ApplyData     *apply_data,
                 GCancellable  *cancellable,
                 GError       **error)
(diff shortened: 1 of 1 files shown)


eos-updater/apply.c
```diff
@@ -229,6 +230,96 @@ update_remote_branches (OstreeRepo   *repo,
...
 }
 
+static gboolean
+check_boot_free_space (GError **error)
+{
+  g_autofree gchar *cmdline = NULL;
+  struct statvfs boot_stat;
+  guint64 boot_free_bytes;
+  /* Require at least 100 MiB free on /boot */
+  const guint64 boot_min_free_bytes = 100 * 1024 * 1024;
```
wjt: It seems wrong to hardcode this number. And if I understand correctly each UKI is now more than 100 MiB so it's not sufficient to have 100 MiB free on `/boot`. Can we not calculate how much space will be required precisely?

eos-updater/apply.c
```diff
@@ -256,6 +347,20 @@ apply_internal (ApplyData     *apply_data,
   if (!ostree_sysroot_load (sysroot, cancellable, error))
     return FALSE;
 
+  /* Check that /boot has enough free space before deploying. A new
+   * kernel/initramfs image may need to be written there. If not enough
+   * space is available, try to free some by removing an unused deployment.
+   */
+  if (!check_boot_free_space (error))
+    {
+      if (!g_error_matches (*error, G_IO_ERROR, G_IO_ERROR_NO_SPACE))
```
wjt: You don't know that this function was not called with `error` set to `NULL`. You should pass `&local_error` to `check_boot_free_space`, and propagate it to `error` if needed. However...

eos-updater/apply.c
```diff
@@ -256,6 +347,20 @@ apply_internal (ApplyData     *apply_data,
...
+        return FALSE;
+      g_clear_error (error);
+
+      if (!remove_one_unused_deployment (sysroot, cancellable, error))
+        return FALSE;
```
wjt: This patch would make the upgrade operation fail in many more cases than before, e.g.:

1. If `/boot` cannot be stat:ed for some reason
2. If removing the rollback deployment fails
3. etc.

I think that this check should never make the `apply` operation fail. At most it should log a warning. Otherwise, if there is some bug in the free-space-checking operation, then OS updates will never be applied, and we won't be able to fix the bug. It is better to allow Apply to proceed. (That's even true in the case where we believe that `/boot` is too full. We should still try to proceed - it's possible we are wrong.)

eos-updater/apply.c
```diff
@@ -229,6 +230,96 @@ update_remote_branches (OstreeRepo   *repo,
...
+  g_autoptr(GPtrArray) new_deployments = NULL;
+  OstreeDeployment *to_remove = NULL;
+
+  ostree_sysroot_query_deployments_for (sysroot, NULL, &pending, &rollback);
+
+  /* Prefer removing the rollback (older) over the pending deployment */
+  if (rollback != NULL)
+    to_remove = rollback;
+  else if (pending != NULL)
+    to_remove = pending;
```
wjt, line 314: Why would it help to remove the pending deployment? IIUC the ESP is only updated when the pending deployment is finalized (on shutdown). So it follows that an existing pending deployment can't be consuming any space on the ESP.

eos-updater/apply.c
```diff
@@ -229,6 +230,96 @@ update_remote_branches (OstreeRepo   *repo,
...
+  deployments = ostree_sysroot_get_deployments (sysroot);
+  new_deployments = g_ptr_array_new_with_free_func (g_object_unref);
+  for (guint i = 0; i < deployments->len; i++)
+    {
+      OstreeDeployment *d = deployments->pdata[i];
+      if (d != to_remove)
+        g_ptr_array_add (new_deployments, g_object_ref (d));
+    }
+
+  if (!ostree_sysroot_write_deployments (sysroot, new_deployments, cancellable, error))
```
wjt: I believe this will work:

```suggestion
  deployments = ostree_sysroot_get_deployments (sysroot);
  g_ptr_array_remove (deployments, to_remove);

  if (!ostree_sysroot_write_deployments (sysroot, deployments, cancellable, error))
```

eos-updater/apply.c
```diff
@@ -229,6 +230,96 @@ update_remote_branches (OstreeRepo   *repo,
   return TRUE;
 }
 
+static gboolean
+check_boot_free_space (GError **error)
```
wjt: I think it is confusing that the return value `FALSE` from this function can mean either "an actual error occurred" or "boot has insufficient space".

wjt (changes requested):

eos-updater/apply.c
starnight: It is hard to get the new update OSTree's UKI, because it has not deployed.  But, we can get current booted OSTree's UKI `/boot/ostree/eos-xxxx/payg-image-$(uname -r).efi`, which is copied from `/lib/modules/$(uname -r)/payg-image.efi`.  I think we can calculate the minimum size from the size of `/lib/modules/$(uname -r)/payg-image.efi` * 1.1.  The more 10% is for the potential UKI image growing size.

eos-updater/apply.c
starnight: After system boots up normally, current booted ostree deployment which is the default ostree deployment will has a leading star, and the previous ostree deployment will be tagged as **rollback**:
```
$ ostree admin status
* eos 64623165ae27033dad554e592e3cd455fcab2cae412b5cfb69483f4e48ad073c.0
    Version: 6.0.11~beta1
    origin refspec: eos:os/eos/amd64/latest3
    GPG: Signature made Mon 09 Mar 2026 02:43:17 PM CST using RSA key ID 9E08D8DABA02FC46
    GPG: Good signature from "EOS OSTree Signing Key 1 <pgp+eosk1@endlessos.org>"
    GPG: Key expires Fri 30 Jun 2034 12:18:11 AM CST
    GPG: Signature made Mon 09 Mar 2026 02:43:17 PM CST using RSA key ID FCF17B17F1F8E157
    GPG: Good signature from "EOS Flatpak Signing Key 1 <pgp+efsk1@endlessos.org>"
    GPG: Key expires Tue 27 May 2031 06:50:32 AM CST
  eos be0cc3bb68a385a8ab07bbaea41b3ea2c6a432802cc42a8cbd88323442354660.0 (rollback)
    Version: 6.0.10
    origin refspec: eos:os/eos/amd64/latest3
    GPG: Signature made Wed 28 Jan 2026 12:50:00 PM CST using RSA key ID 9E08D8DABA02FC46
    GPG: Good signature from "EOS OSTree Signing Key 1 <pgp+eosk1@endlessos.org>"
    GPG: Key expires Fri 30 Jun 2034 12:18:11 AM CST
    GPG: Signature made Wed 28 Jan 2026 12:50:00 PM CST using RSA key ID FCF17B17F1F8E157
    GPG: Good signature from "EOS Flatpak Signing Key 1 <pgp+efsk1@endlessos.org>"
    GPG: Key expires Tue 27 May 2031 06:50:32 AM CST
```

If system boots up with **rollback** ostree deployment, we will see star moves the original rollback ostree deployment for this boot, and the default ostree deployment becomes **pending** for the next boot.
```
$ ostree admin status
  eos 64623165ae27033dad554e592e3cd455fcab2cae412b5cfb69483f4e48ad073c.0 (pending)
    Version: 6.0.11~beta1
    origin refspec: eos:os/eos/amd64/latest3
    GPG: Signature made Mon 09 Mar 2026 02:43:17 PM CST using RSA key ID 9E08D8DABA02FC46
    GPG: Good signature from "EOS OSTree Signing Key 1 <pgp+eosk1@endlessos.org>"
    GPG: Key expires Fri 30 Jun 2034 12:18:11 AM CST
    GPG: Signature made Mon 09 Mar 2026 02:43:17 PM CST using RSA key ID FCF17B17F1F8E157
    GPG: Good signature from "EOS Flatpak Signing Key 1 <pgp+efsk1@endlessos.org>"
    GPG: Key expires Tue 27 May 2031 06:50:32 AM CST
* eos be0cc3bb68a385a8ab07bbaea41b3ea2c6a432802cc42a8cbd88323442354660.0
    Version: 6.0.10
    origin refspec: eos:os/eos/amd64/latest3
    GPG: Signature made Wed 28 Jan 2026 12:50:00 PM CST using RSA key ID 9E08D8DABA02FC46
    GPG: Good signature from "EOS OSTree Signing Key 1 <pgp+eosk1@endlessos.org>"
    GPG: Key expires Fri 30 Jun 2034 12:18:11 AM CST
    GPG: Signature made Wed 28 Jan 2026 12:50:00 PM CST using RSA key ID FCF17B17F1F8E157
    GPG: Good signature from "EOS Flatpak Signing Key 1 <pgp+efsk1@endlessos.org>"
    GPG: Key expires Tue 27 May 2031 06:50:32 AM CST

```

Both **pending** and **rollback** are deployed ostree deployments.  If here comes another new updated ostree deployment, it is tagged as **staged**, instead of **pending**, nor **rollback**.  After `eos-updater-ctl apply`, the **staged** one is deployed and shows up, but has not been finalized.  System cannot undeploy current booted ostree deployment, only **unused** deployment.  So, the **unused** deployment must be either **rollback**, or **pending**.

starnight: The behavior of undeploying unused ostree deployment follows `ostree admin upgrade`.

eos-updater/apply.c
starnight: Thanks for guiding the glib's error design! :)

Fixed in the new commits.

eos-updater/apply.c
starnight: Modified as the new commits.

eos-updater/apply.c
starnight: Nice!  Thanks!

eos-updater/apply.c
starnight: The new commit passes argument enough_space into `check_boot_free_space()` to get the information.

eos-updater/apply.c
```diff
@@ -256,6 +365,28 @@ apply_internal (ApplyData     *apply_data,
...
 
+  /* Check that /boot has enough free space before deploying. A new
+   * kernel/initramfs image may need to be written there. If not enough
+   * space is available, try to free some by removing an unused deployment.
+   */
+  if (!check_boot_free_space (&enough_boot_space, &local_error))
+    {
+      g_propagate_error (error, g_steal_pointer (&local_error));
+      g_warning ("%s", local_error->message);
+      g_clear_error (&local_error);
```
wjt: This is guaranteed to crash because `g_steal_pointer (&local_error)` assigns `NULL` to `local_error`, so `local_error->message` will dereference a `NULL` pointer. Please test this code path.

```suggestion
      g_warning ("Failed to check for free space: %s", local_error->message);
      g_propagate_error (error, g_steal_pointer (&local_error));
```

wjt (changes requested):

eos-updater/apply.c
starnight: Check [g_steal_pointer()](https://docs.gtk.org/glib/func.steal_pointer.html)'s description:
> Sets pp to NULL, returning the value that was there before.

Wow!  It is not simply returning the pointer.  Thanks for the hint!

starnight: It is a hard to test the code path get into this if yes condition.  `check_boot_free_space()` only returns **false** when there is no `/proc/cmdline`, or no `/boot`.

But, normal path works.

starnight: Fixed as the new commit.

eos-updater/apply.c
```diff
@@ -256,6 +365,27 @@ apply_internal (ApplyData     *apply_data,
...
     return FALSE;
 
+  /* Check that /boot has enough free space before deploying. A new
+   * kernel/initramfs image may need to be written there. If not enough
+   * space is available, try to free some by removing an unused deployment.
+   */
+  if (!check_boot_free_space (&enough_boot_space, &local_error))
+    {
+      g_warning ("%s", local_error->message);
+      g_propagate_error (error, g_steal_pointer (&local_error));
```
wjt: Sorry to not notice this in the previous round but this is still incorrect. Setting `error` here without early-returning from the function means that if `ostree_sysroot_stage_tree` fails, `error` will already have been filled in: incorrect behaviour.

Since we are ignoring the error we should just free it and continue:

```suggestion
      g_clear_error (&local_error);
```

eos-updater/apply.c
```diff
@@ -256,6 +365,27 @@ apply_internal (ApplyData     *apply_data,
   if (!ostree_sysroot_load (sysroot, cancellable, error))
     return FALSE;
 
+  /* Check that /boot has enough free space before deploying. A new
+   * kernel/initramfs image may need to be written there. If not enough
+   * space is available, try to free some by removing an unused deployment.
+   */
+  if (!check_boot_free_space (&enough_boot_space, &local_error))
+    {
+      g_warning ("%s", local_error->message);
```
wjt:
```suggestion
      g_warning ("Failed to check free space on /boot: %s", local_error->message);
```

wjt (changes requested):
Sure, it's not trivial to  test but also not impossible: rename `/boot`? Overmount something on `/proc` that hides `/proc/cmdline`? Temporarily patch the code to read the wrong paths?

eos-updater/apply.c
starnight: Thanks!  And, fixed.

eos-updater/apply.c
starnight: Thanks!  And, fixed.

wjt (approved):

wjt:
I notice that this is targeting `eos6.0`. What is the plan for the `master` branch?

starnight:
These commits will be cherry-picked to master branch, too.  I make them on top of eos6.0 for having the CI build simply.  We disabled EOS master packages' legacy CI build for EOS 7, due to different packaging systems. https://github.com/endlessm/eos-build-meta/issues/85

starnight:
Thanks for the approvement. I'm going to merge it.

starnight:
Synchronized between master and eos6.0 branches.

